PR #63, opened 2018-03-13
Improved login flow

### LINKED ISSUE #41: Improved login flow
We should open a browser to the VIP Go Dashboard's new token interface and prompt the user to copy the token.

joshbetz:
Fixes #41

### DIFF
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -30,6 +30,17 @@ const rootCmd = async function() {
 		// Bypass helper function
 		args.parse( process.argv );
 
+		const c = await inquirer.prompt( {
+			type: 'confirm',
+			name: 'continue',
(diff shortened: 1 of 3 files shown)


src/bin/vip.js
```diff
@@ -26,6 +30,8 @@ const rootCmd = async function() {
 		// Bypass helper function
 		args.parse( process.argv );
 
+		opn( tokenURL, { wait: false } );
```
mjangda: Making a note here so I don't forget: we should warn the user that we'll be opening a browser before we do so. Having a CLI tool suddenly kick me into a browser with a login button would be a bit jarring experience.

joshbetz: Good call. I added a prompt in https://github.com/Automattic/vip/pull/63/commits/3d3d0724f9d8ee731f07f7c3dc8ee5466ad7dc40. Happy to change the wording.

PR #66, opened 2018-03-22
Query a single sync

joshbetz:
When a sync starts, get the Sync ID and use that to poll for updates
instead of polling for the latest updates.

### DIFF
--- a/src/bin/vip-sync.js
+++ b/src/bin/vip-sync.js
@@ -18,7 +18,7 @@ import { formatEnvironment } from 'lib/cli/format';
 
 const appQuery = `id,name,environments{
 	id,name,defaultDomain,branch,datacenter,syncProgress{
-		status,steps{name,status}
+		status,sync,steps{name,status}
 	}
 }`;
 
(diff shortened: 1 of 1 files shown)


src/bin/vip-sync.js
```diff
@@ -83,7 +82,25 @@ command( { appContext: true, appQuery: appQuery, childEnvContext: true, requireC
...
+						query: gql`query App( $id: Int, $sync: Int ) {
+							app( id: $id ){
+								id,name,environments{
+									id,name,defaultDomain,branch,datacenter,syncProgress( sync: $sync ){
+										status,sync,steps{name,status}
+									}
+								}
+							}
+						}`,
+						fetchPolicy: 'network-only',
```
joshbetz, line 97: @nickdaugherty @scottsweb ☝️Pointing this out because we may need to do something similar in the UI any time we're polling for updates. I spent some time today trying to figure out why the API data wasn't updating. 🙃

nickdaugherty: Interesting, so this bypasses cache then?

joshbetz: Yep.

I believe there's also a cachePolicy that defines caching behavior. In this case, the query was already cached, so it doesn't help here. The response is pretty small, so I don't think we care that it's cached. We can't use the cache for polling though.

PR #68, opened 2018-03-29
Change token redirect URL

scottsweb:
Relies on https://github.com/Automattic/vip-ui/pull/54 being deployed


scottsweb:
This is ready to go now

PR #95, opened 2018-05-01
Repo detection: search all git remotes

### LINKED ISSUE #92: Check all git remotes
When we automatically detect the repo, we should look at all git remotes and use the first one that matches `wpcomvip/*`. This will resolve any issues where our repo is not the "origin".

joshbetz:
Rather than require the wpcomvip repo to be the "origin", we can use the
first remote that pushes to wpcomvip.

Fixes #92

### DIFF
--- a/src/lib/cli/repo.js
+++ b/src/lib/cli/repo.js
@@ -16,12 +16,29 @@ export default async function getRepoFromGitConfig(): Promise<string> {
 
 	const config = ini.parse( fs.readFileSync( file, 'utf-8' ) );
 
-	let url = config[ 'remote "origin"' ].url;
-	url = url.replace( /.git$/, '' );
-	url = url.replace( 'https://github.com/', '' );
-	url = url.replace( 'git@github.com:', '' );
+	// Find the first 'wpcomvip' remote
+	for ( const key in config ) {
+		if ( 'remote' !== key.substring( 0, 6 ) ) {
+			continue;
+		}
 
-	return url;
+		if ( ! config[ key ].url ) {
+			continue;
+		}
+
+		if ( 0 > config[ key ].url.indexOf( 'wpcomvip/' ) ) {
+			continue;
+		}
+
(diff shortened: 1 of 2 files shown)


src/lib/cli/repo.js
```diff
@@ -16,12 +16,29 @@ export default async function getRepoFromGitConfig(): Promise<string> {
...
+		if ( 0 > config[ key ].url.indexOf( 'wpcomvip/' ) ) {
+			continue;
+		}
+
+		let repo = config[ key ].url;
+		repo = repo.replace( /.git$/, '' );
+		repo = repo.replace( 'https://github.com/', '' );
+		repo = repo.replace( 'git@github.com:', '' );
+
+		return repo;
```
mjangda, line 38: Should we add some tests for this?

joshbetz: 👍 created a new issue to track that https://github.com/Automattic/vip/issues/96

src/lib/cli/repo.js
```diff
@@ -16,12 +16,29 @@ export default async function getRepoFromGitConfig(): Promise<string> {
...
+	}
+
+	return;
```
mjangda, line 41: This doesn't match the file-not-found `return` above (`''` vs `undefined`). Should it?

src/lib/cli/repo.js
```diff
@@ -16,12 +16,29 @@ export default async function getRepoFromGitConfig(): Promise<string> {
...
+		if ( 'remote' !== key.substring( 0, 6 ) ) {
+			continue;
+		}
 
-	return url;
+		if ( ! config[ key ].url ) {
+			continue;
+		}
+
+		if ( 0 > config[ key ].url.indexOf( 'wpcomvip/' ) ) {
```
mjangda, line 29: Should we allow this for `Automattic/` as well?

mjangda (commented):
Noted two small things.

Issue #86, opened 2018-04-17
sync: Unauthorized error


mjangda:
My test account has `read` access to the `wpcomvip/vip-go-test-03` repo (site 103). Running `sync` without params throws an error instead of the environment list.

```
$ vip sync
Error: GraphQL error: Unauthorized
```

Running `sync --app 103` gives the following output (unsure if related to permissions or sync issues on the site):

```
$ vip sync --app 103
===================================
+ app: test-03
+ environment: staging
===================================
 Are you sure you want to sync from production? Yes
Note: Someone recently ran a data sync on this site
Note: Please wait a few minutes before trying again

  syncing: test-03
     from: PRODUCTION
       to: staging


✓ Data Sync is finished for test-03
```

mjangda:
The issue is from here:

https://github.com/Automattic/vip-go-api-public/blob/master/src/models/Repo.js#L73-L78

We're only allowing `admin` users to do repo lookups and for some reason, all users are being pulled in as a `member` user.

Also, we're not actually doing any capability checking for repo lookups beyond whether someone is an admin.

mjangda:
This was fixed at the API level.

PR #106, opened 2018-05-04
Only lookup repo + app if valid repo found

mjangda:
via appContext. Previously, if a repo was not found, we would attempt to do a query with the repo name as `undefined`, which would fail and throw an `Error: GraphQL error: Unauthorized`

See #86 (although, that issue still needs an API fix).


joshbetz (approved):

mjangda:
This code was removed in #113.

PR #113, opened 2018-05-26
Disable git context lookup

mjangda:
We need some API updates (https://github.com/Automattic/vip-go-api-public/issues/87) for this to work correctly. Remove for now until that is fixed.


joshbetz:
Looks like the test failures are just eslint.

mjangda:
Fixed the lint issues.

joshbetz (approved):

Issue #32, opened 2018-02-13
Usage Data (Stats)


joshbetz:

mjangda:
A couple of implementation thoughts:

- Can maybe borrow code from https://github.com/Automattic/wp-calypso/blob/master/client/lib/analytics/index.js
- Will need to use HTTP POSTs to do the actual tracking since we can't rely on the browser-based libraries in the CLI context (e.g. see https://github.com/40Digits/CLI-Analytics).
- Can generate an anonymized id for each user (`uuid/v4`) that persists between invocations (and as a bonus between the CLI and Dashboard -- we can pass it through from CLI to Dashboard on login)
- Would be nice to disable for development (so we don't pollute production stats).
- Would be nice to delay the pings so they don't block or slow down execution.

We can have a simple analytics module that exports a `trackEvent()` method

Things we want to track:
- What commands are being run
- Errors that are hit
- `vip` version
- OS and version
- node + npm version
- Bonus: Which app context method is being used? (`--app`; select from list; repo lookup?)
- Bonus: how long each command takes.

As an extension to this, we will also need some for of disclosure about what data we're collecting and possibly an easy opt-out (maybe just a config flag?) but that can be handled in a separate PR

joshbetz:
We have an uncaught error handler that would be a good spot to start tracking uncaught errors:

https://github.com/Automattic/vip/blob/25b6a759fe6a3ef1881c203d5683789655711f4b/src/lib/cli/command.js#L22-L28

joshbetz:
Just before this callback is probably a good spot to track actions. It's after we set up any necessary context and before the command is actually run:

https://github.com/Automattic/vip/blob/25b6a759fe6a3ef1881c203d5683789655711f4b/src/lib/cli/command.js#L221-L241

PR #117, opened 2018-06-08
Add UUID

joshbetz:
We need UUIDs to track usage data. We don't care who the user actually
is, but we need to be able to see problematic flows.

This generates a unique ID that cannot be reversed or
otherwise used to figure out who the user is. The ID is persisted in
whatever keychain is used on that given system.

See #32

### DIFF
--- a/src/lib/token.js
+++ b/src/lib/token.js
@@ -4,6 +4,7 @@
  * External dependencies
  */
 import jwtDecode from 'jwt-decode';
+import uuid from 'uuid/v1';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


src/lib/token.js
```diff
@@ -4,6 +4,7 @@
  * External dependencies
  */
 import jwtDecode from 'jwt-decode';
+import uuid from 'uuid/v1';
```
mjangda: I think we should probably use `v4` as it's more random.

mjangda (approved):
🚢

PR #122, opened 2018-06-18
Integrate Analytics

mjangda:
With the analytics lib introduced in #118, we can now actually integrate it to track things.

- [x] Integrate with login
- [x] Integrate with command framework (e.g. validation, help, context, etc.)
- [x] Integrate with `vip app`
- [x] Integrate with `vip app list`
- [x] Integrate with `vip sync`
- [ ] Add disclaimer on login
- [ ] Add opt-out
- [x] Make sure analytics failures are non-critical
- [x] Review all events to make sure they are necessary and useful
- [ ] Switch uuid to GH ID after login.


src/lib/cli/command.js
```diff
@@ -29,21 +28,17 @@ function uncaughtError( err ) {
...
-	const tracker = await initAnalytics();
-
 	const options = this.parse( argv );
 
 	const validationError = validateOpts( options );
 	if ( validationError ) {
-		console.log( validationError.toString() );
+		const error = validationError.toString();
+
+		trackEvent( 'command_validation_error', { error } );
```
joshbetz: Is there any risk of process.exit() firing before this Promise resolves?

joshbetz:
Overall looks good. There are some instances where I'm not sure if `trackEvent` will resolve before the CLI exits. The idea of "fire and forget" is nice, but I'm not sure how it works in practice since we're not working with a long-running process.

src/lib/cli/command.js
mjangda: Oh, good catch. We need `await` before all those calls.

mjangda:
> The idea of "fire and forget" is nice, but I'm not sure how it works in practice since we're not working with a long-running process.

Those are all supposed to be blocking async actions (which is now fixed).

I'm not totally happy about the implementation (both the blocking aspect and the extra complexity it adds). It would be nice to move towards a single request instead (maybe firing on `beforeExit` + `uncaughtException`) and a cleaner implementation (maybe using events?).

mjangda:
Going to move the the disclaimer, opt-out, and Tracks user alias to new Issues + PRs.

PR #131, opened 2018-06-21
Update args from 3.0.8 to 5.0.0

joshbetz:
* More compact layout (https://github.com/leo/args/pull/124)
* Fix options parsing (https://github.com/leo/args/pull/116)
* Updates dependencies

* https://github.com/leo/args/releases/tag/5.0.0
* https://github.com/leo/args/releases/tag/4.0.0

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -50,10 +50,20 @@ args.argv = async function( argv, cb ): Promise<any> {
 	// Check for updates every day
 	updateNotifier( { pkg, isGlobal: true, updateCheckInterval: 1000 * 60 * 60 * 24 } ).notify();
 
-	// Show help if no args passed
 	// `help` and `version` are always defined as subcommands
-	const defaultSubcommandCount = 2;
-	if ( this.details.commands.length > defaultSubcommandCount && ! this.sub.length ) {
+	const customCommands = this.details.commands.filter( c => {
+		switch ( c.usage ) {
+			case 'help':
+			case 'version':
+				return false;
+
+			default:
+				return true;
+		}
+	} );
+
+	// Show help if no args passed
+	if ( !! customCommands.length && ! this.sub.length ) {
 		await trackEvent( 'command_help_view' );
 
 		this.showHelp();
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -51,7 +51,9 @@ args.argv = async function( argv, cb ): Promise<any> {
 	updateNotifier( { pkg, isGlobal: true, updateCheckInterval: 1000 * 60 * 60 * 24 } ).notify();
 
 	// Show help if no args passed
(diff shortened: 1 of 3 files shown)


joshbetz:
Needs testing

joshbetz:
The most obvious advantages for us are the updated dependencies and the more compact layout:

Before:

```
$ vip app

  Usage: vip app [options] [command]

  Commands:

    help  Display help
    list  List your VIP Go apps

  Options:

    -f, --format [value]  Format results (defaults to "table")
    -h, --help            Output usage information
    -v, --version         Output the version number

  Examples:

    - Pass an app name or ID to get details about that app

    $ vip app <app>


    - Get details about the app with ID 123

    $ vip app 123


    - Get details about the app named vip-test

    $ vip app vip-test
```

After:

```
$ vip app
  Usage: vip app [options] [command]

  Commands:
    help     Display help
    list     List your VIP Go apps
    version  Display version

  Options:
    -f, --format [value]  Format results (defaults to "table")
    -h, --help            Output usage information
    -v, --version         Output the version number

  Examples:
    - Pass an app name or ID to get details about that app
    $ vip app <app>

    - Get details about the app with ID 123
    $ vip app 123

    - Get details about the app named vip-test
    $ vip app vip-test
```

src/lib/cli/command.js
```diff
@@ -50,8 +50,20 @@ args.argv = async function( argv, cb ): Promise<any> {
...
+				return false;
+
+			default:
+				return true;
+		}
+	} );
+
 	// Show help if no args passed
-	if ( this.details.commands.length > 1 && ! this.sub.length ) {
+	if ( !! customCommands.length && ! this.sub.length ) {
```
chrean, line 66: Why are there two different `bang` policies here, both on a `.length` params?

chrean (commented):
Looks good, I have only a minor question.

src/lib/cli/command.js
joshbetz: It's a common Javascript way to cast something to a Boolean. The first casts the value to a Boolean but inverts it so the second one is needed to invert it back.

chrean: I know, I just wanted to know why `!! customCommands.length` has 2 bangs while `! this.sub.length` has only one, thought they were similar.

mjangda: We're checking to see if we have custom commands defined (`!! customCommands.length`; looking for `true`) and not passing any args (`! this.sub.length`; looking for `false`).

I think for clarity, this would be better written as:

```
if ( customCommands.length > 1
    && this.sub.length === 0 ) {
```

or take it a step further:

```
const haveCustomCommands = customCommands.length > 1;
const runningSubcommand = this.sub.length === 0;

if ( haveCustomCommands && ! runningSubcommand ) {
```

PR #135, opened 2018-06-21
Throw on token set failures.

mjangda:
It's a critical issue and we should bail immediately.

### DIFF
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -102,7 +102,15 @@ const rootCmd = async function() {
 			return;
 		}
 
-		Token.set( token.raw );
+		try {
+			Token.set( token.raw );
+		} catch ( e ) {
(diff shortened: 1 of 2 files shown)


src/bin/vip.js
```diff
@@ -102,7 +102,15 @@ const rootCmd = async function() {
...
 
-		Token.set( token.raw );
+		try {
+			Token.set( token.raw );
+		} catch ( e ) {
+			await trackEvent( 'login_command_token_submit_error', {
+				error: e.message,
+			} );
+
+			throw e;
```
joshbetz, line 112: Should we output an error message and bail instead of throwing here?

mjangda: My thinking was that we'd fall back to the uncaught error handler which includes the pointer to contact us with the backtrace.

joshbetz (approved):

PR #136, opened 2018-06-21
Fix Error: spawn UNKNOWN

joshbetz:


src/bin/vip.js
```diff
@@ -107,10 +107,13 @@ const rootCmd = async function() {
 		await trackEvent( 'login_command_token_submit_success' );
 
 		// Exec the command we originally  wanted
-		const spawn = require( 'child_process' ).spawn;
-		const bin = process.argv[ 1 ];
 		const argv = process.argv.slice( 2 );
-		spawn( bin, argv, { stdio: 'inherit' } );
+		if ( argv.length ) {
+			console.log( argv );
+			return args.runCommand( { usage: process.argv.slice( 2 ) } );
```
mjangda: `ReferenceError: args is not defined` :)

src/bin/vip.js
```diff
@@ -107,10 +107,13 @@ const rootCmd = async function() {
...
+		}
+
+		return args.showHelp();
```
mjangda: Because we only define these when a token is set, we end up with an empty help:

```
  Usage: vip.js [options] [command]

  Commands:

    help  Display help

  Options:

    -h, --help     Output usage information
```

joshbetz: Good point. Should be fixed by ef34c55

src/bin/vip.js
```diff
@@ -107,10 +107,13 @@ const rootCmd = async function() {
...
-		const spawn = require( 'child_process' ).spawn;
-		const bin = process.argv[ 1 ];
 		const argv = process.argv.slice( 2 );
-		spawn( bin, argv, { stdio: 'inherit' } );
+		if ( argv.length ) {
+			return args.runCommand( { usage: process.argv.slice( 2 ) } );
+		}
+
+		const { spawn } = require( 'child_process' );
+		spawn( process.argv[ 0 ], process.argv.slice( 1 ), { stdio: 'inherit' } );
```
mjangda, line 116: :boom:

mjangda:
Works now!

mjangda (approved):

PR #137, opened 2018-06-21
Fix permissions issues for Windows

joshbetz:
The fs constants don't seem to work at all and files end up as read-only
which means we can't write any secrets. This solution is not perfect,
but should be OK for now.


joshbetz:
Closing in favor of #138

Issue #116, opened 2018-06-07
Pagination for the app list


simonwheatley:
Currently we only show the first page of 10 apps for `vip app list`, we should allow pagination to show more (and maybe increase the page size).

joshbetz:
I'm not sure if this is directly related to data sync?

simonwheatley:
I did muse a little before adding it to that project board (project board is private, for anyone following along, sorry), figuring that someone would push back if appropriate. (We need to review that board soon, it’s got way too many unprioritised issues on it.)

The issue as it relates to data sync is that some clients have more applications than they can see, so they find data sync partially inaccessible.

joshbetz:
I would like to loop over all pages and automatically pipe to `less` like `git log`, but I think we need to find out how that works on Windows first.

PR #139, opened 2018-06-22
Add $PAGER support

joshbetz:
Adds support to use a pager for long commands. By default we use `less -FRX`
which disables the init step (so the screen doesn't completely clear
when less starts), sends the raw output to less (so colors work), and
exits if less than one full screen is output.

We also honor the $PAGER environment variable, so the program that's
used for paging can be customized per http://manpages.ubuntu.com/manpages/bionic/en/man1/man.1posix.html

An easy way to test this is to change the `vip app list` to output 30
results instead of 10.

This is a good start, but it would be better if we could page over
results and write 10 at a time. It would also be nice if that was
pausable (like stream.pause()) so we could lazy load results (or never
load results) that aren't going to be on screen anyway.

See: #116

### DIFF
--- a/src/lib/cli/pager.js
+++ b/src/lib/cli/pager.js
@@ -7,6 +7,18 @@ import { spawn } from 'child_process';
 
 let proc;
 export default function pager() {
+	let less;
+	switch ( process.platform ) {
+		case 'win32':
+			// PROGRA~1 is the short name for Program Files
+			// we're using it here to avoid the space which is broken by .split( ' ' )
+			less = 'C:\\PROGRA~1\\Git\\usr\\bin\\less.exe -FRX';
+			break;
+
+		default:
+			less = 'less -FRX';
+	}
+
 	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
 	const bin = args.shift();
 
--- a/src/lib/cli/pager.js
+++ b/src/lib/cli/pager.js
@@ -1,0 +1,23 @@
+// @flow
+
+/**
+ * External dependencies
+ */
+import { spawn } from 'child_process';
+
+let proc;
+export default function pager() {
+	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
+	const bin = args.shift();
+
+	proc = spawn( bin, args, { stdio: [ -1, 1, 2 ] } );
+	proc.on( 'exit', () => {
+		proc.stdin.emit( 'done' );
+	} );
+
+	return proc.stdin;
+}
+
+export function end() {
+	proc.end();
+}
--- a/src/lib/cli/pager.js
+++ b/src/lib/cli/pager.js
@@ -19,7 +19,7 @@ export default function pager() {
 			less = 'less -FRX';
 	}
 
-	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
+	const args = ( process.env.PAGER || less ).split( ' ' );
 	const bin = args.shift();
 
 	proc = spawn( bin, args, { stdio: [ 'pipe', process.stdout, process.stderr ] } );
--- a/src/lib/cli/pager.js
+++ b/src/lib/cli/pager.js
@@ -10,7 +10,7 @@ export default function pager() {
 	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
 	const bin = args.shift();
 
-	proc = spawn( bin, args, { stdio: [ -1, 1, 2 ] } );
+	proc = spawn( bin, args, { stdio: [ 'pipe', process.stdout, process.stderr ] } );
 	proc.on( 'exit', () => {
 		proc.stdin.emit( 'done' );
 	} );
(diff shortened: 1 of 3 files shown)


joshbetz:
My current plan is to:

1. Refactor the formatData and related code to work on 1 line at a time and iterate with `for..of`
2. Where necessary, return custom iterables

This is nice because after step 1 everything will continue to work. The object we return is already iterable.

In cases where we want to return lots of data, we can implement a custom iterable that makes N API requests. We could make 1 request for every row or (more likely) only make a request every i.e. 10 rows. Since iterables support pausing, we can also not make more requests than we need -- maybe make 1 extra request in the background so that we never have to wait for things to load.

Note: This means we need to change the cli-table we use to return a full row for the header, but skip the first line of sequential rows so that each row doesn't have a double line between them. Pretty easy to implement, just something to keep in mind.

Implementing iterables: https://medium.freecodecamp.org/demystifying-es6-iterables-iterators-4bdd0b084082

joshbetz:
Actually... solving the immediate issue by bumping the app limit to 100. We can add lazy load support later.

joshbetz:
Tested on Windows and MacOS.

src/lib/cli/pager.js
```diff
@@ -0,0 +1,23 @@
...
+ * External dependencies
+ */
+import { spawn } from 'child_process';
+
+let proc;
+export default function pager() {
+	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
+	const bin = args.shift();
+
+	proc = spawn( bin, args, { stdio: [ -1, 1, 2 ] } );
```
mjangda: Minor: Can we switch up these numbers to constants?

joshbetz: Yeah, good call. I had been experimenting with these while debugging. Fixed now.

src/lib/cli/pager.js
```diff
@@ -7,6 +7,18 @@ import { spawn } from 'child_process';
 
 let proc;
 export default function pager() {
+	let less;
+	switch ( process.platform ) {
+		case 'win32':
+			// PROGRA~1 is the short name for Program Files
+			// we're using it here to avoid the space which is broken by .split( ' ' )
+			less = 'C:\\PROGRA~1\\Git\\usr\\bin\\less.exe -FRX';
```
joshbetz, line 16: @mjangda this works in cmd.exe and Git bash for me.

mjangda: I guess there will be the edge case of users who choose not to install stuff in `Program Files`. 

How easy would it be to fallback to `console.log` in cases where we can't find an appropriate pager lib? Is it worth factoring in?

src/lib/cli/pager.js
```diff
@@ -0,0 +1,35 @@
...
+			break;
+
+		default:
+			less = 'less -FRX';
+	}
+
+	const args = ( process.env.PAGER || 'less -FRX' ).split( ' ' );
```
mjangda: `'less -FRX'` needs to be replaced with `less`

src/lib/cli/pager.js
joshbetz: I think it would be easier to bundle `less.exe` and reference it relative to the CLI.

joshbetz: I thought maybe https://github.com/Automattic/vip/pull/139/commits/157f9e634b2a1537cd914431aead1582daf02436 would work, but it still triggers the uncaught error handler :(

joshbetz:
@mjangda 887ee33 falls back to `process.stdout`. I think this is pretty good now.

src/lib/cli/pager.js
chrean: My 2 cents as I am sure you already discussed this, but I think if we want to support Windows properly, we should probably either check for `less.exe` existence on target systems by using something like `where`, or at least give clients an option to specify a PATH to the exe? Or is there a more reliable way to understand if the binary is present on Windows (perhaps checking `%path`?).
If we'd chose to bundle it instead, which version?
Probably not a huge deal, depending on how badly we want to support Windows.

chrean (commented):
Only left a comment on Windows' `less`, nothing super-important though.

src/lib/cli/pager.js
joshbetz: I dealt with this in https://github.com/Automattic/vip/pull/139/commits/887ee33f668c15de5b96110fcf11f85245bff73f by falling back to sending output directly to `process.stdout` if the `less` or `less.exe` binary doesn't exist (or any other problems we run into trying to exec that binary).

If we want to check whether the binary exists first, I think we'd want a pure Node implementation of `where`/`which`/`type` because otherwise it's going to get pretty complicated. We need to `exec()` to run the program that checks for the existence of another program, get the output from that, and then decide where we're going to output. It also adds another "if Windows" conditional because the program to check if another program exists is different.

We do always check `process.env.PAGER`, so if you've got less installed in a non-standard place, you could specify it.

Git is required to run `npm install`, which is currently required to install our CLI, so I think it's going to be rare that `less.exe` is missing, but if it is, there's a graceful fallback.

chrean: Right, those will be edge cases and we have a fallback.

mjangda:
> I think this is pretty good now.

👍

PR #142, opened 2018-06-29
Trim whitespace from tokens

### LINKED ISSUE #141: When accepting token, strip trailing or prefixed whitespace
A user had difficulty authenticating, as the token they were pasting had a trailing space character. Should we strip trailing and prefixed whitespace from the token before checking it?

joshbetz:
As tokens are copy/pasted, it's easy to include extra whitespace. We can
just ignore that whitespace since there shouldn't be any whitespace in
the token anyway.

Fixes #141

### DIFF
--- a/__tests__/lib/token.js
+++ b/__tests__/lib/token.js
@@ -12,11 +12,13 @@ describe( 'token tests', () => {
 		expect( token.valid() ).toEqual( true );
 		expect( token.expired() ).toEqual( false );
 	} );
+
 	test( 'should correctly validate token missing an id', () => {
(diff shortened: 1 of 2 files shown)


mjangda:
Minor, but would be nice to add tests for a valid token with trailing/leading whitespace and an invalid token that's just whitespace.

__tests__/lib/token.js
```diff
@@ -27,13 +29,28 @@ describe( 'token tests', () => {
...
 	} );
+
 	test( 'should not validate expired token', () => {
 		const t = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MiwiaWF0IjoxNTE1NzExMDY5LCJleHAiOjE1MTU3OTc0Njl9.hZ-mAeoFAahak9WXqAVTOKEU7R_f1VsZfS5HqZOm-a4';
 		const token = new Token( t );
 		expect( token.valid() ).toEqual( false );
 		expect( token.expired() ).toEqual( true );
 	} );
 
+	test( 'should correctly validate token with invalid whitespace', () => {
```
mjangda, line 40: 👍 

(If we have more patterns like this, we can look at https://github.com/mattphillips/jest-each for looping through and running the same test against a given test dataset.)

mjangda (approved):

PR #143, opened 2018-06-29
Remove package-lock.json

joshbetz:
Fail the build if we forget to update the package-lock.json file.

`--exit-code` causes git diff to return 0 if there are no differences and
1 if there are differences.


joshbetz:
This is pretty annoying because apparently different versions of Node also generate different `package-lock.json`. In that case, it seems like we should remove it from the repo and add to `.gitignore`.

mjangda:
> In that case, it seems like we should remove it from the repo and add to .gitignore.

I've found that older npm versions don't auto-update on `npm install`/`npm update`.

Should we enforce a specific version during development (maybe 7.x since that's the minimum we support)?

joshbetz:
> Should we enforce a specific version during development (maybe 7.x since that's the minimum we support)?

You can also disable it locally with `npm config set package-lock false`. I don't know if we can easily enforce a certain npm version? I'm not super clear on what the benefits of package-lock are and if it's worth enforcing everything that writes code for this repo to use an old npm version.

mjangda:
Closing; it's annoying to deal with but I think we probably should keep it around. `npm` has some tools to simplify things like merge conflicts as well so we can try using those to minimize the annoyances.

PR #146, opened 2018-07-12
Don't fork PAGER unless we have data to output

### LINKED ISSUE #145: vip app list: Don't pipe to $PAGER if no apps found

joshbetz:
We were forking `less` before even making the API request for data.
If there was an error (for example, if the network was disconnected)
less would eat the error and fail to exit because p.end() was only
called if there was data returned.

Fixes #145


mjangda (approved):
Looks fixed; 🚢

PR #163, opened 2018-09-06
Add primary domain and remove default domain

### LINKED ISSUE #155: app: default domain is incorrect
Should show the primary domain, not the `go-vip.co` address.

elazzabi:
Fixes #155 

This changes default domain with primary domain

### How to test
1. Clone branch
2. Run `npm link`
3. Run `vip app <id>` (ex sites having a domain mapped: 844, 647...)
4. Make sure the `primary domain` column is correct
5. Run `npm unlink` when done

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,9 @@
     "publish-please": "publish-please",
     "prepublishOnly": "publish-please guard"
   },
+  "jest": {
+    "testURL": "http://localhost/"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/Automattic/vip.git"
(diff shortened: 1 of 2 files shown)


mjangda:
@elazzabi can you include testing instructions, please?

elazzabi:
@mjangda oups my bad. Included now!

nickdaugherty (dismissed):
Tested and works great!

Something I just noticed is that the `vip app list` command has different default output than `vip app <id>` (doesn't include some fields including primaryDomain). Ideally those would be consistent, but that's something to fix in a different branch.

PR #178, opened 2018-11-01
Allow publishing from either `master` or `release`

nickdaugherty:
Tweaks our `.publishrc` config to allow publishing from either the `master` or `release*` branches.

Sometimes we need to release patches without committing them directly to master, this will allow us to do that.

Without this, `npm publish-please` will fail when validating the branch.

### DIFF
--- a/.publishrc
+++ b/.publishrc
@@ -4,7 +4,7 @@
     "uncommittedChanges": true,
     "untrackedFiles": true,
     "sensitiveData": true,
-    "branch": "master",
+    "branch": /(master|release)/,
     "gitTag": true
   },
   "confirm": true,
--- a/.publishrc
+++ b/.publishrc
@@ -4,7 +4,7 @@
     "uncommittedChanges": true,
     "untrackedFiles": true,
     "sensitiveData": true,
-    "branch": /(master|release)/,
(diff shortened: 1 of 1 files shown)


.publishrc
```diff
@@ -4,7 +4,7 @@
     "uncommittedChanges": true,
     "untrackedFiles": true,
     "sensitiveData": true,
-    "branch": "master",
+    "branch": /(master|release)/,
```
mjangda: If the condition we want is:

> allow publishing from either the `master` or `release*` branches.

Then this needs to change a bit. Probably something like `/^(master|(release\/\d\.\d))$/` (either `master` or `release/x.y`)?

nickdaugherty: I _think_ that's the situation we want, right? 1 release branch per release.

nickdaugherty: Updated in 5b81295 - tweaked the regex slightly to match patch versions - `release/1.0.0`

joshbetz:
The documentation says:

> Force usage of the `npm publish tag` there necessary, so you'll be sure you're not publishing preview version of your package as a release version.

I don't see any documentation on how to do that, but I'm wondering if we just change `publishCommand` to `npm publish tag` if that will do it.

PR #210, opened 2018-11-16
Add login subcommand

### LINKED ISSUE #196: Console error upon login
After completing the login flow, an error is logged to the console:

```
  Error: spawn vip-login ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:238:19)
    at onErrorNT (internal/child_process.js:413:16)
    at process.internalTickCallback (internal/process/next_tick.js:72:19)
```

The CLI works after that - the token is saved and subsequent commands are authenticated, so it's not the auth process itself, just something with how the process is shutting down.

## Version

This is on v1.1.1 on OSX Mojave.

elazzabi:
This adds the login subcommand to the `cli`.

It also fixes these types of errors: (https://github.com/Automattic/vip/commit/6c26502a92e0501524283ad4e9b5ed7fcc30a2ab)
```
  Error: spawn vip-login ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:238:19)
    at onErrorNT (internal/child_process.js:413:16)
    at process.internalTickCallback (internal/process/next_tick.js:72:19)
```

### How to test:
1. clone this branch
2. execute `vip logout`
3. execute `npm link`

#### First test:
4. execute `vip aaa`, **you should be redirected** to the `login` flow
5. when prompted with `Ready? (Y, n)`, respond with `n`. You **should quit the command**.

#### Second test:
4. execute `vip aaa`, and complete the `login` workflow
5. when the `login` workflow is done, your `aaa` command **should execute after**. and you should **get an error that the subcommand isn't defined**

fixes #196

### DIFF
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -28,6 +28,7 @@ const rootCmd = async function() {
 				await trackEvent( 'logout_command_execute' );
 			} )
 			.command( 'app', 'List and modify your VIP Go apps' )
+			.command( 'login', 'Login to the vip cli' )
 			.command( 'sync', 'Sync production to a development environment' )
 			.argv( process.argv );
 	} else {
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -5,8 +5,6 @@
  * External dependencies
  */
 import args from 'args';
-import opn from 'opn';
-import inquirer from 'inquirer';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


src/bin/vip.js
```diff
@@ -27,101 +22,30 @@ const rootCmd = async function() {
...
-		console.log();
-
-		console.log( `  To get started, we need an access token for your VIP account. We'll open ${ tokenURL } in your web browser; follow the instructions there to continue.` );
-		console.log();
+		const { spawn } = require('child_process');
 
-		await trackEvent( 'login_command_execute' );
+		// run login first
+		// can not run the login command here as they are defined only if user is logged in
+		const loginSpawn = spawn( process.argv[ 0 ], [ 'dist/bin/vip-login.js' ], { stdio: 'inherit' } );
```
nickdaugherty, line 50: Rather than requiring an `exit` callback, could we maybe use `spawnSync` to block until `login` finished?

src/bin/vip.js
```diff
@@ -27,101 +22,30 @@ const rootCmd = async function() {
...
-			prefix: '',
-		} );
+		loginSpawn.on( 'exit', async ( code, signal ) => {
+			let token = await Token.get();
 
-		if ( ! c.continue ) {
-			await trackEvent( 'login_command_browser_cancelled' );
+			if ( token && token.valid() ) {
+				// user is logged in now, run desired command:
+				return spawn( process.argv[ 0 ], process.argv.slice( 1 ), { stdio: 'inherit' } );
```
nickdaugherty, line 57: When I use `vip login` to login, and complete the flow, I get a redundant message:

```
 Ready? Yes
 Access Token: [hidden]
You are now logged in!
You are already logged in!
```

I think we need to check if the target command was `login`, and if so, skip the second `spawn` here.

nickdaugherty (changes requested):
Overall looks good - left a few notes.

I'm not sure if it's related to this PR (I hadn't noticed it before at least), but `vip logout` is throwing a tracks-related error for me:

```
$ vip logout

  ReferenceError: trackEvent is not defined
    at Args.command (/Users/nickdaugherty/www/vip/dist/bin/vip.js:26:7)
    at process._tickCallback (internal/process/next_tick.js:43:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:763:11)
    at startup (internal/bootstrap/node.js:308:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:878:3)
```

Oh also, `vip logout` tries to run `vip login` automatically when you're logged out, which is 🤣 ... guess we need an exception for that one 😄 .

```
$ vip logout

  Welcome to
 
... blah blah...

  To get started, we need an access token for your VIP account. We'll open https://dashboard.wpvip.com/me/cli/token in your web browser; follow the instructions there to continue.

 Ready? No
```

elazzabi:
other than the `trackEvent` error (a missing import, not sure what `eslint` was doing 🤣 ), all those seems to be some edge cases we need to fix:
+ if it's `login` do not execute the second spawn
+ if it's `logout` do not execute the `login`

I'm not sure if we'll find more edge cases in the future...

We can fix those two ^ by skipping things, or we can rethink how we are approaching things (login before running a command, declaring commands only if logged in (https://github.com/Automattic/vip/issues/211)...) and see if we can make it better once and for all

What do you think @nickdaugherty @joshbetz ?

nickdaugherty:
> We can fix those two ^ by skipping things, or we can rethink how we are approaching things

For now I'd just suggest adding the exceptions for those 2 commands. There probably _are_ other edge cases we haven't found, but until we know what they are, we won't know if changing the approach will have an impact on those.

joshbetz:
Are we sure `vip login` is necessary? I think it might only seem confusing because our (internal) CLI has a login command. It should be possible to fix the root issue, `Error: spawn vip-login ENOENT`, without adding this subcommand. My main objection is that the CLI should just handle login for you since every command requires you to be logged in. Is there any case when you'd need just the login command?

nickdaugherty:
I've seen several developers (internally) attempting to use `vip login` (myself included) - I guess b/c they are assuming that's how you log in, or are used to other tools that use that workflow.

Putting myself in the shoes of a new user that's unfamiliar with how the CLI tool works, the first thing I would probably do is `vip -h` to get a lay of the land. If that doesn't list anything about authentication, it might make me a little confused (until I read docs or tried a command).

Definitely agree that having the auto-login is nice (and we still have that), this just exposes the same functionality so it can be run manually if desired.

joshbetz:
> If that doesn't list anything about authentication, it might make me a little confused (until I read docs or tried a command).

This is another thing that needs to be fixed because right now it doesn't list anything about anything if you're not logged in. I do think we need better help docs.

elazzabi:
Do you think fixing the help subcommand should be in this PR or another one? Are we ok to add the `login` subcommand?

elazzabi:
As discussed, we'll skip adding the `login` subcommand. Instead:
1. Help should display all commands even if not logged in
2. If command isn't correct, display a better message (other than spawn error)

Will probably start another PR and close this :)

cc @nickdaugherty @joshbetz @mjangda

elazzabi:
this is fixed by #214

PR #229, opened 2019-01-01
Replace inquirer with enquirer

elazzabi:
This replaces `inquirer` with `enquirer`

It also introduces an `autocomplete` feature when selecting an app.

### Tests
1. get PR
2. run `npm i`
3. run `npm link`
4. run `npm sync`
5. you'll get **an autocomplete** list to chose an app (first 100 for now, more introduce a timeout)
6. start typing and select an app
7. after confirmation you'll get prompted to select an environment from **a list**
8. abort (to not launch a sync :) )


nickdaugherty (commented):
Changes here look good, but we should hold off merging pending the outcome of #227

elazzabi:
Removed the changes from #227 , this should be ready for another review @nickdaugherty :)

nickdaugherty (approved):
LGTM 🚢

PR #242, opened 2019-01-25
Rename scripts folder to helpers

elazzabi:
When we deploy a new version, we deploy an extra `/scripts` folder containing two files.

`prepare-config` prepares the config we'll use to build, thus not used and shouldn't be uploaded to npm
`check-version` is used as a post installation step and will be in the `dist` folder after the build

with these change we can turn on the sensitive data check from `publish-please` back https://github.com/Automattic/vip/blob/master/.publishrc#L6

### Tests
1. Get PR
2. Run `npm run build`
3. No errors should be thrown
4. Run `npm run postinstall`
5. No errors should be thrown and no logs (file log error only if there is one)


elazzabi:
Looks like the `postinstall` step is executed after running a normal npm install on the package itself 👀

elazzabi:
@joshbetz & @nickdaugherty would like your input on this:

We are using `publish-please` to publish our package to npm, and in the validation steps, there is a step to check for sensitive data. This step is currently disabled as it fails due to our `scripts` folder (considered as sensitive files by `publish-please`)

In this PR, I tried to remove the `scripts` folder by moving files inside `src` and `dist`. However, the `postinstall` script now is: `node ./dist/check-version.js` which suppose that we already built the solution before publishing to `npm` (true in case of publishing). But given this script is also executed after running a simple `npm install` on the project itself (locally, on Travis...) the install throws an error because there is no `./dist` folder.

According to `publish-please` we can only override the list of files and not add exceptions for our script files for example ([here](https://github.com/inikulin/publish-please) _Checking for the sensitive and non-essential data in the npm package_ section)

I don't think overriding all the list is good as we'll lose lot of checks, but looking at that list, it looks like [they are checking](https://github.com/inikulin/publish-please/blob/master/.sensitivedata#L109) for a folder called `scripts`. So renaming this folder to something else will do the trick.

**What do you think? Should we go with renaming it or do you think of any other workaround?**

nickdaugherty:
What if we override the list with the full list from their defaults, minus `scripts`? That way we get customization while maintaining all the good stuff (at the cost of missing out on updates).

mjangda:
> Should we go with renaming it or do you think of any other workaround?

Renaming the folder seems like a quick and easy fix here.

nickdaugherty (approved):
LGTM

PR #255, opened 2019-02-27
WP CLI prototype

nickdaugherty:
Prototyping WP CLI command

Currently it sets up a stream from local API (so need corresponding PR running locally).

Currently just outputs random data streamed from API over websocket.

```
vip wp option get home
```

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -152,6 +152,16 @@ commandWrapper( {
 			rl.on( 'line', line => {
 				stdinStream.write( line + '\n' );
 			} );
+
+			rl.on( 'SIGINT', () => {
+				rl.question( 'Are you sure you want to exit? ', answer => {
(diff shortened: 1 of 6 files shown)


src/bin/vip-wp.js
```diff
@@ -0,0 +1,123 @@
...
+import Token from '../lib/token';
+
+const appQuery = `id, name, environments {
+	id
+}`;
+
+command( {
+	requiredArgs: 2,
+	appContext: true,
+	childEnvContext: true,
```
nickdaugherty: @elazzabi we need to change this back to `envContext` - `childEnvContext` disallows running on prod, which is not what we want here.

elazzabi: yes, just reverted it 🙌

elazzabi (approved):
Looks good 👀

PR #267, opened 2019-03-11
Add publishing under next instructions

elazzabi:
This adds instructions on how to publish a new version under the `next` tag


CONTRIBUTING.md
```diff
@@ -58,6 +58,20 @@ Then, let's publish:
...
+### Test Releases
+
+Sometimes, we want to release a version we can test before releasing it to the public. In order to that, we need to release it under a tag other than `latest`, usually `next`. By default, `npm` install from the `latest` tag, so if `@next` isn't specified explicitely in the installation command like `npm install @automattic/vip@next`, a user will not get this version.
+
+In order to do that, please follow this:
+
+1. Manually change the version in `package.json` and `package-lock.json` to a dev version. Example: `1.4.0-dev1`
+2. Go to publish-please's config in `.publishrc`
+3. Change the `publishTag` to `next` and `gitTag` to `false` (publish-please will expect the latest commit to have a git tag, but we don't want it in this case)
+4. Commit your changes
```
nickdaugherty: What branch would this be committed to?

elazzabi: given we don't work with branches for new versions, all this is currently done on `master`. if we start supporting multiple branches in the future, at least for each major version, this needs to be done on the appropriate branch: `master` for current major version, then on every branch for older versions we are still supporting

nickdaugherty: Alright, probably worth adding that to the readme here for clarity.

elazzabi: Thanks for the feedback, does this looks good https://github.com/Automattic/vip/pull/267/commits/53df3e1ca2f97ed51147813e04ebce9654572634?

nickdaugherty (approved):
LGTM

CONTRIBUTING.md
nickdaugherty: Yep LGTM

PR #268, opened 2019-03-11
Use type instead of name

### LINKED ISSUE #266: Wrong environment name on production sites
On production sites, we're currently displaying the `name` in environment pickers and using the `name` in the `--env` parameter.

This should actually use the `environment_name` of a site, so that pickers and `--env` both use `production` for the prod environment.

elazzabi:
fix #266 

This should display `production` for the production environment

### Steps to test
1. Get this PR
2. Run `npm i` & `npm link`
3. Test it on an app, ex `647`

### DIFF
--- a/src/bin/vip-app.js
+++ b/src/bin/vip-app.js
@@ -9,7 +9,7 @@ import chalk from 'chalk';
 /**
  * Internal dependencies
  */
-import command from 'lib/cli/command';
+import command, { getEnvIdentifier } from 'lib/cli/command';
 import app from 'lib/api/app';
 import { trackEvent } from 'lib/tracker';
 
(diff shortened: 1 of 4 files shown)


joshbetz:
I think there might have been some confusion because of how these fields are named in the API. It looks like the `env.name` is the `client_site_id.name` and the `env.type` is the `client_site_id.environment_name`. I would have expected the `env.name` to  be `client_site_id.environment_name`.

nickdaugherty:
Ya it is confusing - what you say is right @joshbetz - each environment has a Type (production, staging, develop, etc), then an optional Name for disambiguation if there are multiple environments of a given type.

Then an App has a Name (not a Type), which is `client_site.name`.

elazzabi:
Do we move this to the API level. Do we know if we are going to break something if we change that on the API level?

nickdaugherty:
I've added some code to make the env picker work on sites that have multiple environments of a given type, which would have stopped working when switching to `type` over `name`.

Now, when there are multiple environments of a given type, they should be specified with `--env=develop.dev-01`, rather than `--env=dev-01`.

nickdaugherty (approved):
LGTM, but I defer the review of my additions to you @elazzabi

elazzabi:
Looks good to me too 👍

PR #273, opened 2019-03-12
Use separate tokens per host

### LINKED ISSUE #262: Tie auth token to API_HOST
To support having different tokens for production and local development, we should tie the token that is saved / read to the current value of `API_HOST` - we could store the tokens in JSON format, keyed by the host.

This would provide us with the ability to be logged into both local dev and production at the same time, preventing the need to log out then in again when switching between them.

nickdaugherty:
To make local dev easier, we should key tokens by API host, so that we can be logged in to both production and a local instance at the same time.

This means we no longer need to logout / login when switching between environments, which will make development faster and easier.

NOTE - this is designed to be backwards compatible with existing token saving - it will behave the same as before unless `API_HOST=<some-other-host>` is set in the env vars.

Fixes #262

## Testing

1. Be already logged into production
1. Start Parker API locally
1. Pull PR - `update/262-separate-tokens-per-host`
1. `npm run build`
1. (Optionally `npm link`)
1. Test production request again (should still be logged in and should work normally) - `./dist/bin/vip-app-list.js`
1. Test local request - `API_HOST=http://localhost:4000 ./dist/bin/vip-app-list.js`
1. Should be logged out - token is invalid
1. Logout on local dev - `API_HOST=http://localhost:4000 ./dist/bin/vip.js logout`
1. Login on local dev - `API_HOST=http://localhost:4000 ./dist/bin/vip.js` (use token for local API instance)
1. Try local request again (should succeed) - `API_HOST=http://localhost:4000 ./dist/bin/vip-app-list.js`
1. Test production request again (should succeed normally) - `./dist/bin/vip-app-list.js`

### DIFF
--- a/__tests__/lib/token.js
+++ b/__tests__/lib/token.js
@@ -2,7 +2,7 @@
 /**
  * Internal dependencies
  */
-import Token from 'lib/token';
+import Token, { SERVICE } from 'lib/token';
 
 describe( 'token tests', () => {
 	it( 'should correctly validate token', () => {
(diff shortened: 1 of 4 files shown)


elazzabi (approved):
Work as expected 🙌 

Thank you for implementing this 🙌

nickdaugherty:
One windows test is succeeding but then timing out, boo.

PR #278, opened 2019-03-19
Fix error handling

### LINKED ISSUE #277: WP-CLI: Better error code when permission denied
When a user attempts to run CLI commands without the correct permissions (for example a site they don't have access to) an uncaught exception will be thrown. This should probably be a UnauthorizedError 

Currently it's something along the lines of:
`TypeError: Cannot read property 'statusCode' of undefined`

## Steps to reproduce:
`vip wp --app=123 -- option get home`

nickdaugherty:
Fixes the `unauthorizedLink` to first check `networkError` (because it can be `null`).

Then adds generic GraphQL error handling to the network link, to abort the CLI if an error is encountered. NOTE - at some point it may be necessary or expected to receive _some_ errors (not all GraphQL errors are fatal) - we can revisit this and maybe add some exceptions in that case.

Fixes #277

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -29,7 +29,7 @@ export default async function API(): Promise<ApolloClient> {
 	}
 
 	const unauthorizedLink = onError( ( { networkError } ) => {
-		if ( networkError.statusCode === 401 ) {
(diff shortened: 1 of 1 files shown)


dchymko (approved):
LGTM

PR #281, opened 2019-03-22
Make wp shell impossible in sub-shell mode

elazzabi:
This makes running `wp shell` impossible in the sub-shell mode

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -141,9 +141,15 @@ commandWrapper( {
 
 				const startsWithWp = line.startsWith( 'wp ' );
 				const empty = 0 === line.length;
+				const isShellCommand = 'wp shell' === line;
 
 				if ( empty || ! startsWithWp ) {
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -141,12 +141,18 @@ commandWrapper( {
 
 				const startsWithWp = line.startsWith( 'wp ' );
 				const empty = 0 === line.length;
+				const isShellCommand = 'wp shell' === line;
```
nickdaugherty: I wonder if we should be more flexible in the check here, to handle any params (like `--url`) that get tacked on:
```suggestion
				const isShellCommand = line.startsWith( 'wp shell ' );
```

nickdaugherty (commented):
What do you think about a more permissive check for `wp shell`?

elazzabi:
@nickdaugherty you're right we need more permissive check, just committed 👍

nickdaugherty (approved):
🚢

elazzabi:
one test is failing on Windows, can be fixed by relaunching the test, merging

PR #288, opened 2019-03-27
Add error message for rate limit

### LINKED ISSUE #283: Handle rate limit better
Currently when a user send a batch of more than 2 commands, an error like this raises:

```
  Error: Unhandled "error" event. (Rate limit exceeded)
    at Socket.emit (events.js:186:19)
    at Socket.<anonymous> (/usr/local/lib/node_modules/@automattic/vip/node_modules/component-bind/index.js:21:15)
    at Socket.Emitter.emit (/usr/local/lib/node_modules/@automattic/vip/node_modules/component-emitter/index.js:133:20)
    at Socket.emit (/usr/local/lib/node_modules/@automattic/vip/node_modules/socket.io-client/lib/socket.js:138:10)
    at Socket.onpacket (/usr/local/lib/node_modules/@automattic/vip/node_modules/socket.io-client/lib/socket.js:256:12)
    at Manager.<anonymous> (/usr/local/lib/node_modules/@automattic/vip/node_modules/component-bind/index.js:21:15)
    at Manager.Emitter.emit (/usr/local/lib/node_modules/@automattic/vi....
```
We need to catch those errors and display a better message

elazzabi:
This handles the rate limit error and displays an error message:

`Error: you can not run more than two commands in the same time against an environment.`

Ideally we should check for the `code` of the error object instead of the message, but looks like the error received isn't an object, it's a string containing the error message :/

### Testing
1. Get this branch
2. Run `npm link`
3. Enter the subshell mode for an environment
4. Paste the following
```
wp option get home
wp option get home
wp option get home
```
5. You should see the error message
6. Try with two only
```
wp option get home
wp option get home
```
7. Should work

fix #283

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -96,7 +96,7 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 
 	socket.on( 'error', err => {
 		if ( err === 'Rate limit exceeded' ) {
-			console.log( chalk.red( '\nError:' ), 'Maximum concurrent running commands have been reached, please allow some to finish before starting another.' );
+			console.log( chalk.red( '\nError:' ), 'Rate limit exceeded: Please wait a moment and try again.' );
 			return;
 		}
 
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -89,7 +89,17 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 		console.log( 'There was an error with the authentication:', err.message );
 	} );
 
+	IOStream( socket ).on( 'error', err => {
+		// This returns the error so it can be catched by the socket.on('error')
+		return err;
+	} );
+
 	socket.on( 'error', err => {
+		if ( err = 'Rate limit exceeded' ) {
+			console.log( chalk.red( '\nError:' ), 'you can not run more than two commands in the same time against an environment.' );
+			return;
+		}
+
 		console.log( err );
 	} );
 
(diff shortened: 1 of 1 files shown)


simonwheatley:
Where does "two" come from? What happens if an application has more or less than two CLI containers?

elazzabi:
@simonwheatley currently, every site has one container that can run 2 commands at the same time. But if we have more than one, that error message isn't valid anymore, my bad. Thank you for spotting this.

Any idea how we can phrase this ^?

nickdaugherty:
> Any idea how we can phrase this

How about - "Maximum concurrent running commands have been reached, please allow some to finish before starting another."

simonwheatley:
Let's go with @nickdaugherty's suggestion 👍

nickdaugherty:
This "rate limit" error is actually unrelated to the maximum number of concurrent commands allowed - it's a rate limit in Parker designed to prevent abuse, so the wording I suggested is actually wrong. It needs to be something like:

"Rate limit exceeded: Please wait a moment and try again".

Aside - is there a way we can prevent copy/pasting a big bunch of commands? Ideally the subshell only allows 1 to be triggered at a time.

elazzabi:
> This "rate limit" error is actually unrelated to the maximum number of concurrent commands allowed - it's a rate limit in Parker designed to prevent abuse, so the wording I suggested is actually wrong. It needs to be something like:
> 
> "Rate limit exceeded: Please wait a moment and try again".
> 
> Aside - is there a way we can prevent copy/pasting a big bunch of commands? Ideally the subshell only allows 1 to be triggered at a time.

I updated the message accordingly. I guess the `line` event is raised whenever a new line is submitted. Not sure if we can block this with some config in `readline` but we can hack it using a global var or something. We can look at this in the future I think.

nickdaugherty (approved):

Issue #293, opened 2019-03-31
Help and Version commands on a WP command display the VIP CLI help and version


dchymko:
We probably want the ability for a user to be able to access the wp cli help (outside of the subshell help)

## Steps to reproduce
`vip --app=68 -- help`
`vip --app=68 -- version`

It appears that you can bypass the default help/version command parsing in the argv library. However, this prevents access to the help and version in the VIP CLI.
https://github.com/Automattic/vip/blob/master/src/lib/cli/command.js#L39
`const options = this.parse( argv, { help: false, version: false } );`

elazzabi:
looks like `help`, `-h` and `--help` display the help screen, I wanted to see if we can configure this, i.e. displaying help only if we pass `-h` or `--help` but looks like it's an all or nothing flag...

Not sure how can we go around this, or if we need to use another package with more flexibility. Something like [commander](https://www.npmjs.com/package/commander)

Edit: less space too

- args (currently used): 31.5kb minified ([bundlephobia](https://bundlephobia.com/result?p=args@5.0.1))
- commander: 10.1kb minified ([bundlephobia](https://bundlephobia.com/result?p=commander@2.19.0))

elazzabi:
Fixed by https://github.com/Automattic/vip/pull/295

PR #295, opened 2019-04-02
Remove help and version in favor of -h/--help and -v/--version

### LINKED ISSUE #162: Help for subcommands
Currently when you pass the `-h` param, you get the help for all top level commands, not specific help for a subcommand:

```
$ vip -h
  Usage: vip [options] [command]

  Commands:
    app      List and modify your VIP Go apps
    help     Display help
    logout   Logout from your current session
    sync     Sync production to a development environment
    version  Display version

  Options:
    -h, --help     Output usage information

$ vip app -h
  Usage: vip [options] [command]

  Commands:
    app      List and modify your VIP Go apps
    help     Display help
    logout   Logout from your current session
    sync     Sync production to a development environment
    version  Display version

  Options:
    -h, --help     Output usage information
```

elazzabi:
_This is a PR to see how the help and version will look like if we remove `help`/`version` and replace it with `-h`/`-v` and `--help`/`--version` flags instead._

This will make the help displays only if `-h` or `--help` are passed, which is the correct behavior. Before this, in order for the help to be shown, you need `-h`, `--help` or `help` but that causes some problems with subcommands see #162.

This fix #162 and #293 

### How to test:

1. Get current branch
2. Run `npm link`
3. Run `vip -h` you should get the help for vip
4. Run `vip app -h` or any subcommand, you should get the help for that subcommand

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -410,8 +410,8 @@ export default function( opts: any ): args {
 	}
 
 	// Add help and version to all subcommands
-	a.option( 'help', 'Output the help for the (sub)command', false );
-	a.option( 'version', 'Output the version number', false );
+	a.option( 'help', 'Output the help for the (sub)command' );
+	a.option( 'version', 'Output the version number' );
 
 	return a;
 }
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -401,6 +401,10 @@ export default function( opts: any ): args {
 		a.option( 'format', 'Format results', 'table' );
 	}
 
+	// Add help and version to all subcommands
+	a.option( 'help', 'Output the help for the (sub)command', false );
+	a.option( 'version', 'Output the version number', false );
+
 	return a;
 }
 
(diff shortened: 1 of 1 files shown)


nickdaugherty (commented):
I personally like this a lot more, and it feels more consistent with other CLI tools I'm used to, besides WP CLI which is backwards anyway (`wp help post`).

Question - where does `(disabled by default)` come from in the help, and can we remove it?

```
$ ./dist/bin/vip.js -h
...
  Options:
    -h, --help     Output the help for the (sub)command (disabled by default)
    -v, --version  Output the version number (disabled by default)
```

nickdaugherty:
Also ping @mjangda and @joshbetz for thoughts as this changes existing behavior.

elazzabi:
@nickdaugherty looks like it was caused by the `false` being used when declaring the options. Fixed by this https://github.com/Automattic/vip/pull/295/commits/e696304ef06058d2dcfde78f48400d8b48a83b43

joshbetz:
Agree, this looks good to me.

nickdaugherty (approved):
🚢

elazzabi:
Tests are passing but timing out, merging

PR #297, opened 2019-04-03
Fix wp subshell cursor

nickdaugherty:
`chalk` was interferring with the display of the cursor - moving the trailing whitespace to outside the template literal fixes it.


elazzabi (approved):
Works as expected 👍

nickdaugherty:
Test failures are only on one version of Node on Windows (always fails) - going to merge.

PR #300, opened 2019-04-03
Hook up input pipe for WP CLI

nickdaugherty:
Adds input piping to WP CLI commands, for things like prompts and `more`.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -122,6 +122,8 @@ commandWrapper( {
 
 			const promptIdentifier = `${ appName }.${ getEnvIdentifier( opts.env ) }`;
 
+			let commandRunning = false;
+
 			subShellRl = readline.createInterface( {
 				input: process.stdin,
 				output: process.stdout,
@@ -132,6 +134,10 @@ commandWrapper( {
 			} );
 
 			subShellRl.on( 'line', async line => {
+				if ( commandRunning ) {
+					return;
+				}
+
 				// Check for exit, like SSH (handles both `exit` and `exit;`)
 				if ( line.startsWith( 'exit' ) ) {
 					subShellRl.close();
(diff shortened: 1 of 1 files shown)


simonwheatley:
I'm testing this on app ID 68. When I enter the pseudoshell and try `wp help` I _do_ get the help text in a `more` or `less` UI. I can move around and quit, however all keystrokes (like using `G` or `g` to navigate the text, and `q` to quit) are recorded against the subshell prompt:

```
$ vip --app=68 -- wp
✔ Which environment? · testing
Welcome to the WP CLI shell for the testing environment of example (example.com)!
example.testing:~$ wp help
example.testing:~$ Gg Gq
```

I then can't delete those characters, and the pseudo shell is unresponsive.

Edit to add: I can quit the pseudo shell with `ctrl_c`, so it's not totally unresponsive.

simonwheatley:
This is somewhat of a step forward, in that one can now see the help text. Tempting to merge and log the rest of my feedback as an issue?

elazzabi:
I do get those chars after quitting, but I can remove them when hitting backspace 🤔 

@nickdaugherty what do you think, should we go even with this? I'll see if I can find a way to remove it

simonwheatley:
Aha! I can't remove these characters with backspace because my invisible cursor is at the start of the string, on character 0. If I use `delete` instead of `backspace` I can remove them.

elazzabi:
after the conversation on Slack: this is probably a problem with the CLI used, as the cursor is showing for a normal CLI

nickdaugherty:
Fixed more of these issues, still not perfect. Maybe good enough to ship then keep iterating, but going to keep working on it today.

elazzabi (approved):
Good to ship, we'll continue iterating afterwards

elazzabi:
Tests passing on Windows, just failing due to timeout

PR #308, opened 2019-04-08
Add environment aliases support

nickdaugherty:
Adds a new way to specify the app + environment, an "alias". Examples:

```
vip @my-environment -- wp option get home
vip @my-environment.production -- wp option get home
vip @my-environment.develop.instance-01 -- wp option get home
```

The previous way to specify app/env, the `--app` and `--env` options, still work for backwards compatibility (but cannot be combined with the new alias approach).

### DIFF
--- a/__tests__/lib/cli/envAlias.js
+++ b/__tests__/lib/cli/envAlias.js
@@ -1,0 +1,195 @@
+/**
+ * Internal dependencies
+ */
+import envAlias from '../../../src/lib/cli/envAlias';
+
+const parseEnvAliasSpy = jest.spyOn( envAlias, 'parseEnvAlias' );
+
+describe( 'utils/cli/envAlias', () => {
+	afterEach( parseEnvAliasSpy.mockClear );
+
+	describe( 'isAlias()', () => {
+		it.each( [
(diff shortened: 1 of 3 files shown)


elazzabi (approved):
code looks good 👍 
tested locally and all looks good 🙌

mjangda:
I'd like to pause on shipping/merging this, if that's okay. It's definitely a nicer approach than `--app + --env` given that we can more easily surface this for copy-pasting via `vip app (list)?` and the Dashboard. 

What I'd like to talk through and come to a decision on first is the whole `environment_name` vs `instance name` issue to make sure we're in agreement around how that should work now and in the future. Maybe something we can discuss on the call on Thursday?

nickdaugherty:
@mjangda as discussed, this will technically be backwards compatible if we decided to drop the idea of "instance names" and just require unique environment types:

Current:

`@some-app.develop.dev-01`

Hypothetical future change (unique env types):

`@some-app.develop-01`

The extra "instance name" piece just wouldn't be used on those sites so I think as long as we're happy with the `@alias` style approach here, we can go with this PR for now.

Issue #309, opened 2019-04-09
WP-CLI: Quoted string value for parameter with spaces misinterpreted


simonwheatley:
```
$ wp post list --posts_per_page=5 --meta_key=byline --meta_compare="NOT EXISTS" 
Error: Too many positional arguments: EXISTS"
```

Ref: https://github.com/wp-cli/wp-cli/issues/1928

via a client 🙂

nickdaugherty:
Have some code for this that works locally, need to work out something on prod still.

nickdaugherty:
This was fixed by https://github.com/Automattic/vip/pull/310 and a change to the runner.

PR #310, opened 2019-04-10
WP CLI: Missing quotes in wp cli commands

nickdaugherty:
We need to re-quote certain parameters before sending them to the API, as double quotes are removed in on the command line when parsed for `process.argv` in Node.

See #309

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -115,13 +115,13 @@ commandWrapper( {
 	envContext: true,
 	appQuery,
 } )
-	.argv( process.argv, async ( arg, opts ) => {
-		const isShellMode = 'shell' === arg[ 0 ];
-		const isSubShell = 0 === arg.length;
-		const cmd = arg.join( ' ' );
+	.argv( process.argv, async ( args, opts ) => {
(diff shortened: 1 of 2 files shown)


nickdaugherty:
There's a server side component that needs adjusted also, but this is ready to go for the `vip` side.

src/bin/vip-wp.js
```diff
@@ -115,13 +115,17 @@ commandWrapper( {
...
+		const isShellMode = 'shell' === args[ 0 ];
+		const isSubShell = 0 === args.length;
+
+		// Have to re-quote anything that needs it before we pass it on
+		const quotedArgs = requoteArgs( args );
+
+		const cmd = quotedArgs.join( ' ' );
 
 		// Store only the first 2 parts of command to avoid recording secrets. Can be tweaked
-		const commandForAnalytics = arg.slice( 0, 2 ).join( ' ' );
```
dchymko, line 124: At some point we could mimic the sanitizeCommands logic here. But probably want to hold off on that until we're sure all the cases are covered

dchymko (approved):
LGTM!

PR #312, opened 2019-04-16
Clone process.argv to avoid overwriting it

nickdaugherty:
If we overwrite it, subcommands fail b/c process.argv no longer has the alias in it when parsed.


mjangda (approved):
👍

PR #320, opened 2019-04-24
Alias the vip User ID with the anonymized id in Tracks

### LINKED ISSUE #316: Tracks: update anonymized ID
When a user first runs `vip`, we assign an anonymized ID. When they log in, we should send the de-anonymize event to Tracks with their VIP user ID (for more useful tracking across CLI/API/Dash). Will need to check with the Tracks team the specs of the event.

dchymko:
On user login, associate the vip user id with the tracks anonymous id. This will allow better tracking of user events across various services.

Fixes #316

### DIFF
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -13,7 +13,7 @@ import { prompt } from 'enquirer';
  */
 import command from 'lib/cli/command';
 import Token from 'lib/token';
-import { trackEvent } from 'lib/tracker';
+import { trackEvent, aliasUser } from 'lib/tracker';
 
 // Config
 const tokenURL = 'https://dashboard.wpvip.com/me/cli/token';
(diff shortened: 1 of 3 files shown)


src/bin/vip.js
```diff
@@ -115,6 +115,9 @@ const rootCmd = async function() {
 			throw e;
 		}
 
+		//de-anonymize user for tracking
+		await aliasUser( token.id );
```
mjangda, line 119: We'll need to update the `uuid` in `lib/token` as well, otherwise we'll keep sending the anon ID with future events.

dchymko: Okay, I think the following commit addresses that.

dchymko:
This appears to be working in tracks correctly. If I login, I see the aliasUser event, and then subsequent events have my userID instead of the anonymized one.

mjangda (approved):
🚢

PR #324, opened 2019-04-29
Send a newline after results when format is count or ids

### LINKED ISSUE #322: Using count formatting the result shows briefly and then vanishes
1. Open a WP-CLI sub-shell: `vip --app=$APP -- wp`
2. Run a post listing with `--format=count`, for example `wp post list --posts_per_page=5 --format=count`
3. Result shows briefly and then is overwritten by the prompt

count-vanishing (1)

Note that the result outputs from WP-CLI with no line break character, so if you run this in a "normal" shell the result is preprended to the next line, e.g. `5user@machine:~$`.

A vanishing Count? 🧛‍♂️

nickdaugherty:
WP-CLI doesn't send a newline on these, so the next prompt() ends up resetting the line.

Fixes #322

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -228,7 +228,12 @@ commandWrapper( {
 
 					// Need a newline - WP CLI doesn't always send one :(
 					// https://github.com/wp-cli/wp-cli/blob/779bdd16025cb718260b35fd2b69ae47ca80cb91/php/WP_CLI/Formatter.php#L129-L141
-					if ( line.includes( '--format=count' ) || line.includes( '--format="count"' ) || line.includes( '--format=ids' ) || line.includes( '--format="ids"' ) ) {
+					if ( line.includes( '--format=count' ) ||
+						line.includes( '--format="count"' ) ||
+						line.includes( '--format=\'count\'' ) ||
+						line.includes( '--format=ids' ) ||
+						line.includes( '--format="ids"' ) ||
+						line.includes( '--format=\'ids\'' ) ) {
 						process.stdout.write( EOL );
 					}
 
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -225,6 +226,12 @@ commandWrapper( {
 
 					commandStreams.stdoutStream.unpipe( process.stdout );
 
+					// Need a newline - WP CLI doesn't always send one :(
+					// https://github.com/wp-cli/wp-cli/blob/779bdd16025cb718260b35fd2b69ae47ca80cb91/php/WP_CLI/Formatter.php#L129-L141
+					if ( line.includes( '--format=count' ) || line.includes( '--format=ids' ) ) {
```
mjangda: Do we care that this won't work if `--format="count"`?

mjangda: Alternatively, can we fix in the API instead?

nickdaugherty: I guess we should handle quotes there too for consistency.

My worry is introducing "non-standard" behavior when piping output to other processes so probably better to leave it "raw" and just fix the interactive experience. Though this does affect one-off commands too (but not as badly).

It _might_ actually need to be fixed at the WP-CLI level ultimately. Need to do some more research.

nickdaugherty: Hmm:

```
wp post list --posts_per_page=10 --format="count"
Error: Parameter errors:
 Invalid value specified for 'format' (Render output in a particular format.)
```

nickdaugherty: Handled quoted versions in 8feb1ef

mjangda: Probably also need `--format='count'` :)

nickdaugherty: Heh, added in b93505f

mjangda (approved):
🚢

PR #325, opened 2019-04-29
Don't kick from wp subshell on error

### LINKED ISSUE #317: Don't kick the user out of the WP-CLI shell when errors occur
```
example.develop:~$ wp post list
Error: `wp post list` must use a `--posts_per_page` parameter with a value less than 1000
$ # prompt is no longer in the WP-CLI shell
```

We shouldn't drop the user out of the WP-CLI shell on error.

nickdaugherty:
When there is an error from the API in the interactive shell mode, we shouldn't kick the user out of the shell. Instead, display the error and re-prompt.

Fixes #317

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -14,7 +14,7 @@ import readline from 'readline';
 /**
  * Internal dependencies
  */
-import API, { API_HOST } from 'lib/api';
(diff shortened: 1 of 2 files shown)


mjangda (approved):
Tested and works 👍

PR #341, opened 2019-05-08
Track install event

joshbetz:
Initially to help identify the install base of each version.


dchymko:
Note that this was reverted today in #441

PR #303, opened 2019-04-05
WP CLI - More comprehensive subshell cursor fix

nickdaugherty:
Resets the cursor at the beginning by sending a control code. This prevents instances where the cursor disappears, usually after `enquirer` prompts.

PR #343, opened 2019-05-09
Only reset the cursor in subshell mode

### LINKED ISSUE #342: Control character breaks for loop
``` bash
$ for POST_ID in $(vip @app.env -- wp post list --posts_per_page=3 --format=ids); do echo $POST_ID; vip @app.env -- wp post get "$POST_ID" --format=json; done;
1807
Error: Could not find the post with ID 0.
1793
{JSONBLAH}
1791
{JSONBLAH}
```

Looking in the command logs, we see this:

```
wp post get [?25h1807 --format=json
```

@joshbetz identified this code: 

https://github.com/Automattic/vip/blob/a51be47fb5e46a6039f5473854475bfd195f55b1/src/bin/vip-wp.js#L139

joshbetz:
In subshell mode, we need to reset the cursor as it can get messed
up by the environment prompts. It shouldn't be necessary outside
the subshell mode though. And, in fact, if you're using the output
in other commands, it can cause unexpected behavior.

Fixes: #342

Previously: #303

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -135,10 +135,10 @@ commandWrapper( {
 		let rl;
 		let subShellRl;
 
-		// Reset the cursor (can get messed up with enquirer)
-		process.stdout.write( '\u001b[?25h' );
-
 		if ( isSubShell ) {
+			// Reset the cursor (can get messed up with enquirer)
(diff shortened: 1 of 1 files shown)


nickdaugherty (approved):
LGTM

PR #349, opened 2019-05-16
Include --help for unknown command (fixes #348 )

### LINKED ISSUE #348: Update help text on unknown command.
Since #82  when you type in an unknown command we show:
```
 Error: `help` is not a valid subcommand. See `vip help`
```

I just ran into the issue where I tried to type `vip help` as the command suggest but got an error. 
```
$ vip help
Error: `help` is not a valid subcommand. See `vip help`
```

The simplist solution here is to change the text to read `See --help` .

BrookeDot:
Provides a more verbose message when an unknown command is run. Instead of `see vip help` the message will now read `vip --help`.

fixes #348

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -120,7 +120,7 @@ args.argv = async function( argv, cb ): Promise<any> {
 			error: `Invalid subcommand: ${ subcommand }`,
 		} );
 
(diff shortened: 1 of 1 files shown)


mjangda (approved):
🚢

PR #353, opened 2019-05-20
Add --yes option to wp-cli command

joshbetz:
Skips the confirmation prompt in production environments. This can
be useful when you're trying to loop over a list of sites and don't
want to have to manually confirm each site.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -116,6 +116,7 @@ commandWrapper( {
 	envContext: true,
 	appQuery,
 } )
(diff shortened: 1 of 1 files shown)


mjangda (approved):
Works great!

```
$ ./dist/bin/vip-wp.js @123 -- option get home
+ command: wp option get home
✔ Are you sure you want to run this command on PRODUCTION for site example-com (123)? (y/N) · true
https://example.com
$ ./dist/bin/vip-wp.js @123 --yes -- option get home
https://example.com
$ ./dist/bin/vip-wp.js @123 -y -- option get home
https://example.com
```

PR #358, opened 2019-05-24
WP-CLI: Close underlying socket.io connection to prevent reconnects

nickdaugherty:
We're currently not properly closing the connection when commands end in subshell, so they keep trying to reconnect after the command is complete.

This can manifest as several "rate limit exceeded" errors if a user has run several commands in the subshell and their internet connection drops/reconnects - all the orphaned socket.io connections attempt to reconnect at the same time, causing the rate limit error message.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -107,7 +107,7 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 		console.log( err );
 	} );
 
-	return { stdinStream, stdoutStream };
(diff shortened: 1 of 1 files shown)


nickdaugherty:
Side-note - we can thank cats for turning off my power strip / wifi which caused a bunch of the rate limit errors when it finally reconnected, leading me to realize the connections were still open 😹

simonwheatley:
Amazing!

elazzabi (approved):
looks good!

Thank you for working on this 👍

Issue #357, opened 2019-05-23
WP CLI: Interactive commands in subshell "hang" after first prompt


dchymko:
## Description
Commands that have multiple prompts (called via --prompt) do not work as expected in the subshell

## Steps to reproduce
```
$ vip wp --app <site id>
Welcome to the WP CLI shell for the develop environment of develop (example.com)!
develop.develop:~$ wp user create --prompt
1/15 <user-login>: 
```
"hangs" at this point.

## Workaround
Call the command outside of the subshell
 `vip <@siteid>-- wp user create --prompt`

dchymko:
I tried running ` wp user create --prompt` outside of the shell ( #357) and it it hangs after entering data for the first prompt. 
```
./dist/bin/vip-wp.js --app=1163 -- user create  --prompt
1/15 <user-login>: testuser
<hang>
```

It works as expected with the current VIP command

```
vip --app=1163 -- wp user create --prompt
1/15 <user-login>: testuser
2/15 <user-email>: email@domain.tld
...
```

joshbetz:
Sounds like the second input hangs in general #464

dchymko:
I don't think the two are related unfortunately. In 464, the CLI wasn't sending any data to the socket after the first command. In this case, the data is being sent but for some reason the output isn't coming back.

running `DEBUG=* ./dist/bin/vip-wp.js` was super helpful in seeing whats going on behind the scenes, (although the socket code is super chatty and outputs a lot of debug stuff!)

PR #364, opened 2019-06-06
Always use readline to parse input

### LINKED ISSUE #363: WP-CLI: Fix stdin pipe for `less` and similar commands
If you run `wp help` or anything where the output is longer than the current window, `less` kicks in so you can page through the results. You should be able to use standard vim keybindings (j, k, u, d, spacebar, etc) to scroll. The problem is that we don't submit those characters until you press return. The result is that we end up with scrolling not working quite correctly and disjointed output because of the newlines generated by return.

This is actually not an issue in the subshell experience though. Since we're using `readline` to capture stdin, all the input is captured and sent to the remote command automatically.

My proposal is to always use readline. The subshell experience would continue to work the same way it does now. If you want to run a single command, we'd use readline to capture stdin, and obviously stop capturing input when the command exits. In many cases we'd start capturing input and immediately stop, which is fine. In cases where the command is waiting for input, readline would handle that and pipe it to the web socket, and on to the CLI container.

joshbetz:
This ensures we always use readline to parse input and therefore
don't rely on bash to decide when input is worthy of being sent
to stdin (i.e. on return).

This effectively makes everything subshell mode, though if you run a
specific command, we return to your system shell as soon as the command
finishes.

Noting that this also removes the special "wp shell" mode as well, since the
normal mode is now able to support all input. We could add that back
if there are other reasons to have it.

The primary thing this fixes is `wp help` or other similar long running
commands that expect input over time. Previously input was only sent
when the user pressed return, but now we're able to capture all input.

Fixes #363

## Testing

Some simple test cases:

* Run `wp help`, verify scrolling works correctly
* Run `wp shell`, verify shell works correctly
* Run `wp`, verify subshell mode still works
* Test invalid commands

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -262,7 +262,8 @@ commandWrapper( {
 
 		if ( ! isSubShell ) {
 			subShellRl.write( 'wp ' + cmd + '\n' );
-		} else {
-			subShellRl.prompt();
+			return;
 		}
+
+		subShellRl.prompt();
 	} );
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -11,6 +11,7 @@ import SocketIO from 'socket.io-client';
 import IOStream from 'socket.io-stream';
 import readline from 'readline';
 import { EOL } from 'os';
+import { Writable } from 'stream';
 
 /**
  * Internal dependencies
@@ -159,9 +160,21 @@ commandWrapper( {
 
 		let commandRunning = false;
 
+		const mutableStdout = new Writable( {
+			write: function( chunk, encoding, callback ) {
+				if ( ! this.muted ) {
+					process.stdout.write( chunk, encoding );
+				}
+
+				callback();
+			},
+		} );
+
+		mutableStdout.muted = false;
+
 		const subShellSettings = {
 			input: process.stdin,
-			output: process.stdout,
+			output: mutableStdout,
 			terminal: true,
 			prompt: '',
 			historySize: 0,
@@ -281,7 +294,9 @@ commandWrapper( {
 		} );
 
 		if ( ! isSubShell ) {
+			mutableStdout.muted = true;
 			subShellRl.write( `wp ${ cmd }\n` );
+			mutableStdout.muted = false;
 			return;
 		}
 
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -215,8 +215,12 @@ commandWrapper( {
 					console.log( e );
 				}
 
-				subShellRl.prompt();
+				if ( ! isSubShell ) {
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -132,227 +130,138 @@ commandWrapper( {
...
+					line.includes( '--format=ids' ) ||
+					line.includes( '--format="ids"' ) ||
+					line.includes( '--format=\'ids\'' ) ) {
+					process.stdout.write( EOL );
+				}
 
-		commandStreams.stdoutStream.on( 'error', err => {
-			// TODO handle this better
-			console.log( err );
+				if ( ! isSubShell ) {
```
joshbetz, line 281: If this is a single command, we exit here instead of prompting again.

src/bin/vip-wp.js
```diff
@@ -132,227 +130,138 @@ commandWrapper( {
...
-		commandStreams.stdoutStream.on( 'end', () => {
+		subShellRl.prompt();
+
+		subShellRl.on( 'SIGINT', () => {
+			subShellRl.close();
 			process.exit();
 		} );
+
+		if ( ! isSubShell ) {
+			subShellRl.write( 'wp ' + cmd + '\n' );
```
joshbetz: If this is a single command, we write the command to the prompt.

nickdaugherty (commented):
Hmm, seeing some weird behavior testing this out:

Command echoed back to console:

```
$ ./dist/bin/vip-wp.js @297.develop option get home
wp option get home
https://vip-test-develop.go-vip.co
```

CLI hangs waiting for input on error:

```
$ ./dist/bin/vip-wp.js @297.develop unsupported-command
wp unsupported-command
Error: The 'wp unsupported-command' command is not permitted by VIP users at this time

```

joshbetz:
> CLI hangs waiting for input on error:

Fixed in 26c0fc7.

> Command echoed back to console:

Fixed in 6dda2f9

joshbetz:
@nickdaugherty @elazzabi any other concerns here?

elazzabi (approved):
Tried this with subshell and a single line command, works as expected!

nickdaugherty:
Can we release a new dev version that includes this?

joshbetz:
Yeah, for sure. I'm working through some renovate bot dependency updates and then will do a dev release.

dchymko:
Hmm, I thought I had posted a comment on here about it breaking the `--prompt` flow. Previously, --prompt would only work outside of subshell mode, now it doesn't:
```
./dist/bin/vip-wp.js --app=1163 -- user create --prompt
1/15 <user-login>: newuser
<hangs>
```

This did work before. Not sure how much of a priority this is. We can always fix in the original issue (#357)

PR #375, opened 2019-06-13
Catch app/environment prompt errors on SIGINT

joshbetz:
If you SIGINT in an app or environment list, the prompt throws an error.
We need to catch that error or else our uncaught error handler kicks in.

Fixes an uncaught error:

```
  undefined
```

Alternatively, we could update the uncaught error handler to silently exit if the
error is undefined.

## Testing

1. Checkout this PR
1. npm run build
1. `vip wp`
1. CTRL-C
1. Observe there is no error

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -190,7 +190,7 @@ args.argv = async function( argv, cb ): Promise<any> {
 					process.exit();
 				}
 
-				if ( ! err.message ){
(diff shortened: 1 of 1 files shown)


elazzabi (dismissed):
Looks good.

A small enhancement, we usually use `chalk.red( 'Error: ') + error message`, I think it's better if we can keep the same styling everywhere

PR #409, opened 2019-07-08
Run terminate handler when command is finished or cancelled

### LINKED ISSUE #416: Handle CTRL-C gracefully in WP CLI
## Description

We need to handle when a user cancels a long-running WP CLI command

- Forward the Ctrl-C character (x03) to the command runner 
- Call an endpoint in the API to mark the command as cancelled (and do any other cleanup)

dchymko:
## Description
When a command is prematurely terminated (via CTRL-C), we need to run cleanup tasks such as notifying the command runner to stop the commands.

## Steps to Test
1. Pull down PR and related API PR
2. Run `wp user create --prompt` and wait for the first prompt
3. Ctrl-C to exit
4. Verify it was cancelled (in the app dashboard)
5. Repeat for `wp shell`
6. Repeat for the psuedo shell
7. Try a few "long-running" commands (may be difficult to hit the timing on small sites) `wp option list` `wp jetpack sync start`

Note that for commands with a lot of output (`wp option list` for example), the command may still stream output text for a second or two after it completes. So if you cancel it at that point, it will show as 'completed' since it did actually finish running.

Closes #416


src/bin/vip-wp.js
```diff
@@ -289,9 +299,14 @@ commandWrapper( {
...
-		subShellRl.on( 'SIGINT', () => {
-			subShellRl.close();
-			process.exit();
+		subShellRl.on( 'SIGINT', async () => {
+			await trackEvent( 'wpcli_cancel_command', {
+				command: commandForAnalytics,
+			} );
+
+			//write out CTRL-C/SIGINT
+			subShellRl.write( '\x03\n' );
```
darthhexx: Please just send the '\x03' character, without the newline, and you will be in business :) The runner is now expecting a data packet of length 1 with just the value of 0x3.

src/bin/vip-wp.js
```diff
@@ -111,6 +134,28 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 	return { stdinStream, stdoutStream, socket };
 };
 
+const shutdownHandler = async ( terminateRunningCommand, guid ) => {
```
elazzabi: can we make the signature more explicit here? The difference between `shutdownHandler( true, guid )` and `shutdownHandler( false, guid )` isn't explicit until you visit the function definition.

we can opt for: `shutdownHandler( guid, { exitProcess: true } )` or something similar

src/bin/vip-wp.js
```diff
@@ -280,7 +329,7 @@ commandWrapper( {
 
 				if ( ! isSubShell ) {
 					subShellRl.close();
-					process.exit();
+					shutdownHandler( false );
```
elazzabi: I guess leaving this `process.exit();` is better than wrapping it with `shutdownHandler( false );`. Especially if we aren't adding anything in `shutdownHandler` other than calling `process.exit();` inside (which is the current case when passing `false`)

src/bin/vip-wp.js
```diff
@@ -131,6 +176,8 @@ commandWrapper( {
 		const { id: appId, name: appName } = opts.app;
 		const { id: envId, type: envName } = opts.env;
 
+		let cmdGuid;
```
elazzabi, line 178: I think we can use `cliCommand.guid` directly 🤔

elazzabi (changes requested):
Looks good, with some minor changes.

+ We should make sure we remove the `console.log` before merging

dchymko:
Updated the test steps with some examples.

src/bin/vip-wp.js
dchymko: cliCommand is defined inside the `subShellRl.on( 'line', async line => {` event handler, but isn't available in the `subShellRl.on( 'SIGINT', () => {` handler,  hence setting a variable outside of that.

src/bin/vip-wp.js
dchymko: Given your other comment about leaving the process.exit() I just removed that additional parameter from the method.  I had envisioned using the shutdownHandler in other cases and was trying to make it more generic.  If we need to use it in multiple places, we can revisit, but for now simplification is better.

elazzabi (approved):

elazzabi:
Thank you for working on this @dchymko ! Looks good to me (minus the `console.log`s), I'll run some tests and report back

elazzabi:
Tested locally with all cases in the test steps, looks good!

dchymko:
Thanks for testing @elazzabi. The console.logs that are currently there are intentional and match other flows we have. I assume you were referring to the one removed in 2113c81?

I'll hold off merging this until the endpoint is in Automattic/vip-go-api-public#653

dchymko:
Thanks for testing @elazzabi. The console.logs that are currently there are intentional and match other flows we have. I assume you were referring to the one removed in 2113c81?

I'll hold off merging this until the endpoint has been merged in.

elazzabi:
@dchymko oh yes, I was referring to those. All looks good 👍

PR #410, opened 2019-07-11
Reconnect to job if disconnected

### LINKED ISSUE #417: Reconnect to job if disconnected
## Expected/Desired Behavior

After disconnections, commands should reconnect silently for the user

## Actual Behavior

Not working ^

elazzabi:
## Description

This fix the problem with reconnections

There are still some bugs:
- When disconnected, an error is written to the console, do we want this?
- Another bug is in the test steps bellow

## Steps to Test

Example:

1. Check out PR
2. Make sure all APIs are up with all necessary things to make WP-CLI work in localhost
3. Run `npm link`
4. Run `API_HOST=http://localhost:4000 vip wp shell`
5. Interact with the shell
6. Cause the public API to reload by changing something on the code (can be a console log or something)
7. You'll see an error `Error: Connection aborted` alongside a `lost connection` <- this is added just for dev purpose and should be removed once we are ready
8. Once the API is up again, you'll see a `reconnected` in the console, this also is added just for dev purpose and will be removed afterwards
9. Continue your interaction with the shell
10. You'll receive an `invalid command, please pass a valid WP CLI command.` error. **This is a bug** we still need to see why it's raising. It looks like it tries to send another request even if we are in shell mode.

Fix #417


nickdaugherty (commented):
This is cool - a few notes that come to mind:

* Of course we'll have to wait to merge until the CLI Runner supports the re-connection stuff (does it now?)
* Can we abstract the pipe/unpipe stuff into a reusable function?

elazzabi:
@nickdaugherty 
> Of course we'll have to wait to merge until the CLI Runner supports the re-connection stuff (does it now?)

I tested with the `wp shell` and it does reconnect. Because each time we send a request, it's treated and the result is sent, and we wait for another request. For long running commands where the piping is always going on, I think we need the backend parts to make it work.

> Can we abstract the pipe/unpipe stuff into a reusable function?

yep, we should definitely do this.

nickdaugherty:
> I tested with the wp shell and it does reconnect.

That's because it's creating a _new_ shell command. Unless the CLI Runner already has reconnect support built-in, each time we emit the `cmd` event, we're spawning a new process on the CLI Runner side.

elazzabi:
@nickdaugherty I guess, if we want to test this locally too, we want to be sure that the reconnection logic is mocked/working as production?

elazzabi:
@nickdaugherty I moved the pipe/unpipe stuff to its own function. I want to start working on the offset, and planning to make it on another PR to separate concerns.

We can keep both open until we are satisfied of the result before merging, or we can merge this one if we are good.

joshbetz:
We'll also need to address a similar issue in the API, right? If the API is reloaded, it needs to re-connect to CLI containers.

> We can keep both open until we are satisfied of the result before merging, or we can merge this one if we are good.

If this works and the only thing that's missing is the offset work, I'd say we should merge it.

elazzabi:
> We'll also need to address a similar issue in the API, right? If the API is reloaded, it needs to re-connect to CLI containers.

Yes, the easiest way is to let the client initiate the API reconnection to the container. So, if the API is reconnected, it will not initiate any reconnection. When a client is reconnected, the API will receive the request and then reconnect to the proper CLI container. So it's initiated by the client not the API itself.

elazzabi:
> If this works and the only thing that's missing is the offset work, I'd say we should merge it.

I think it's working (minus the console.log I added for debugging). But unless we have this supported on the API, we can't be sure it's 100% working 👀

joshbetz:
> When a client is reconnected, the API will receive the request and then reconnect to the proper CLI container. So it's initiated by the client not the API itself.

Got it, that makes sense.

nickdaugherty:
Cool - if the CLI Runner (the golang server) isn't yet setup to support reconnections to the same command, we can't merge this b/c we'll end up spawning the same process twice if a user's connection hiccups.

@darthhexx are we safe to reconnect to an existing command (even if there is no resume / replay support yet?) Want to be sure we don't spawn multiple instances of the same command (still sends the unique GUID).

darthhexx:
It's not safe to reconnect until the resume support is complete and deployed. Right now the GUID is only used to determine the log file name for the session. I have some time allocated to WP CLI runner development this week though, so it should be live soon.

joshbetz:
@darthhexx what's the status of resume support? No rush, I just want to make sure we plan the work appropriately on our side.

darthhexx:
The Remote Runner PR has been updated with all the new functionality.

simonwheatley:
This PR fixes PLAT-99

elazzabi:
Closing this as it's been part of https://github.com/Automattic/vip/pull/420 recently merged

Issue #418, opened 2019-07-16
Calculate and send and send an offset for WP-CLI commands


elazzabi:
# Expected/Desired Behavior

When a command reconnects to the API, we should continue piping from the last point. We need to keep track of how much data we received and send an offset once reconnected to the API.

## Actual Behavior

We still don't have this.

joshbetz:
We should be careful not to implement too much of the reconnect logic in the CLI in case we want to run CLI commands from other places in the future. Some potential options might be a browser extension or directly in the dashboard.

joshbetz:
This is where the issue was originally created: 643 on the API

elazzabi:
Fixed https://github.com/Automattic/vip/pull/420

PR #420, opened 2019-07-17
Add offset handling for WP CLI commands

elazzabi:
## Description

This calculates and send an offset when reconnecting to WP CLI commands.

## Steps to Test

Nothing to test now, waiting for all parts to be ready.


elazzabi:
This contains lot of changes from #410 too. So should be easier once that one is merged

src/bin/vip-wp.js
```diff
@@ -94,6 +112,10 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 		console.log( 'There was an error with the authentication:', err.message );
 	} );
 
+	socket.on( 'disconnect', err => {
+		console.log( 'lost connection' );
```
dchymko: We may want to differentiate between a dropped connection and a command that ended. Currently this will appear after every successful command completes.

joshbetz: We should also send these to stderr (console.error) so they don't get streamed to the command output. If someone is running an export, we don't want to inject our error messages into their export file.

joshbetz:
Noting that this is working now. We should clean up the `console.log()` messages and then I think this can be merged.

src/bin/vip-wp.js
elazzabi: those were added for debug purpose, but as you suggested we should log to the stderr, and have a differentiation between disconnect because the command is done, and disconnect because a connection is dropped or something.

Looking at this [list of events](https://stackoverflow.com/questions/24224287/list-of-socket-io-events), it looks like we can use `connect_error`.

What do you think @dchymko ?

elazzabi: I ended up using `reconnect_attempt` with a `There was an error connecting to the server. Retrying...` message in the stderr

elazzabi:
@joshbetz @dchymko I made some final changes to remove console logs, and improve error logging. Can you take a look for any missing info and run a final test when you have some time?

dchymko (approved):
Tested, LGTM. One nice to have would be to not dump the stack trace on a disconnect but I don't think that's critical.

It also seems that the command cancellation seems a bit wonky now. Did you notice that? I can take a look after his is merged.

elazzabi:
@dchymko yeah, this stops working if the `reconnect` event is fired.

I think it's because this handler is no longer pointing to the right stream https://github.com/Automattic/vip/pull/420/files#diff-6bf77c3843951e6b92795206746dd5ceR326

We should probably point it to the right one when reconnected. Will look into it in another PR if you didn't already started.

elazzabi:
fix #417 
fix #418

PR #429, opened 2019-07-19
Reverts https://github.com/Automattic/vip/pull/324

### LINKED ISSUE #413: Additional trailing characters (breaks shell helpers)
```
VIPVIPVIP$ for POST_ID in $(vip @testgutenberg.production --yes -- wp post list --posts_per_page=1 --format=ids); do echo "Post ID: $POST_ID"; vip @testgutenberg.production --yes -- wp post get "$POST_ID" --format=json; done;
Post ID: 954
Error: invalid command, please pass a valid WP CLI command.
+-----------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------+
| Field                 | Value                                                                                                                                                     |
+-----------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------+
<snip />
```

When I pipe this straight to the pasteboard, the output of the command is:

```
954

```

Note the additional trailing linebreaks, which I think we shouldn't have.

Trailing characters also added in pseudo-shell:

```
VIPVIPVIP$ vip @testgutenberg.production wp
Welcome to the WP CLI shell for the PRODUCTION environment of testgutenberg (testgutenberg.com)!
testgutenberg.production:~$ wp post list --posts_per_page=1 --format=ids
954

testgutenberg.production:~$

joshbetz:
Fixes #413

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -10,7 +10,6 @@ import { stdout } from 'single-line-log';
 import SocketIO from 'socket.io-client';
 import IOStream from 'socket.io-stream';
 import readline from 'readline';
-import { EOL } from 'os';
 import { Writable } from 'stream';
 
 /**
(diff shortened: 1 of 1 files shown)


joshbetz:
@nickdaugherty in that PR, you mentioned:

> the next prompt() ends up resetting the line

Do you remember what you mean by "resetting"? I think WP-CLI forgoes the newline on purpose so you can pipe the results to another command. This PR seems to work for me, but I want to make sure we're not introducing a regression here.

nickdaugherty:
> Do you remember what you mean by "resetting"?

I believe I meant that it clears it out, see Simon's screencast here to see what was fixed:

https://github.com/Automattic/vip/issues/322

joshbetz:
Ok cool. That example still seems to work with this PR. We've refactored a bit since then, which probably explains why it works now.

elazzabi (approved):
thank you! 🙏 

Looks good to me

PR #435, opened 2019-07-24
WP CLI: Don't call cancelCommand endpoint when command is canceled

dchymko:
## Description
I moved the command cancellation logic into the API by detecting the Ctrl-C character there. This should simplify the flow and provide better consistency.

## Steps to Test
Same as #409 for the basic cancellation
Also added a check for a 2nd CTRL-C to immediately exit (in case for some reason the socket doesn't close and the process.exit() never runs). This one may be difficult to test without adding in a timeout line after:
https://github.com/Automattic/vip/blob/master/src/bin/vip-wp.js#L310

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -133,28 +133,6 @@ const launchCommandAndGetStreams = async ( { guid, inputToken } ) => {
 	return { stdinStream, stdoutStream, socket };
 };
 
-const shutdownHandler = async ( guid ) => {
-	try {
-		if ( guid ) {
-			try {
-				const val = await cancelCommand( guid );
-			} catch ( e ) {
-				// If this was a GraphQL error, print that to the message to the line
-				if ( e.graphQLErrors ) {
-					e.graphQLErrors.forEach( error => {
-						console.log( chalk.red( 'Error:' ), error.message );
-					} );
-				} else {
-					// Else, other type of error, just dump it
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -327,17 +307,19 @@ commandWrapper( {
 			} );
 		} );
 
-		subShellRl.on( 'SIGINT', () => {
-			// Handler here can't be async so all methods need to fire and forget
+		subShellRl.on( 'SIGINT', async () => {
+			//if we have a 2nd SIGINT, exit immediately
```
elazzabi, line 311: If there are 2 SIGINT events, we'll still be sending it to the API, but just exiting right away in the client side, right? 🤔

dchymko: I've added a return to make sure it doesn't run the rest for whatever reason.  We don't send a second ctrl-C char (but we could). I think on the 2nd try we want to do the bare minimum because of for some reason it errors or skips it the 2nd time, we don't want the user trapped in the command with no way to get out. This is very much an edge case though. It was actually more for my testing purposes when I was reworking some of the sequence and couldn't get out :)

elazzabi: Hmmm we definitely don't want to block a user without a way out :). Let's leave it here then.

elazzabi (approved):
Looks good 👍

PR #442, opened 2019-07-25
Add the postinstall trackEvent back in

dchymko:
## Description
Adds the postinstall step back in that was revered in #441  

## Steps to Test
Test packaged install:
1. Run `npm pack`
2. Run `npm install -g automattic-vip-1.6.2.tgz` and confirm there are no install errors
3. Might be a good idea to confirm that the track event was sent

Test cloned install: **(this currently does not work)**
1. git clone in a new dir and switch to this branch
2. `npm i` should run without errors


elazzabi (changes requested):
As noted on Slack, what are we trying to achieve with tracking when a new version is installed?

dchymko:
>As noted on Slack, what are we trying to achieve with tracking when a new version is installed?

I think the idea was that we could see how many people were using each version. Probably to see uptake and to also know the scope of users affected if we have to break compatibility down the road for (much) older versions?

dchymko:
Noting that removing the `npm build` from the postinstall step has broken the cloned version install.  Thinking about this a bit more, maybe we don't really care about tracking when it's installed via `git clone` vs `npm i -g`? We probably don't to track these types of "installs", since things like the CI services clone it then run `npm i`

elazzabi:
I think we can use other events such as https://github.com/Automattic/vip/blob/master/src/lib/cli/command.js#L407 to detect what versions are used by clients. If we add a tracking event to install, and use it as a way of detecting version usage, we may end up with outdated info as this event will only fire once (when installing). In all other events, we do track which version is used, so I think any event can be used to have an idea about version usage, and those will be newer than an install event in my opinion.

joshbetz:
cc @simonwheatley for input on whether we need to track the install event.

simonwheatley:
I’m happy that we can use the version property on all events to track this. Assuming that this PR affects only the Tracks event post-install it can be closed unmerged @dchymko

elazzabi:
Thank you for confirming Simon. Closing this!

PR #448, opened 2019-08-07
Re-attach outputStream socket events when reconnecting

dchymko:
## Description
We need to re-attach the error and end events to the new output stream when reconnecting so that the command exits when it ends and is cancelled

There are three use cases here:
- [x] Command disconnects, reconnects and then completes but doesn't exit at end
- [x] Command diconnects, doesn't reach timeout before Ctrl-C pressed
- [x] Command disconnect hit timeout and tries to reconnect. Ctrl-C pressed after his happens

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-wp.js --app=<testSiteId> --yes -- vip test-command long`
4. Disconnect network and wait for connection to timeout.
5. Ctrl-C should end and exit back to the command prompt
6. Repeat but but hit Ctrl-C after the command starts but before the reconnect attempt is made
7. Repeat but wait for the command to complete. Should return to the command prompt.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -360,6 +360,7 @@ commandWrapper( {
 
 			//write out CTRL-C/SIGINT
 			process.stdin.write( cancelCommandChar );
+			currentJob.stdoutStream.end();
 			trackEvent( 'wpcli_cancel_command', {
 				command: commandForAnalytics,
 			} );
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -49,6 +49,36 @@ const unpipeStreamsFromProcess = ( { stdin, stdout: outStream } ) => {
 	outStream.unpipe( process.stdout );
 };
 
+const bindStreamEvents = ( { subShellRl, commandRunning, isSubShell, stdoutStream } ) => {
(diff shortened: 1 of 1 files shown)


joshbetz:
> Ctrl-C should end and exit back to the command prompt

Repeatedly pressing Ctrl-C doesn't bring me back to the prompt until the network connection is re-established. I get stuck at:

```
There was an error connecting to the server. Retrying...
```

dchymko:
It looks like there are 3 use cases here (I added them to the description). I fixed the 2nd one, but I think the one you are seeing is because we call subShellRl.pause() without ever getting to the resume() (behind an `await` which errors on timeout)
https://github.com/Automattic/vip/blob/6a128c50c3176a6f3d4a7592c380a58b767a4b9b/src/bin/vip-wp.js#L294

I'm not sure how to solve this. We might have to put the SIGINT somewhere else (on the process maybe?)

dchymko:
Turns out that the input stream was closed when the connection failed, which is why the SIGINT wasn't being received.  Added a fix in 999c3ee to address that by reconnecting a new stream.

joshbetz (dismissed):

elazzabi (dismissed):

PR #449, opened 2019-08-09
await tracking events

elazzabi:
## Description

We used the fire and forget approach for those tracking events. However, some aren't recorded as the process is already closed with no time to send the events

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -338,11 +338,11 @@ commandWrapper( {
 				subShellRl.resume();
 			} );
 
-			currentJob.stdoutStream.on( 'end', () => {
+			currentJob.stdoutStream.on( 'end', async () => {
 				subShellRl.clearLine();
 				commandRunning = false;
 
-				trackEvent( 'wpcli_command_end', commonTrackingParams );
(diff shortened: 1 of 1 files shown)


joshbetz:
Minor, but what if we await them at the end? This could look like:

```javascript
const tracksEvents = [];

// Do some stuff

tracksEvents.push( trackEvent( 'wpcli_command_execute', commonTrackingParams ) );

// Do some other stuff

await Promise.all( tracksEvents );
```

The advantage here is that we don't wait for them unless we need to.

elazzabi:
@joshbetz When I added this, my thinking was: lets `await` only the events that needs to be awaited, ie those appearing just before the closing of the process: `command_cancel` and `command_end`. For others, they still have time to report, so they aren't awaited, fired and forget about.

What do you think? 🤔

joshbetz:
@elazzabi probably fine for now and definitely simpler to implement. If we notice problems in the future, we can do what I've suggested.

joshbetz (dismissed):

elazzabi:
Thank you for the review Josh. This branch was behind master and had some merging conflicts. Merging now

PR #452, opened 2019-08-14
Add is_vip flag

elazzabi:
## Description

This adds the `is_vip` flag to all tracked events. Apollo will cache the user request so calling it multiple time shouldn't be a problem *in theory*.

However, I think we can leverage other options if we can:
1. Can we get the `is_vip` flag or user information part of the Token class? If yes, we may use it instead of this
2. Do we have a way to call the API and pass the result to the underlying commands? With this we may call the API once, but the problem is we need to add `is_vip` flag to all events in each sub command.

### DIFF
--- a/src/lib/api/user.js
+++ b/src/lib/api/user.js
@@ -24,9 +24,9 @@ export default async function(): Promise<any> {
 			}`,
 		} );
 
-	if ( ! res || ! res.data ) {
+	if ( ! res || ! res.data || ! res.data.me ) {
 		return {};
 	}
 
-	return res.data;
+	return res.data.me;
 }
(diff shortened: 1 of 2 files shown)


src/lib/api/user.js
```diff
@@ -0,0 +1,32 @@
...
+					isVIP
+				}
+			}`,
+		} );
+
+	if ( ! res || ! res.data ) {
+		return {};
+	}
+
+	return res.data;
```
joshbetz: Should we simplify this by returning `res.data.me`? Then above we could just check `user.isVIP`.

elazzabi: yeah you're right, thank you 🙏 

fixed!

joshbetz (approved):

PR #453, opened 2019-08-16
WP CLI: Fix single quotes bug

### LINKED ISSUE #365: Single quotes in WP-CLI subshell mode not interpreted correctly
Subshell Command:

```
some-site.testing:~$ wp option update some-option '[12345]' --format=json
Error: Invalid JSON: '[152352941]'
```

But a plain command works (non-subshell):

```
vip @<site-id> -- wp option update some-option '[12345]' --format=json
```

Then interestingly, double quotes behaves still differently and doesn't save it as JSON at all (though perhaps unrelated to this specific issue):

```
some-site.testing:~$ wp option update some-option "[12345]" --format=json
Success: Updated 'some-option' option.
some-site.testing:~$ wp option get some-option --format=json
"[152352941]"
```

pandah3:
## Description

Fixing a bug where single quote arguments are not interpreted properly in VIP CLI.
Verifying that the requoted command gets passed to the API.
WIP

Fixes #365 

## Steps to Test

WIP

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-wp.js`
4. Run a command that involves single quotes (e.g.- `wp option update blogdescription 'Single quotes'`
5. Verify there's no error

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,16 +10,16 @@
       "integrity": "sha512-UHI+7pHv/tk9g6WXQKYz+kmXTI77YtuY3vqC59KIqcoWEjsJJSG6rAxKaLsgj3LDyadsPrCB929gVOKM6Hui0w==",
       "dev": true,
       "requires": {
-        "chokidar": "2.1.5",
-        "commander": "2.19.0",
-        "convert-source-map": "1.5.1",
-        "fs-readdir-recursive": "1.1.0",
-        "glob": "7.1.2",
-        "lodash": "4.17.15",
-        "mkdirp": "0.5.1",
-        "output-file-sync": "2.0.1",
-        "slash": "2.0.0",
-        "source-map": "0.5.7"
+        "chokidar": "^2.0.4",
+        "commander": "^2.8.1",
+        "convert-source-map": "^1.1.0",
+        "fs-readdir-recursive": "^1.1.0",
+        "glob": "^7.0.0",
+        "lodash": "^4.17.13",
+        "mkdirp": "^0.5.1",
+        "output-file-sync": "^2.0.0",
+        "slash": "^2.0.0",
+        "source-map": "^0.5.0"
       },
       "dependencies": {
         "lodash": {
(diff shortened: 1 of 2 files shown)


joshbetz:
Is there more context for why we need to replace single quotes with double quotes? Initially my concern is they're not quite functionally equivalent, which could lead to unexpected bugs.

For example, what happens if we do something like this?

```
wp option update blogdescription '$FOO'
```

pandah3:
It seems like the Go runner isn't running the commands through bash so variables won't be evaluated even if it's in double quotes:
[https://github.com/Automattic/Cron-Control/pull/176/files#diff-8c7f033f212cbb3c6b1ec76074b914cfR598](url)

I was replacing double quotes with single quotes because that's what the CLI runner is taking, if I'm not mistaken: 
[https://github.com/Automattic/Cron-Control/pull/176/files#diff-8c7f033f212cbb3c6b1ec76074b914cfR311](url)

I agree that having some distinction between single quotes and double quotes would be good, but that's probably something that needs to be fixed in the runner.

joshbetz:
@pandah3 awesome, thanks for those details. I think it would be helpful to add an inline comment to that effect where we're replacing the quotes.

joshbetz:
It looks like the quotes get passed into the argument:

```
$ vip @59 wp option update blogdescription "Hello world"
Success: Updated 'blogdescription' option.

$ vip @59 wp option get blogdescription
&quot;Hello world&quot;
```

The same thing also happens if I use single quotes:

```
$ vip @59 wp option update blogdescription 'Hello world'
Success: Updated 'blogdescription' option.

$ vip @59 wp option get blogdescription
&quot;Hello world&quot;
```

joshbetz:
Closing this one for now

simongcc:
Hello!
I think there is still problem representing single quotes, for instance, there is any object with single quote inside

```
$test = array(
    'foo' => "I don't"
);
```

The following is working in php shell script / command line

1. wp option update some_option_name "{\"foo\":\"I don't\"}" --format=json #with ' in don't
2. wp option update some_option_name '{"foo":"I dont"}' --format=json # without ' in dont

The following does not work in both shell script and command line

- wp option update some_option_name '{"foo":"I don\'t"}' --format=json #with ' in don't

it will regard the command is not ended yet. The escaped slash "\'" did not being parsed correctly.

PR #458, opened 2019-08-22
Log some caught exceptions to rollbar

joshbetz:
We're already logging uncaught exceptions. Sometimes we also want to
log caught exceptions.

Fixes PLAT-288


elazzabi:
Yay, thank you. I think we can use `rollbar.error()` instead of `.log()`

I think with `.error()`  they got reported as Errors in rollbar which we can use to set alerts, etc.

elazzabi (approved):
Yay looks good 👍

PR #474, opened 2019-09-20
WP-CLI Add event handler for wp-cli cancel events

dchymko:
## Description
Adds an event handler to gracefully handle remote wp-cli command cancel events.

## Steps to Test

Example:

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-wp.js --yes  @appid -- shell`
4. Cancel the command remotely
5. Verify that the cancel message displays and the process exits.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -175,7 +175,7 @@ const launchCommandAndGetStreams = async ( { guid, inputToken, offset = 0 } ) =>
 	socket.on( 'cancel', message => {
 		console.log( `Cancel received from server: ${ message }` );
 		socket.close();
-		process.exit();
+		process.exit( 1 );
 	} );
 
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -172,6 +172,12 @@ const launchCommandAndGetStreams = async ( { guid, inputToken, offset = 0 } ) =>
 		console.log( 'There was an error with the authentication:', err.message );
 	} );
 
+	socket.on( 'cancel', message => {
+		console.log( `Cancel received from server: ${ message }` );
+		socket.close();
+		process.exit();
```
nickdaugherty: Should we use a non-zero exit code?

dchymko: Addressed in beb644ed7345a56831e1b7db3f4b6a238706e390

PR #478, opened 2019-10-12
Fix interactive commands not working correctly

### LINKED ISSUE #464: Interactive shell: second command failing
## Expected/Desired Behavior

Run more than one command via the interactive shell

## Actual Behavior

The second command run in an interactive shell hangs. 
There's zero output and the command needs to be killed `control + C`. This produces the output: `Command cancelled by user` but the shell remains in an unresponsive state. 
`Control + C` again and we exit the interactive shell.

- Begin an interactive shell session using `vip @XXX.develop wp`
- Run an initial command doesn't matter which
- Run a second command, doesn't matter which, and the interactive shell hangs

```
vip --version
1.7.0
```

dchymko:
## Description

This fixes problems running commands that require user input (such as `wp user create --prompt`). 

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `/dist/bin/vip-wp.js --yes  @site.env --  user create --prompt`
4. Confirm that entering data at a prompt results in the next prompt (or command completion)

Fixes #464


src/bin/vip-wp.js
```diff
@@ -388,6 +388,16 @@ commandWrapper( {
 			} );
 		} );
 
+		// Fix to re-add the \n character that readline strips when terminal == true
+		process.stdin.on( 'data', data => {
+			//only run this in interactive mode for prompts from WP commands
+			if ( isSubShell && commandRunning && 0 === Buffer.compare( data, Buffer.from( '\r' ) ) ) {
```
elazzabi:
```suggestion
			if ( commandRunning && 0 === Buffer.compare( data, Buffer.from( '\r' ) ) ) {
```

elazzabi (changes requested):
Yay tested locally and it works for the subshell 🕺 

I proposed a change where we can remove the subshell check so it can work outside the subshell too, what do you think?

src/bin/vip-wp.js
dchymko: Good catch. I thought this wasn't impacted outside of the psuedo-shell but forgot that we `subShellRl.write( )` the single commands as well.

dchymko:
@elazzabi Do you think we should merge this and then refactor readline to be `close()`-ed and recreated each time?

nickdaugherty:
My 2 cents - if this causes a known regression, it would be better to go straight to the readline fix.

dchymko:
> My 2 cents - if this causes a known regression, it would be better to go straight to the readline fix.

Whoops, this is no longer the case now that we've reverted the changes to the terminal setting and simply add the newline char instead.  Removed that from the description.

nickdaugherty (approved):
LGTM (not tested)

elazzabi (approved):
Looks good!

PR #516, opened 2020-01-15
WP-CLI: Exit on errors

joshbetz:
If the server returns an unauthorized or rate-limit error, do not
pass go, do not collect $200. (And most importantly, don't try to
reconnect to the socket.)


src/bin/vip-wp.js
```diff
@@ -192,6 +194,8 @@ const launchCommandAndGetStreams = async ( { guid, inputToken, offset = 0 } ) =>
 
 		rollbar.error( err );
 		console.log( err );
+
+		process.exit( 1 );
```
dchymko, line 200: Could we maybe add a threshold for the rate limit errors?  like after 3 attempts, the then `process.exit()`s

PR #547, opened 2020-03-24
Analytics: Prevent tracks calls when DO_NOT_TRACK is truthy

jblz:
## Description

Provide a mechanism to opt-out of telemetry / usage tracking. If the `DO_NOT_TRACK` environment variable is truthy, prevent the sending of data to the "tracks" system.

See: https://consoledonottrack.com/

PLAT-1054

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run the automated tests (`npm run test __tests__/lib/analytics/clients/tracks.js`)
    1. The added test logic should run & be logically-sound
    1. All tests should pass
1. Apply [this patch](https://gist.github.com/jblz/dd87fb4ca0f9f4cde43e0125bc998099) to manually test
1. Run `npm run build`
1. Run `DO_NOT_TRACK=1 ./dist/bin/vip-app-list.js`
    1. Verify your available apps are listed
    1. Verify all events are labeled `wouldNotTrack`
1. Run `./dist/bin/vip-app-list.js` (NOTE: this assumes you do not have the env var set system-wide.  If you do, override it like `DO_NOT_TRACK=0 ./dist/bin/vip-app-list.js` here.)
    1. Verify your available apps are listed
    1. Verify all events are labeled `wouldTrack`

### DIFF
--- a/src/lib/analytics/clients/tracks.js
+++ b/src/lib/analytics/clients/tracks.js
@@ -85,6 +85,10 @@ export default class Tracks implements AnalyticsClient {
 	}
 
 	send( extraParams: {} ): Promise<Response> {
+		if ( process.env.DO_NOT_TRACK ) {
+			return Promise.resolve( 'tracks disabled per DO_NOT_TRACK variable' );
+		}
+
 		const params = Object.assign( {}, this.baseParams, extraParams );
 
 		const method = 'POST';
--- a/src/lib/analytics/clients/tracks.js
+++ b/src/lib/analytics/clients/tracks.js
@@ -44,7 +44,7 @@ export default class Tracks implements AnalyticsClient {
 		};
 	}
 
-	trackEvent( name: string, eventProps = {} ): Promise<Response> {
+	trackEvent( name: string, eventProps = {} ): Promise<any> {
 		if ( ! name.startsWith( this.eventPrefix ) ) {
 			name = this.eventPrefix + name;
 		}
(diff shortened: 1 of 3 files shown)


src/lib/analytics/clients/tracks.js
```diff
@@ -85,6 +85,10 @@ export default class Tracks implements AnalyticsClient {
 	}
 
 	send( extraParams: {} ): Promise<Response> {
+		if ( process.env.DO_NOT_TRACK ) {
+			return Promise.resolve( 'tracks disabled per DO_NOT_TRACK variable' );
```
mjangda, line 91: We're no longer returning the correct type as indicated by the flow type for the function (`Promise<Response>`). Should we update that?

mjangda:
Alternate way to verify: `DO_NOT_TRACK=1 DEBUG=* ./dist/bin/vip-app-list.js`

`debug` will no longer output `@automattic/vip:analytics:clients:tracks send()` entries.

src/lib/analytics/clients/tracks.js
mjangda, line 91: Let's add a `debug` call for logging/debugging purposes.

```suggestion
			debug( 'send() => skipping per DO_NOT_TRACK variable' );

			return Promise.resolve( 'tracks disabled per DO_NOT_TRACK variable' );
```

mjangda (commented):
Spotted two minor things but otherwise this looks great!

mjangda (dismissed):
:ship:

elazzabi:
I think we should add this somewhere to the README. Any ideas on where we can add it? Maybe a new heading or something 🤔

jblz:
Good suggestion, @elazzabi.  Added a README section in 
233e4dc

PR #548, opened 2020-03-27
Update tool version to 1.9.0

jblz:
## Description

There have been a number of dependencies updated & some enhancements & bug fixes since the previous release, so this PR & subsequent release will get those out to users.


elazzabi (approved):
I merged some other package updates today. Should be good to go 👍

PR #566, opened 2020-06-08
Add a basic SQL file checker

dchymko:
## Description

An initial pass of a utility to run basic string-based checks on an SQL import file.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-validate.js <filename.sql>`

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "vip": "dist/bin/vip.js",
     "vip-app": "dist/bin/vip-app.js",
     "vip-app-list": "dist/bin/vip-app-list.js",
-    "vip-import-checker": "dist/bin/vip-app-check.js",
+    "vip-import-validate": "dist/bin/vip-import-validate.js",
     "vip-sync": "dist/bin/vip-sync.js",
     "vip-wp": "dist/bin/vip-wp.js"
   },
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "vip": "dist/bin/vip.js",
     "vip-app": "dist/bin/vip-app.js",
     "vip-app-list": "dist/bin/vip-app-list.js",
+    "vip-import-checker": "dist/bin/vip-app-check.js",
     "vip-sync": "dist/bin/vip-sync.js",
     "vip-wp": "dist/bin/vip-wp.js"
   },
--- a/src/bin/vip-import-check.js
+++ b/src/bin/vip-import-check.js
@@ -1,8 +1,6 @@
 #!/usr/bin/env node
 // @flow
 
-import { prompt } from 'enquirer';
-
 /**
  * External dependencies
  */
@@ -10,7 +8,6 @@ const readline = require( 'readline' );
 const fs = require( 'fs' );
 import chalk from 'chalk';
 const log = require( 'single-line-log' ).stdout;
-import path from 'path';
 
 /**
  * Internal dependencies
@@ -20,7 +17,7 @@ import command from 'lib/cli/command';
 let problemsFound = 0;
 let lineNum = 1;
 
-const errorHandler = ( check ) => {
+const errorCheckFormatter = ( check ) => {
 	if ( check.results.length > 0 ) {
 		problemsFound += 1;
 		console.error( chalk.red( 'Error:' ), `${ check.message } on line(s) ${ check.results.join( ',' ) }.` );
@@ -30,7 +27,7 @@ const errorHandler = ( check ) => {
 	}
 };
 
-const requiredHandler = ( check, type ) => {
+const requiredCheckFormatter = ( check, type ) => {
 	if ( check.results.length > 0 ) {
 		console.log( `✅ ${ check.message } was found ${ check.results.length } times.` );
 		if ( type === 'createTable' ) {
@@ -43,7 +40,7 @@ const requiredHandler = ( check, type ) => {
 	}
 };
 
-const infoHandler = ( check ) => {
+const infoCheckFormatter = ( check ) => {
 	check.results.forEach( item => {
 		console.log( item );
 	} );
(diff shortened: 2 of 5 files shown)


dchymko:
It felt like splitting the checks functionality into a modular block object would make it easier to add/change/test. Having the logic and formatting embedded together seemed suboptimal. But if this seems confusing, I'm open to other ideas.

package.json
```diff
@@ -7,6 +7,7 @@
     "vip": "dist/bin/vip.js",
     "vip-app": "dist/bin/vip-app.js",
     "vip-app-list": "dist/bin/vip-app-list.js",
+    "vip-import-checker": "dist/bin/vip-app-check.js",
```
pandah3: Oops, I think this is the wrong file path?

Also, minor- do we want the names to map to each other (i.e- either both `vip-import-check` or `vip-import-checker`)?
```suggestion
    "vip-import-checker": "dist/bin/vip-import-check.js",
```

src/bin/vip-import-check.js
```diff
@@ -0,0 +1,180 @@
...
+			wpTables.push( tableName );
+		} else if ( ! tableName.match( /^wp_/ ) ) {
+			notWPTables.push( tableName );
+		}
+	} );
+	if ( wpTables.length > 0 ) {
+		console.log( ` - wp_ prefix tables found: ${ wpTables.length } ` );
+	}
+	if ( notWPTables.length > 0 ) {
+		console.error( chalk.red( 'Error:' ), `tables without wp_ prefix found: ${ notWPTables.join( ',' ) } ` );
```
pandah3: Should we have the `problemsFound` counter count these errors as well? It's a bit confusing seeing errors logged but `Total of 0 errors found`:

```
Error: tables without wp_ prefix found: CdP_bro_images,CdP_commentmeta,CdP_comments,CdP_ctimelines,CdP_links,CdP_options,CdP_postmeta,CdP_posts,CdP_term_relationships,CdP_term_taxonomy,CdP_termmeta,CdP_terms,CdP_usermeta,CdP_users 

'siteurl','https://awxcode.com'
'home',''

Total of 0 errors found
```

pandah3 (commented):
Exciting, thank you for adding this 🥳 !

I tested with a couple of client SQL files from Zendesk, and everything is working as expected!
Should we also add a log for next steps the client should take (e.g.- `Your SQL file looks good and is now ready for import` or such)?

Also, do these need unit tests?

src/bin/vip-import-check.js
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env node
+// @flow
+
+/**
+ * External dependencies
+ */
+const readline = require( 'readline' );
+const fs = require( 'fs' );
+import chalk from 'chalk';
```
pandah3: We should also standardize these to `import..from` as well

package.json
markowsiak: Good catch.  I'd consider narrowing the name down to include `sql` in it as well (since there might be a media file import checker as well).

Perhaps something along the lines of `vip-sql-dump-checker`?

src/bin/vip-import-check.js
markowsiak: +1

src/bin/vip-import-check.js
markowsiak: +1

src/bin/vip-import-check.js
```diff
@@ -0,0 +1,180 @@
...
+			checkTablePrefixes( check.results );
+		}
+	} else {
+		problemsFound += 1;
+		console.error( chalk.red( 'Error:' ), `${ check.message } was not found.` );
+		console.error( chalk.yellow( 'Recommendation:' ), `${ check.recommendation }` );
+	}
+};
+
+const infoCheckFormatter = ( check ) => {
```
markowsiak: Interesting-ish find here (eslint/wp-prettier):
 
* on eslint --fix I see the parens go away and some spacings change
* on save (and auto wp-prettier) I see them come back, and spacings change back

markowsiak (commented):
Looking good!

src/bin/vip-import-check.js
mjangda: It's probably a conflict between eslint and prettier rules? I believe we have an eslint rule that doesn't require parens for single param arrow functions.

src/bin/vip-import-check.js
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env node
+// @flow
+
+/**
+ * External dependencies
+ */
+const readline = require( 'readline' );
+const fs = require( 'fs' );
+import chalk from 'chalk';
+const log = require( 'single-line-log' ).stdout;
```
mjangda:
```suggestion
import readline from 'readline';
import fs from 'fs';
import chalk from 'chalk';
import { stdout as log } from 'single-line-log';
```

package.json
mjangda: We should consider the name based out how users will be calling it. As it is now, they would call `vip import check dbdump.sql`. Something about `checker` and `check` just sounds off. 

A better option might be `vip import validate dbdump.sql` or with more specificity (to @markowsiak's point) `vip import validate-sql dbdump.sql`.

mjangda:
> It felt like splitting the checks functionality into a modular block object would make it easier to add/change/test. Having the logic and formatting embedded together seemed suboptimal

Definitely. ideally, all logical bits should live in the `lib` folder and all CLI-level bits, including output / formatting should be in `bin`. I don't think it's a blocker, but definitely something we should work towards.

src/bin/vip-import-check.js
dchymko: Added in https://github.com/Automattic/vip/pull/566/commits/8f50d540602a7c974cf73d5eaec6f20bdfc32b7f

package.json
dchymko: changed to `vip import validate sql` so that we can have other sub-commands in the future 
https://github.com/Automattic/vip/pull/566/commits/0eb1cdc66523abb6b4359ff9bc7204e539bdc043

src/bin/vip-import-check.js
dchymko: Thanks, addressed in https://github.com/Automattic/vip/pull/566/commits/881dbe95cef44e090ae8a98c5b87647e75aa48b8

PR #567, opened 2020-06-12
Add user agent for all API requests

mjangda:
Since sending the generic node-fetch header is not very helpful for debugging.

## To Test

- `npm run build`
- `node dist/bin/vip-app-list.js`
- Check logs to make sure request has a UA matching something like `vip-cli/1.10.0 (node/v12.18.0; darwin/19.5.0; +https://wpvip.com)`

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -15,6 +15,7 @@ import ProxyAgent from 'socks-proxy-agent';
  * Internal dependencies
  */
 import Token from './token';
+import env from './env';
 
 // Config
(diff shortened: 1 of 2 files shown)


vincenzo:
LGTM. Tested and got this in Kibana: 

```
http_user_agent | vip-cli/1.10.0 (node/v10.21.0; darwin/19.5.0; +https://wpvip.com)
```

And thanks for doing this. Having a more specific user agent would've indeed made my previous sifting through the logs much easier 🙂

vincenzo (approved):

PR #569, opened 2020-06-19
Use socks-proxy-agent for socket requests

mjangda:
Will allow us to secure the socket requests as well.

Move the initialization of the proxyagent to a shared lib to avoid duplicate code.

```
## With VIP_PROXY set and proxy active

# GraphQL request:
❯ node ./dist/bin/vip app 123
+ id: 123
+ name: example
+ repo: wpcomvip/example

node ./dist/bin/vip wp @123 option get home
https://example.com

# With proxy not active
node ./dist/bin/vip wp @123 option get home
App 123 does not exist
```

## Steps to Test

Use the following commands to verify things:

```
./dist/bin/vip @309 wp option get home
./dist/bin/vip app 309
```

- Start with `VIP_PROXY` not defined: commands should work (if you have superadmin)
- Set `VIP_PROXY` but do not start proxy: commands should error out
- Set `VIP_PROXY` and also start proxy: commands should work.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -22,6 +22,7 @@ import { confirm } from 'lib/cli/prompt';
 import { trackEvent } from 'lib/tracker';
 import Token from '../lib/token';
 import { rollbar } from 'lib/rollbar';
+import createSocksProxyAgent from 'lib/http/socks-proxy-agent';
 
 const appQuery = `id, name,
 	organization {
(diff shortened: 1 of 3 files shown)


vincenzo:
@mjangda  I can successfully execute the commands with or without VIP_PROXY defined. 🤔

vincenzo (approved):
I re-tested this and now works as expected. Not sure why, but hey :)

mjangda:
> I can successfully execute the commands with or without VIP_PROXY defined.

I think that was expected. However, now that we have the proxy enforcement in place, it should fail when not defined.

PR #575, opened 2020-07-02
Add import files validation

pandah3:
##  Description

Basic validator to validate media files for imports.

Checking for:

1. Folder structure/path validation
    - Single sites:
      - There’s an `uploads` directory
      - Recommended: There’s a year directory within the `uploads` directory
      - Recommended: There’s a month directory within the year directory
      - Recommended path: `uploads/2020/6/picname.jpg`
    - Not checking for multisites at the moment

2. File media types (e.g. – png, jpg, etc.)
    - Check for missing or invalid file types

3. Filename character validation
    - What's allowed:
      - Non-English characters
      - Special characters: `spaces ( ) [ ] ~`
    - What's not allowed:
      - Special characters: `+ & # % = \' \" \ < > : ; , / ? $ * | ` ! { }`

4. Intermediate images
Detect intermediate images, such as:

`panda4000x6000.jpg` (sizing w/ no space/dash/underscore)
`panda-4000x6000.jpg` (dash)
`panda_4000x6000.jpg` (underscore)
`panda test 4000x6000.jpg` (spaces)
`panda_test-4000x6000@2x.jpg` (retina display)

## Steps to Test

1. Check out PR.
1. Run `npm run jest lib/vip-import-validate-files.js`
1. All tests should pass
1. Download this sample `uploads` directory and unzip: 
[uploads.zip](https://github.com/Automattic/vip/files/5014552/uploads.zip)

1. Run `npm install`
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-validate-files.js uploads`
1. Verify messages:

validation

Other test folders:
[folder1.zip](https://github.com/Automattic/vip/files/5004948/folder1.zip)

validation2

PR #582, opened 2020-08-05
Import files: Add multiple nested folders validation

pandah3:
## Description

Building on top of the validation from #575.

This adds validation for multiple subdirectories in the same level in a nested directory tree structure such as:

- Given directory: `uploads`
 - Nested directories: `2020`, `2019`, `2018`, `2017`
  - Nested directories: `01`, `02`, `03`, `04`, `05`, `06`
   - Individual files: `image.jpg`, `image2.jpg, etc.`

## Steps to Test

1. Check out PR.
1. Download this sample `uploads` directory and unzip:
[uploads.zip](https://github.com/Automattic/vip/files/5030577/uploads.zip)

1. Run `npm install`
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-validate-files.js uploads`
1. Verify messages:

- Folder structure validation
- File extension validation
- Filename character validation
- Intermediate images validation

----------------------------------------

log1
log 2
log 3
log 4

### DIFF
--- a/src/lib/vip-import-validate-files.js
+++ b/src/lib/vip-import-validate-files.js
@@ -153,6 +153,7 @@ const recommendAcceptableFileNames = () => {
  * @param {string} directory Root directory, or the given (current) directory
  */
 let files = [];
+let folderStructureObj = {};
 
 export const findNestedDirectories = directory => {
 	let nestedDirectories;
@@ -173,7 +174,10 @@ export const findNestedDirectories = directory => {
 			if ( statSync.isDirectory() ) {
 				return findNestedDirectories( filePath );
 			} else {
-				// Once we hit media files, push them to an array to do individual file validations later on
+				// Once we hit media files, add the path of all existing folders
+				// as object keys to validate folder structure later on
+				folderStructureObj[ directory ] = true;
+				// Also, push individual files to an array to do individual file validations later on
 				return files.push( filePath ); 
 			}
 		} );
--- a/src/lib/vip-import-validate-files.js
+++ b/src/lib/vip-import-validate-files.js
@@ -144,38 +144,38 @@ const recommendAcceptableFileNames = () => {
  *
  * Use recursion to identify the nested tree structure of the folders
  *
- * @param {string} directory Root directory, or the current directory
+ * @param {string} directory Root directory, or the given (current) directory
  */
-export const findNestedDirectories = async directory => {
-	let dir, nestedDir;
+let individualFiles = [];
+
+export const findNestedDirectories = directory => {
+	let nestedDirectories;
 
 	try {
-		// Read what's inside the current directory
-		dir = await readDir( directory );
+		// Read nested directories within the given directory
+		nestedDirectories = fs.readdir( directory );
 
 		// Filter out hidden files such as .DS_Store
-		dir = dir.filter( file => ! ( /(^|\/)\.[^\/\.]/g ).test( file ) );
+		nestedDirectories = nestedDirectories.filter( file => ! ( /(^|\/)\.[^\/\.]/g ).test( file ) );
 
-		const firstNestedFolder = dir[ 0 ]; // The first nested folder of the given directory
-		nestedDir = firstNestedFolder;
+		// For each directory
+		nestedDirectories.forEach( dir => {
+			const filePath = path.join( directory, dir );
+			const statSync = fs.stat( filePath ); // Get stats on the file/folder
 
-		// Once we hit individual media files, stop
-		const regexExtension = /\.\w{3,4}$/;
-		const mediaFiles = regexExtension.test( nestedDir );
-
-		if ( dir !== undefined && mediaFiles ) {
-			return directory;
-		}
+			// Keep looking for nested directories until we hit individual files
+			if ( statSync.isDirectory() ) {
+				return findNestedDirectories( filePath );
+			} else {
+				// Once we hit media files, push them to an array to do invidual file validations later on
+				return individualFiles.push( filePath ); 
+			}
+		} );
 	} catch ( error ) {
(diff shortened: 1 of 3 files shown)


pandah3:
I'd love some input on the output/logs that get displayed @SaxonF 🙇‍♀️ . Namely:

- The larger the client logs get, the messier it gets (the example output images in the description box are outputs from _partial_ import files)
- We need to coordinate the output from the media files validation (this PR & #575) and SQL file validation logs

Media file validation output:
validation simple

SQL file validation output:
Screen Shot 2020-08-05 at 3 03 20 PM

(Sorry for the blurry screenshot ^. Maybe @dchymko has a clearer one.)

SaxonF:
Looks pretty awesome to me @pandah3! Nice one!

I like the grouping by folder as its similar to running a test suite. If we wanted to compress it a bit I guess we could group the "directory exists" success messages unless one of them fails? I also wonder if it would be helpful to include a summary at the end letting people know how close they are to passing?

image

These are just small things that could be added later if they make sense and shouldn't block a release.

src/lib/vip-import-validate-files.js
```diff
@@ -187,80 +199,84 @@ export const findNestedDirectories = async directory => {
...
-	} else {
-		console.log( chalk.yellow( '✕' ), 'Recommended: Structure your WordPress media files into', chalk.magenta( '`uploads/YYYY`' ), 'directories' );
-		errors++;
-	}
+		/**
+			* Logging
+			*/
+
+		// Uploads folder
+		if ( uploadsIndex === 0 ) {
```
markowsiak, line 256: I think the index parts are what didn't pass for me if I supply a path argument that isn't relative to the test execution (ie. `~/Downloads/upload`):

image

We might have to re-evaluate how we reference this.  Maybe we can skip validating what the index is of the uploads directory and just ensure that the year and month are at the upload directory index +1 and +2 respectively?  That's one way, but perhaps there's another if we have to validate the upload directory position?

src/lib/vip-import-validate-files.js
```diff
@@ -142,40 +142,52 @@ const recommendAcceptableFileNames = () => {
...
+		nestedDirectories = nestedDirectories.filter( file => ! ( /(^|\/)\.[^\/\.]/g ).test( file ) );
+
+		nestedDirectories.forEach( dir => {
+			// Concatenate the file path of the parent directory with the nested directory
+			const filePath = path.join( directory, dir );
+			const statSync = fs.statSync( filePath ); // Get stats on the file/folder
+
+			// Keep looking for nested directories until we hit individual files
+			if ( statSync.isDirectory() ) {
+				findNestedDirectories( filePath );
```
markowsiak, line 175: Should this have a maximum call depth to pass?
Or can customers have an unlimited number of nested folders inside of the `yyyy/mm/` directory?

markowsiak (commented):
Looks good @pandah3 ⭐ 
I ran into an issue during test execution, but I think I found out why in the comments below.  Let me know if my suggestion is reasonable, or if the use-case of the path I supplied isn't required 👍

src/lib/vip-import-validate-files.js
dchymko: Somewhat of an edge case but not a bad idea. Maybe we should limit it to "X" and have a warning show if it's reached (and possibly a flag to override that number)

dchymko (dismissed):
LGTM. That recursion thing is a minor edge case IMO from both the likelihood and also because this will run on a user's machine and not on affect sites, etc

src/lib/vip-import-validate-files.js
```diff
@@ -411,3 +418,31 @@ export const logErrorsForIntermediateImages = obj => {
 	}
 	console.log( '------------------------------------------------------------' );
 };
+
+export const summaryLogs = ( folderErrorsLength, intImagesErrorsLength, fileTypeErrorsLength, filenameErrorsLength, totalFiles, totalFolders ) => {
```
elazzabi: Minor feedback

In general, for functions with more than 2 arguments, it's hard to remember the order of arguments. Another way of doing this is by accepting an object as an argument with named attributes:

```js
export const summaryLogs = ( { folderErrorsLength, intImagesErrorsLength, fileTypeErrorsLength, filenameErrorsLength, totalFiles, totalFolders } )
```

When calling the function, it's easier than remembering the order:
```js
summaryLogs( {
  folderErrorsLength: folderValidation.length,
  intImagesErrorsLength: intermediateImages.tally,
  fileTypeErrorsLength: errorFileTypes.length,
  filenameErrorsLength: errorFileNames.length,
  totalFiles: files.length,
  totalFolders: nestedDirectories.length
} )
```

elazzabi (commented):
Looks good. Left a minor feedback

src/lib/vip-import-validate-files.js
pandah3: Thank you for the feedback 🙏 ! I think in this instance, it's working as expected- if the supplied media folder is `Downloads/uploads`, the recommended structure is simply an `uploads` folder (without the `Downloads` folder). Maybe the wording needs to be clearer?

src/lib/vip-import-validate-files.js
pandah3: I'll do some more research into what the limits are and see if we should add one in a following PR!

src/lib/vip-import-validate-files.js
pandah3: Gah I meant to do this but forgot! Thank you! Fixed in 1abd712!

pandah3:
Thank you everyone for the feedback! I added a summary per Saxon's suggestion! Will add any follow ups in a new PR.

summary

pandah3:
Merging!

PR #585, opened 2020-08-17
SQL Import: Get presigned S3 URLs from the API, Upload SQL files, & kick off an import

jblz:
## Description

Includes:
* Validate the SQL file before attempting to upload it.
* Compress uncompressed files greater than a threshold
* Get a pre-signed upload URLs from the public API (which gets it from the private API)
* If the file is larger than an arbitrary size threshold, use the S3 Multipart uploading scheme:
  * Get pre-signed upload URL / request data to upload each piece of the file
  * Use the URLs to `PUT` each piece
  * Get pre-signed upload URL / request data to "Complete" the transfer
  * Use the URL to complete the transfer
* ...or if the file is smaller, just use the `PutObject` API
* Calculate the md5 checksum
* Include required info in a public API call to `StartImport`
* tests!

### Manually Test

1. Run the public API with the branch & instructions in Automattic/vip-go-api-public#965 (including the GOOP piece)
1. Check out this PR.
1. Run `npm run build`
1. Run `API_HOST=http://localhost:4000 ./dist/bin/vip-import-sql.js @1000.staging ./__fixtures__/client-file-uploader/db-dump-ipsum-67mb.sql`
    1. You should see no errors and the file should be uploaded in a single part
1. Run `API_HOST=http://localhost:4000 ./dist/bin/vip-import-sql.js @1000.staging $PATH_TO_SOME_LARGEISH_SQL_FILE` (where `$PATH_TO_SOME_LARGEISH_SQL_FILE` is any file that's larger than about 100 MB -- so it won't be smaller than 32MB when gzipped)
    1. You should see no errors and the file should be uploaded in multiple parts
1. After each SQL file is uploaded, a `StartImport` mutation should be properly initiated (Follow the test plan in each server for the expected behavior)

### Automated Tests

`npm run jest __tests__/lib/client-file-uploader.js`


src/lib/site-import/db-file-import.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+/**
+ * External dependencies
+ */
+
+export async function importFile() {}
+
+export function currentUserCanImportForApp( app: any ): boolean {
+	// TODO: implement
```
jblz: Here is a spot I could use a hand implementing

src/lib/site-import/db-file-import.js
```diff
@@ -0,0 +1,39 @@
...
+	return !! app;
+}
+
+export function isSupportedApp( app: any ): boolean {
+	// TODO: implement
```
jblz, line 28: Here is a spot I could use a hand implementing

src/bin/vip-import-sql.js
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	console.log( 'You are about to import a SQL file to site:' );
+	console.log( `ID: ${ app.id }` );
+	console.log( `Name: ${ app.name }` );
+	console.log( `Primary Domain Name: ${ primaryDomainName }` );
+
+	try {
+		const uploadURL = await getUploadURL( organization.id, app.id, 'omgwtfbbq.sql' );
+		console.log( await uploadURL.text() );
```
jblz: Next up is using the URL to actually upload a file.

jblz (commented):
@Automattic/vip-platform-patisserie I'm going to move on to actually uploading the file.  I notated a couple of places I could use a hand with implementation.

src/lib/api.js
```diff
@@ -53,12 +53,44 @@ export default async function API(): Promise<ApolloClient> {
...
 		cache: new InMemoryCache(),
 	} );
+
+	/**
+	 * Call the Public API with an arbitrary path (e.g. to connect to REST endpoints)
+	 * @param {*} path API path to pass to `fetch` -- will be prefixed by the API_HOST
+	 * @param {*} options options to pass to `fetch`
+	 * @returns {Promise} Return value of the `fetch` call
+	 */
+	apiClient.apiFetch = ( path: string, options = {} ) =>
```
jblz: How does everyone feel about this helper to make an authenticated call to the public API outside the GraphQL endpoint? I didn't see a handy way to do it if it exists.

src/lib/site-import/db-file-import.js
dchymko: We are just checking here that this is a WordPress site, or are we checking for a feature flag as well?

src/lib/api.js
dchymko: This seems like a fine solution to me!

src/lib/site-import/db-file-import.js
jblz: @dchymko This function is just for a WordPress-type site (so WP preprod & prod app types).

There are probably other constraints imposed by the back-end that we'll want to catch in this block as well (some subsite-specific things, presence of lock meta, etc.), but those can be added separately.

RE: feature flags, I'm not yet familiar with how we do those in this project.  We'll want to add that in too if it's not a huge lift.

Thanks for taking a look!

src/lib/client-file-uploader/index.js
```diff
@@ -0,0 +1,19 @@
...
+ * @flow
+ * @format
+ */
+
+/**
+ * Internal dependencies
+ */
+import API from 'lib/api';
+
+export async function getUploadURL( organizationId: number, appId: number, objectBaseName: string ): Promise<any> {
```
jblz: This file / function should be pulled out...  It snuck through a rebase.

src/bin/vip-import-sql.js
```diff
@@ -0,0 +1,78 @@
...
+	console.log( '** Welcome to the WPVIP Site SQL Importer! **\n' );
+
+	if ( ! currentUserCanImportForApp( app ) ) {
+		err( 'The currently authenticated account does not have permission to perform a SQL import.' );
+	}
+
+	if ( ! isSupportedApp( app ) ) {
+		err( 'The type of application you specified does not currently support SQL imports.' );
+	}
+
```
jblz, line 62: TODO: Integrate file validation here

__tests__/lib/site-import.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Internal dependencies
+ */
+import siteImport from 'lib/site-import';
+import dbFileImport from 'lib/site-import/db-file-import';
+
+describe( 'site import tests', () => {
+	describe( 'siteImport', () => {
+		it( 'should should have an object called dbSiteImport', () => {
```
elazzabi: Duplicate should here
```suggestion
		it( 'should have an object called dbSiteImport', () => {
```

src/bin/vip-import-sql.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Internal dependencies
+ */
+import command from 'lib/cli/command';
+import { currentUserCanImportForApp, isSupportedApp } from 'lib/site-import/db-file-import';
+import { getFileMeta, uploadFile } from 'lib/client-file-uploader';
+
+/**
+ * - Include `import_in_progress` state & error out if appropriate (this likely needs to be exposed in the data graph)
```
elazzabi, line 25: We'll definitely need a query for this if we want to show the progress in CLI/Dashboard

src/bin/vip-import-sql.js
```diff
@@ -0,0 +1,78 @@
...
+
+const err = message => {
+	console.log( chalk.red( message.toString().replace( /^(Error: )*/, 'Error: ' ) ) );
+	process.exit( 1 );
+};
+
+command( {
+	appContext: true,
+	appQuery,
+	requiredArgs: 1, // TODO print proper usage example
```
elazzabi, line 44: Do you think it makes sense to add a confirmation step given this action will alter data?

src/bin/vip-import-sql.js
```diff
@@ -0,0 +1,78 @@
...
+	if ( ! currentUserCanImportForApp( app ) ) {
+		err( 'The currently authenticated account does not have permission to perform a SQL import.' );
+	}
+
+	if ( ! isSupportedApp( app ) ) {
+		err( 'The type of application you specified does not currently support SQL imports.' );
+	}
+
+	console.log( 'You are about to import a SQL file to site:' );
+	console.log( `ID: ${ app.id }` );
```
elazzabi: We've been using this function so far to console data: https://github.com/Automattic/vip/blob/master/src/lib/cli/format.js#L11

Do you think it makes sense to use it here too to have the same logging across the CLI?

src/lib/client-file-uploader.js
```diff
@@ -0,0 +1,447 @@
...
+import API from 'lib/api';
+
+const KILOBYTE_IN_BYTES = 1024;
+const MEGABYTE_IN_BYTES = KILOBYTE_IN_BYTES * 1024;
+
+// Files smaller than MULTIPART_THRESHOLD will use `PutObject` vs Multipart Uploads
+export const MULTIPART_THRESHOLD = 32 * MEGABYTE_IN_BYTES;
+const UPLOAD_PART_SIZE = 5 * MEGABYTE_IN_BYTES;
+
+export interface GetSignedUploadRequestDataArgs {
```
elazzabi, line 46: These are captured by `flow` right? We should use it more often 🤔

src/lib/client-file-uploader.js
```diff
@@ -0,0 +1,447 @@
...
+export type FileMeta = {
+	basename: string,
+	md5: string,
+	fileName: string,
+	size: number,
+};
+
+export async function checkFileAccess( fileName: string ): Promise<void> {
+	// Node 8 doesn't have fs.promises, so fall back to this
+	return new Promise( ( resolve, reject ) => {
```
elazzabi, line 331: If this provides the same objective as `fs.promises` then all good. Otherwise, I think we can use `fs.promises` given code is transpiled by babel and Node 8 reached the end of life already.

elazzabi (dismissed):
Looks good. Left some small notes/questions

src/lib/client-file-uploader.js
jblz: We're still running tests in node 8:

https://github.com/Automattic/vip/blob/bbfbb8f9693aa2f6ba5411bf9695478c9e56cb80/.circleci/config.yml#L9

It was failing on `fs.promises` & this works around it.  If we're no longer supporting node 8, we should take it out & bump our minimum node version (e.g. here https://github.com/Automattic/vip/blob/bbfbb8f9693aa2f6ba5411bf9695478c9e56cb80/package.json#L46)

markowsiak: +1 for bumping to at least node 10 (probably 12).
But.... we should probably do that in a separate PR to ensure we don't introduce another variable of change.

markowsiak (commented):
Looking awesome 👍

src/bin/vip-import-sql.js
markowsiak: https://github.com/Automattic/vip/pull/593 👍

src/lib/client-file-uploader.js
```diff
@@ -0,0 +1,600 @@
...
+				isCompressed: false,
+			},
+		} );
+	} catch ( e ) {
+		throw `Unable to upload file checksum: ${ e }`;
+	}
+
+	console.log( 'Checksum uploaded.' );
+
+	// TODO -- send md5 hash along with the host action payload instead of uploading separately
```
jblz: Jeff: DO THIS!

src/bin/vip-import-sql.js
jblz: It definitely makes sense to add a confirm step.

I tried to specify `requireConfirm` in the `command` invocation, but it threw an error.  It looks pretty easy to fix, but will probably require a change to the lib.

src/lib/client-file-uploader.js
jblz: done!

markowsiak (commented):
No blockers found here.
🚢   when the end-to-end test follows through as we expect.

- pending merge of Public API's PR #994
- pending addition of db containers for the test site in question

PR #588, opened 2020-08-19
Drop Node 8 Support, Update Babel, & Support modern syntax across node versions

jblz:
## Description

The main goal of this PR is to add language support for [optional chaining](https://github.com/tc39/proposal-optional-chaining/tree/8461024a040d9c41ec039e2d4291fe2dfda36574) & [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing/tree/5b2a58ee3a56c2b5a5eb30a30959e32ae61d4c49). Both are [Stage 4 proposals](https://tc39.es/process-document/) which means they "will be included in the soonest practical standard revision."

* Install `@babel/preset-env` dev dependency
* Use `@babel/preset-env` targeting node v8.9.0 (current >= target in package.json)
* Allow optional chaining & nullish coalescing in the flowconfig. 
* Target transpilation to node 10

Additionally, this drops support for node 8 as it's EOL'd.

## Steps to Test

1. Check out PR.
1. rm your `node_modules` folder
1. `npm i`
1. Run `npm run build`
    * There should be no build errors
1. Run some commands
    * They should behave identically as the upstream branch
1. Automated tests should pass
1. Built bundle size should not be drastically increased


package.json
```diff
@@ -51,8 +51,9 @@
   "homepage": "https://github.com/Automattic/vip#readme",
   "devDependencies": {
     "@babel/cli": "7.8.4",
-    "@babel/core": "7.9.0",
+    "@babel/core": "^7.11.1",
```
jblz: This is not essential for this change and I'm happy to pull it out if someone finds a good reason.

babel.config.js
```diff
@@ -1,5 +1,13 @@
 module.exports = {
 	presets: [
+		[
+			'@babel/preset-env',
+			{
+				targets: {
+					node: '8.9.0', // Keep this in sync with package.json engines.node
```
markowsiak: Should we bump this to 10.x (in both places)?
8.x has been EOL I think since December?

markowsiak (dismissed):
Just a non-blocking question. LGTM 🚢

babel.config.js
markowsiak: That might mean updating the public docs as well: https://wpvip.com/documentation/vip-go/vip-cli/#installing-vip-cli-locally

jblz:
The node 8 test has been removed and won't fulfill the expectation.  Once this is merged, we should update those test expect settings.

uxcitizen:
When this gets merged, can whoever merges it please update the docs here about the required version: https://docs.wpvip.com/technical-references/vip-cli/installing-vip-cli/?

Thank you!

mjangda:
We can just update the docs now :) No need to wait for this to be merged and released.

elazzabi:
Updated minimum versions in the documentation.

In order to comply with the semver, the next version should be a major giving we are dropping support for Node 8 cc @markowsiak

elazzabi (dismissed):
Tested locally! Works fine ❤️

PR #589, opened 2020-08-20
Add Tracks events for validation commands

dchymko:
## Description
This sends Tracks event for the media and SQL import validation commands. This will give us better insight into the issues that clients encounter with importing data and media.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-validate-sql.js import.sql `
1. Verify the event as sent to Tracks
1. Alternatively, run the command with `DEBUG=*` in front of it to confirm the track request was sent

### DIFF
--- a/src/bin/vip-import-validate-files.js
+++ b/src/bin/vip-import-validate-files.js
@@ -182,3 +182,4 @@ command( { requiredArgs: 1, format: true } )
 
 		await trackEvent( 'import_validate_files_command_success', allErrors );
 	} );
+	
--- a/src/bin/vip-import-validate-files.js
+++ b/src/bin/vip-import-validate-files.js
@@ -25,6 +25,7 @@ import {
 	logErrorsForInvalidFilenames,
 	summaryLogs,
 } from '../lib/vip-import-validate-files';
+import { trackEvent } from 'lib/tracker';
 
 // Promisify to use async/await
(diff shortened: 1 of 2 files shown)


pandah3:
We use the error object in the media files validation file multiple times so I assigned it to a variable in b07fb5c.

I'm seeing the Tracks events being sent when running `DEBUG=* node ./dist/bin/vip-import-validate-files.js uploads`, but I don't see them on the Tracks page itself. Tried it several times with both the SQL and media files events, but no luck.

Going to do some more digging to see what's going on.

pandah3 (commented):
We use the error object in the media files validation file multiple times so I assigned it to a variable in b07fb5c.

I'm seeing the Tracks events being sent when running `DEBUG=* node ./dist/bin/vip-import-validate-files.js uploads`, but I don't see them on the Tracks page itself. Tried it several times with both the SQL and media files events, but no luck.

Going to do some more digging to see what's going on.

pandah3:
Tracks events were sending, but they were getting rejected because the props needed to be in `snake_case` rather than `camelCase`. Fixed in 49d1443 + b0e67af.

pandah3 (approved):
Whew, glad we got that fixed! LGTM 🚀 🎉 !

PR #601, opened 2020-11-25
[SQL Validation] Add validation for trigger statements

pandah3:
##  Description

There was a failed import recently due to a `CREATE TRIGGER` statement being added to a SQL file.
`TRIGGER` statements require `SUPER` permissions, which is only reserved for SysEng.

This adds a static check for any `TRIGGER` statements (case sensitive).
Initially, this was checking for a ``CREATE (DEFINER=`root`@`host`) TRIGGER`` statement, but because the `DEFINER` is optional, it created an unnecessarily complicated regex check.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Prepare a SQL file with some `TRIGGER` statements
1. Run `./dist/bin/vip-import-validate-sql.js <file.sql>`
(e.g- `./dist/bin/vip-import-validate-sql.js ./test.sql`)
1. Should output something like:

```bash
Finished processing 4006 lines.

✅ USE statement was found 0 times.

✅ CREATE DATABASE statement was found 0 times.

Error: TRIGGER statement on line(s) 825, 941, 1005.
Recommendation: Remove these lines

✅ DROP DATABASE statement was found 0 times.

✅ ALTER USER statement was found 0 times.

✅ DROP TABLE was found 122 times.

✅ CREATE TABLE was found 122 times.
 - wp_ prefix tables found: 122 

'siteurl','https://site-develop2.go-vip.net'
'home',''
'home',''
'home',''
'home',''
'home',''
'home',''
'home','home'
'home','publish'
'home','inherit'

Total of 1 errors found
```


elazzabi (approved):
Looks good 👍

PR #604, opened 2020-11-26
Ignore an empty VIP_PROXY arg

mjangda:
Because it's awkward to have to unset the env var to not use it.

`VIP_PROXY="" vip` should ignore proxy but currently doesn't.


markowsiak (approved):
🚢

PR #605, opened 2020-11-30
PLAT-1575: Adding @automattic/vip-search-replace into the public cli

markowsiak:
## Description

This PR aims to add search and replace functionality to two places in the public CLI.

1. In a stand-alone command that can be used to run s&r without performing a DB import
1. As part of the SQL import command, so that s&r can occur before the import is executed

Note: you can pass an `--in-place` option to either command to have the search and replace utility overwrite the original input file.  When not provided and input of `filename.ext` creates an output file of `filename.out.ext`.

## Steps to Test

**Example to test s&r alone:**

1. Check out PR.
1. `npm i`
1. Get a SQL file to run s&r on (eg. `scp vipdev@go:/home/vipdev/Git/vip-go-internal-cli/test-dumpl.sql ./`)
1. Run `npm run build && VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-search-replace test-dumpl.sql --search-replace=countrycode.com,c0untryc0de.com --search-replace=thing1,thing2`
1. Verify that the original file created is present and contains the original SQL details with the s&r completed

**Example to test s&r with an import:**

1. Check out PR
1. `npm i`
1. Get a SQL file that you want to upload
1. *** Setup your local to use the Import flow (need to add details here)
1. Run `npm run build && VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-sql test-dumpl.sql --search-replace=countrycode.com,c0untryc0de.com`
1. See that the import succeeded and that the s&r operated correctly before the import

---
_This PR was started by: [git wf pr](https://github.com/groupon/git-workflow/releases/tag/v1.1.0)_


src/lib/search-and-replace.js
```diff
@@ -0,0 +1,111 @@
...
+				resolve( outputFile );
+			} )
+			.on( 'error', () => {
+				console.log( chalk.red( 'Oh no! We could not write to the output file.' ) );
+				reject();
+			} );
+	} );
+
+	const endTime = process.hrtime( startTime );
+	const end = endTime[ 1 ] / 1000000; // time in ms
```
markowsiak, line 184: Do we have a normal unit of time that we send data to tracks in?

src/bin/vip-import-sql.js
```diff
@@ -38,79 +41,93 @@ const err = message => {
...
+		const [ fileName ] = arg;
 
-	await trackEventWithEnv( 'import_sql_command_execute' );
+		const trackEventWithEnv = async ( eventName, eventProps = {} ) =>
+			trackEvent( eventName, { ...eventProps, appId: env.appId, envId: env.id } );
 
-	console.log( '** Welcome to the WPVIP Site SQL Importer! **\n' );
+		await trackEventWithEnv( 'import_sql_command_execute' );
 
-	if ( ! currentUserCanImportForApp( app ) ) {
```
jblz, line 59: I've been meaning to pull out the `currentUserCanImportForApp` function entirely since it's pretty useless as is & should be handled by the `appQuery`, I think:

https://github.com/Automattic/vip/blob/74a834868dcf9e1dbde2d356077d6c6150ba6814/src/lib/site-import/db-file-import.js#L27-L30

...but it seems unrelated to the rest of this PR.  Can we leave this be for the purposes of this PR & address it separately, please?

src/bin/vip-import-sql.js
```diff
@@ -38,79 +41,93 @@ const err = message => {
...
-	await validate( fileName, true );
+		if ( ! isSupportedApp( app ) ) {
+			await trackEventWithEnv( 'import_sql_command_error', { errorType: 'unsupported-app' } );
+			err( 'The type of application you specified does not currently support SQL imports.' );
+		}
 
-	/**
-	 * TODO: We should check for various site locks (including importing) prior to the upload.
-	 */
+		const inputFile = await searchAndReplace( fileName, searchReplace, { isImport: true, inPlace: opts.inPlace } );
```
jblz: The option input is described as `in-place` above...does it translate to camelCase automatically?

Additionally, I just want to make sure it defaults to `false` and the original input file isn't clobbered inadvertently.

src/bin/vip-import-sql.js
```diff
@@ -38,79 +41,93 @@ const err = message => {
...
+		await validate( inputFile, true );
 
-	const api = await API();
+		if ( ! isSupportedApp( app ) ) {
+			err( 'The type of application you specified does not currently support SQL imports.' );
+		}
 
-	try {
-		const { fileMeta: { basename, md5 }, result } = await uploadImportSqlFileToS3( { app, env, fileName } );
+		console.log( formatData( [
```
jblz: This looks to have come through in the merge inadvertently.  This output was moved to the `command` `requireConfirm` code in #610.

jblz (changes requested):
Still reviewing, but left a merge-related issue & a couple of other small-ish things.

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,111 @@
...
+	const replacedStream = await replace( readStream, replacements );
+	const result = await new Promise( ( resolve, reject ) => {
+		replacedStream
+			.pipe( writeStream )
+			.on( 'finish', () => {
+				console.log( chalk.green( 'Search and Replace Complete!' ) );
+				resolve( outputFile );
+			} )
+			.on( 'error', () => {
+				console.log( chalk.red( 'Oh no! We could not write to the output file.' ) );
```
jblz: This error will be user-facing, correct?  Are there any actionable suggestions we can make here? Perhaps to check for available disk space or file system permissions...that sort of thing.

src/lib/search-and-replace.js
jblz, line 184: > Do we have a normal unit of time that we send data to tracks in?

I think it depends on the latency range -- ms is probably a good precision-level for this, IMO.

src/bin/vip-import-sql.js
markowsiak: @jblz I don't think I did this intentionally.  I'm going to check and see if this resulted from that merge in from master.

src/bin/vip-import-sql.js
markowsiak: That's correct.  `--in-place` becomes `inPlace`, and similarly `--search-replace` becomes `searchReplace`.  I couldn't find this in the args README (docs) but I logged it out to check.  This occurs in commander as well.

I checked while testing and omitting `--in-place` results in line https://github.com/Automattic/vip/pull/605/files#diff-b92d26906cc64aca57a3c65329e117834c0c66339d30de1959ed88e6efadd2feR83 using the `outputFileReplacement` logic.  Including it results in the `inPlaceReplacement` logic being executed.

src/bin/vip-import-sql.js
markowsiak: Adding this back in 👍

src/bin/vip-import-sql.js
markowsiak: Cleaning this up to match 👍

src/lib/search-and-replace.js
markowsiak: Adding in your suggestions 👍

.flowconfig
```diff
@@ -1,5 +1,5 @@
 [ignore]
-node_modules
+.*/node_modules/.*
```
jblz: This breaks module resolution in my editor:

Screen Shot 2020-12-08 at 4 53 53 PM

What's the goal in this change?

markowsiak: Oh, interesting.  I added that change because flow was linting code inside node_modules and failing on my local.

markowsiak: I believe I added the same in `vip-ui`.
Notably, I also started to notice some module resolution issues in my editor as well, but it seemed unpredictable when.

jblz: I'll pull it out for now -- let's revisit in another PR if it's still failing for you.

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,124 @@
...
+
+const debug = debugLib( '@automattic/vip:lib:search-and-replace' );
+
+const flatten = arr => {
+	return arr.reduce( function( flat, toFlatten ) {
+		return flat.concat( Array.isArray( toFlatten ) ? flatten( toFlatten ) : toFlatten );
+	}, [] );
+};
+
+const inPlaceReplacement = async filename => {
```
jblz: The fn name `inPlaceReplacement` makes me think this would do the replacement... how about:

```suggestion
const inPlaceReplacementFilenames = async filename => {
```

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,124 @@
...
+	const copyFile = util.promisify( fs.copyFile );
+	await copyFile( filename, tmpFilePath );
+
+	return {
+		inputFile: tmpFilePath,
+		outputFile: filename,
+	};
+};
+
+const outputFileReplacement = filename => {
```
jblz: Similarly, the fn name `outputFileReplacement` doesn't make it clear to me that this is returning a mapping of file names.

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,124 @@
...
+	};
+};
+
+export type searchReplaceOptions = {
+	isImport: boolean,
+	inPlace: boolean,
+};
+
+export const searchAndReplace = async (
+	filename: string,
```
jblz: `fileName` should be camelCase throughout.

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,124 @@
...
+	}, [] );
+};
+
+const inPlaceReplacement = async filename => {
+	await confirm(
+		[],
+		'Are you sure you want to run search and replace on your input file? This operation is not reversible.'
+	);
+
+	const tmpFilePath = path.join( os.tmpdir(), ( +new Date() ).toString( 36 ) );
```
jblz: Instead of constructing temporary file names we think will be unique, let's use the built-in functionality to create a temporary directory and just put temp files there with basename intact.

https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback

src/lib/search-and-replace.js
```diff
@@ -0,0 +1,124 @@
...
+	return {
+		inputFile: tmpFilePath,
+		outputFile: filename,
+	};
+};
+
+const outputFileReplacement = filename => {
+	return {
+		inputFile: filename,
+		outputFile: suffix( filename, '.out' ),
```
jblz: This should be a file that gets created in the temp dir as well.

jblz (changes requested):
This is working really well & is really close!

The only real blocker I see is that we shouldn't clobber any output files (e.g. creating a file w/ a suffix without making sure a file isn't already there).

I'll try to commit a fix that uses a temp dir this eve.

jblz (dismissed):
🚀 🎆

PR #606, opened 2020-12-01
Fix all the lint warnings

markowsiak:
## Description

This PR cleans up all the lint warnings in the repo.  Fixes were provided by:

* updating the version of our eslint rules to a more current one
* skipping eslint where we need to use camelcase
* skipping eslint where we need to import flow types
* skipping eslint where lines are too long and it doesn't seem reasonable to break it up
* removing unused variables and functions
* fixing how we spy on console.log

## Steps to Test

1. Checkout the master branch and run `npm run lint`.  Note all the warnings.
1. Checkout this branch and run `npm t` and note there are no warnings and tests pass

🎅 

---
_This PR was started by: [git wf pr](https://github.com/groupon/git-workflow/releases/tag/v1.1.0)_

### DIFF
--- a/src/bin/vip-import-validate-files.js
+++ b/src/bin/vip-import-validate-files.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 // @flow
 
+/* eslint-disable camelcase */
+
 /**
  * External dependencies
  */
@@ -28,13 +30,11 @@ import {
 import { trackEvent } from 'lib/tracker';
 
 // Promisify to use async/await
-const syncStat = promisify( fs.statSync );
 const stat = promisify( fs.stat );
-const readDir = promisify( fs.readdir );
 
 command( { requiredArgs: 1, format: true } )
 	.example( 'vip import validate files <file>', 'Run the import validation against the file' )
-	.argv( process.argv, async ( arg, options ) => {
+	.argv( process.argv, async arg => {
 		await trackEvent( 'import_validate_files_command_execute' );
 		/**
 		 * File manipulation
--- a/src/bin/vip-import-validate-files.js
+++ b/src/bin/vip-import-validate-files.js
@@ -1,8 +1,6 @@
 #!/usr/bin/env node
 // @flow
 
-/* eslint-disable camelcase */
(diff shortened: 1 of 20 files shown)


src/bin/vip-import-validate-files.js
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 // @flow
 
+/* eslint-disable camelcase */
```
mjangda: Should probably just move this to apply to lines 175-180?

src/bin/vip-wp.js
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env node
 // @flow
 
+/* eslint-disable camelcase */
```
mjangda: Similar to the other comment, can we apply this inline where it's relevant?

src/lib/analytics/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
mjangda: Similar to the other comment, can we apply this inline where it's relevant?

src/bin/vip-wp.js
```diff
@@ -114,28 +113,6 @@ const getTokenForCommand = async ( appId, envId, command ) => {
 		} );
 };
 
-const cancelCommand = async ( guid ) => {
```
mjangda, line 117: Should we keep this for when we want to add cancel support?

mjangda (commented):
One minor thing, but otherwise looks great!

src/bin/vip-wp.js
markowsiak: As far as I could tell that function is not used anywhere 🤷‍♂️  but I could be missing some context?

mjangda: Sorry, I meant more that if we want to add a `vip wp-cancel` command, this would be useful.

markowsiak: I see what you mean.  I can revert it 👍

mjangda (approved):
❤️

PR #610, opened 2020-12-03
SQL Import: Extend command wrapper's requireConfirm functionality

jblz:
## Description

Currently, passing the `requireConfirm` option to the `command` function for anything outside the `sync` command fails:

```
TypeError: Cannot destructure property 'backup' of 'options.env.syncPreview' as it is undefined.
at Args._args.default.argv (/Users/jeff/Documents/gitRepos/vip/dist/lib/cli/command.js:382:11)
```

In order to use it for the sql import, it needs to be enhanced to only check for appropriate inputs and to display only appropriate output.

This is accomplished here by passing a `module` slug to `command` so we can differentiate the `requireConfirm` handling depending on which command invoked it (but sill maintain a standardized API and tracking).

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `API_HOST=http://localhost:4000 ./dist/bin/vip-import-sql.js @5000.production /path/to/a/dbdump.sql`
    1. Verify you are prompted with an "are you sure" message
    1. Verify your sql file is shown in the prompt
1. The `sync` command should be unaffected & display the output as it currently does

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -17,7 +17,6 @@ import gql from 'graphql-tag';
 import command from 'lib/cli/command';
 import { currentUserCanImportForApp, isSupportedApp } from 'lib/site-import/db-file-import';
 import { uploadImportSqlFileToS3 } from 'lib/client-file-uploader';
-import { formatData } from '../lib/cli/format';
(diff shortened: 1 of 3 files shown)


elazzabi (approved):
Tested locally, looks good :shipit:

PR #612, opened 2020-12-03
SQL Imports: correct the output when an import is queued

jblz:
## Description

Change the copy of the success case of the sql import command. No functionality is altered in this PR.

## Steps to Test

1. Verify copy will make sense to testers and appropriately sets expectations for the tool at this stage of development.

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -112,8 +112,13 @@ command( {
 			err( `StartImport call failed: ${ gqlErr }` );
 		}
 
-		console.log( 'Your site is being prepared for the import.' );
-		console.log( '🚧 🚧 🚧 Your database file is importing! 🚧 🚧 🚧\nFeel free to exit this tool if you\'d like. Your import will continue in the background.' );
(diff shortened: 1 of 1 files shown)


src/bin/vip-import-sql.js
```diff
@@ -112,8 +112,13 @@ command( {
...
 		}
 
-		console.log( 'Your site is being prepared for the import.' );
-		console.log( '🚧 🚧 🚧 Your database file is importing! 🚧 🚧 🚧\nFeel free to exit this tool if you\'d like. Your import will continue in the background.' );
+		console.log( '🚧 🚧 🚧 Your sql file import is queued 🚧 🚧 🚧' );
+
+		console.log(
+			`VIP Testers: Thank you for testing!
+			We are currently working on adding support for showing the import status here.
+			For now, you can monitor the \`import_progress\` site meta.`
```
elazzabi: Given this will be public, I think we should remove the mention of the site meta, etc...

elazzabi (dismissed):
Left a small comment, other than that, it looks good 👍

elazzabi:
Thanks. It looks great!

elazzabi (approved):

PR #615, opened 2020-12-08
SQL Import: Fix async bug with validation

jblz:
## Description

I noticed the validation messages coming in out of order when run as a part of the import module.  The validation module is not blocking execution whilst awaiting the callback tied to the `close` event on the `readline.createInterface` instance.

This "promisifies" it and awaits the results prior to handing control back to the calling function.

## Steps to Test

* `./dist/bin/vip-import-validate-sql.js` should "still work" and be unaffected by this change.
* `./dist/bin/vip-import-sql.js` should "still work" and validation messages should not be out of order


markowsiak (approved):
🚢

PR #617, opened 2020-12-09
Search & Replace: Provide control over the output file

jblz:
## Description

As noted in #605, we want to avoid clobbering files for our output & "midput" copies of the input.  This change provides the caller control over where we write with some smart defaults.

Creates a temp directory using [platform-agnostic tooling built into node](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback)

### API

* When `inPlace` is truthy, copy the input file to the temp directory & write to the input file path as before
* When `inPlace` is falsey:
    * When `output` is unspecified, use `process.stdout` (so we can leverage other CLI tools as needed akin to the go binary)
    * When `output` is "path-like", treat it as a file name and write there
    * When `output` is an object, treat it as a write stream and write there
    * When `output` is specified as anything else (including `undefined`), write to a file created in the temporary directory

### CLI

* If `--in-place` is passed, copy the input file to the temp directory & write to the input file path as before
* If `--in-place` is not passed:
    * omitting the "output" flag writes to stdout (like the go binary)
    * `--output /tmp/testing.sql` writes to that file
    * `--output` writes to `${ tmpDirWithPrefix }${ path.sep }${ path.basename( filename ) }`

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `DEBUG=@automattic/vip:lib:search-and-replace API_HOST=http://127.0.0.1:4000 ./dist/bin/vip-search-replace.js /path/to/test.sql --search-replace="xyz.com,abc.com"` with various combinations of `--output` & `--in-place` flags
1. Verify results are intuitive and correct


src/lib/search-and-replace.js
```diff
@@ -86,24 +114,30 @@ export const searchAndReplace = async (
...
+	} );
 
 	const replacedStream = await replace( readStream, replacements, binary );
 	const result = await new Promise( ( resolve, reject ) => {
 		replacedStream
 			.pipe( writeStream )
 			.on( 'finish', () => {
 				console.log( chalk.green( 'Search and Replace Complete!' ) );
-				resolve( outputFile );
+				resolve( output );
```
jblz: It'd be good etiquette to delete our temp files when we're finished with them...
We can include a `cleanUp` callback in this `resolve` statement so the caller can run it when its finished with the data.

src/lib/search-and-replace.js
```diff
@@ -86,24 +114,30 @@ export const searchAndReplace = async (
...
+		await confirm(
+			[],
+			'Are you sure you want to run search and replace on your input file? This operation is not reversible.'
+		);
+	}
 
-	console.log( chalk.blueBright( 'Input File Path: ' ), inputFile );
-	console.log( chalk.blueBright( 'Output File Path: ' ), outputFile );
+	const tmpDir = fs.mkdtempSync( path.join( os.tmpdir(), 'vip-search-replace-' ) );
+	debug( `Created a directory to hold temporary files: ${ tmpDir }` );
```
markowsiak: Should this be moved into `getReadAndWriteStreams`?

markowsiak (commented):
Looks good to me 👍

src/lib/search-and-replace.js
jblz: done. thanks!

PR #622, opened 2020-12-14
Add a validation to check each line for ENGINE!=InnoDB

markowsiak:
## Description

We want to ensure that our SQL validator ensures customers are using the InnoDB Engine on their tables.
This validation checks for ENGINE != InnoDB and warns the user of which lines in the SQL dump are impacting the result.

## Steps to Test

Example:

1. Check out this PR
1. `npm ci` if you haven't in a while
1. Test a file that passes: `npm run build && VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-validate-sql ./__fixtures__/client-file-uploader/db-dump-ipsum-67mb.sql` and note that no validations are triggered.
1. Test a file that fails: `npm run build && VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-validate-sql ./__fixtures__/client-file-uploader/db-dump-ipsum-67mb-myisam.sql` and note that the validation fails and shows two line numbers where ENGINE=MyISAM

---

### DIFF
--- a/__tests__/lib/validations/sql.js
+++ b/__tests__/lib/validations/sql.js
@@ -45,31 +45,31 @@ describe( 'lib/validations/sql', () => {
 		it( 'an error, and exits with code 1', () => {
 			expect( mockExit ).toHaveBeenCalledWith( ERROR_CODE );
 		} );
-		it( 'USE statements', () => {
+		it( 'instances of USE statements', () => {
 			expect( output ).toContain( 'USE statement on line(s) 3' );
 		} );
-		it( 'CREATE DATABASE statements', () => {
+		it( 'instances of CREATE DATABASE statements', () => {
 			expect( output ).toContain( 'CREATE DATABASE statement on line(s) 1' );
 		} );
-		it( 'TRIGGER statements', () => {
+		it( 'instances of TRIGGER statements', () => {
 			expect( output ).toContain( 'TRIGGER statement on line(s) 16' );
 		} );
-		it( 'DROP DATABASE statements', () => {
+		it( 'instances of DROP DATABASE statements', () => {
 			expect( output ).toContain( 'DROP DATABASE statement on line(s) 27' );
 		} );
-		it( 'ALTER USER statements', () => {
+		it( 'instances of ALTER USER statements', () => {
 			expect( output ).toContain( 'ALTER USER statement on line(s) 25' );
 		} );
-		it( 'no DROP TABLE IF EXISTS statements', () => {
+		it( 'no instances of DROP TABLE IF EXISTS statements', () => {
 			expect( output ).toContain( 'DROP TABLE was not found' );
 		} );
-		it( 'no CREATE TABLE statements', () => {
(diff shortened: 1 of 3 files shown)


src/lib/validations/sql.js
```diff
@@ -48,6 +48,29 @@ const infoCheckFormatter = check => {
 	} );
 };
 
+function checkTablePrefixes( tables ) {
```
markowsiak, line 51: I moved this code up to be closer to where it is called.  I found it useful to move it when reading the file.

src/lib/validations/sql.js
```diff
@@ -142,31 +165,17 @@ const checks: Checks = {
 		excerpt: 'Siteurl/home options',
 		recommendation: '',
 	},
+	engineInnoDB: {
+		matcher: /ENGINE=(?!(InnoDB))/i,
```
jblz, line 169: This, & other patterns like it in this file are subject to false positives when the "offending" pattern is present in the sql dump, but in an innocuous place (like post or comment content bodies).

jblz (changes requested):
If we run this against a file that just adds `ENGINE=MyISAM` inside a post content, validation breaks.  I think we need to be smarter about checking only inside actual DDL statements.

src/lib/validations/sql.js
elazzabi: @jblz I think this goes for all other statements too right?

Can this be fixed by making the regex match starting of line? Or you think there is a better approach?

elazzabi (dismissed):
As discussed in Slack, we can move with a simple check for now and make it more robust in future iterations

__tests__/lib/validations/sql.js
```diff
@@ -0,0 +1,82 @@
...
+			} catch ( e ) {
+				debug( 'Error:', e.toString() );
+			}
+			debug( 'output', output );
+		} );
+		it( 'an error, and exits with code 1', () => {
+			debug( 'b1' );
+			expect( mockExit ).toHaveBeenCalledWith( ERROR_CODE );
+		} );
+		it( 'USE statements', () => {
```
jblz: Looks like this is failing the lint check since pulling in the upstream linter changes https://github.com/Automattic/vip/pull/579

Should just need to add a lower-cased word to start these strings.

__tests__/lib/validations/sql.js
```diff
@@ -0,0 +1,82 @@
...
+				output = '';
+				const badSqlDumpPath = path.join( process.cwd(), '__fixtures__', 'validations', 'bad-sql-dump.sql' );
+				await validate( badSqlDumpPath );
+			} catch ( e ) {
+				debug( 'Error:', e.toString() );
+			}
+			debug( 'output', output );
+		} );
+		it( 'an error, and exits with code 1', () => {
+			debug( 'b1' );
```
jblz: Can we drop these `debug`s? The test runner should report which on fails, right?

PR #624, opened 2020-12-23
ApolloClient: Upgrade to v3

jblz:
## Description

I did this upgrade while testing something that ended up not requiring it, but figured it was worth PR'ing to see how it tests.

I followed this guide:

https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run various `./dist/bin/vip-*` commands
1. Verify data are correctly retrieved from & sent to the API server and the CLI functions as before.
1. Automated tests should pass (save `current-stretch` which is currently broken across the board

## Breaking Changes in ApolloClient

As listed here: https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#breaking-cache-changes
...there are some breaking changes, but none look to be a threat to this project.

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -7,10 +7,8 @@
  * External dependencies
  */
 import fetch from 'node-fetch';
-import { ApolloClient } from 'apollo-client';
-import { HttpLink } from 'apollo-link-http';
-import { InMemoryCache } from 'apollo-cache-inmemory';
-import { onError } from 'apollo-link-error';
+import { ApolloClient, HttpLink, InMemoryCache } from '@apollo/client/core';
+import { onError } from '@apollo/client/link/error';
 import chalk from 'chalk';
 
 /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,72 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@apollo/client": {
+      "version": "3.3.6",
+      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.3.6.tgz",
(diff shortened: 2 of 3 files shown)


src/lib/api.js
```diff
@@ -7,10 +7,9 @@
  * External dependencies
  */
 import fetch from 'node-fetch';
-import { ApolloClient } from 'apollo-client';
-import { HttpLink } from 'apollo-link-http';
-import { InMemoryCache } from 'apollo-cache-inmemory';
-import { onError } from 'apollo-link-error';
+import { ApolloClient, ApolloLink, HttpLink } from '@apollo/client/core';
+import { InMemoryCache } from '@apollo/client/cache';
+import { onError } from '@apollo/client/link/error';
```
nickdaugherty, line 11: It seems like with every version, they flip back and forth between separate and a mono package 😆

jblz: @nickdaugherty  This is the way.

src/lib/api.js
```diff
@@ -7,10 +7,8 @@
  * External dependencies
  */
 import fetch from 'node-fetch';
-import { ApolloClient } from 'apollo-client';
-import { HttpLink } from 'apollo-link-http';
-import { InMemoryCache } from 'apollo-cache-inmemory';
-import { onError } from 'apollo-link-error';
+import { ApolloClient, HttpLink, InMemoryCache } from '@apollo/client/core';
```
jblz, line 10: Since we don't use React in this app, we have to use `@apollo/client/core` instead of `@apollo/client`.

mjangda (approved):
Tested a bunch of different commands; all working smoothly.

Thanks for doing this!

PR #625, opened 2021-01-07
file import validator: Allow SVG files

mjangda:
It's a very common file format.

Matching change has been made in our internal tools as well.

## Steps to Test

1. Pull down PR
1. `mkdir tmp-import`
1. `cd tmp-import`
1. `wget https://s0.wp.com/wp-content/mu-plugins/wpcom-smileys/o_O.svg`
2. `cd ..`
2. `npm run build`
1. Run `./dist/bin/vip-import-validate-files.js tmp-import`
1. Verify no errors.

### DIFF
--- a/src/lib/vip-import-validate-files.js
+++ b/src/lib/vip-import-validate-files.js
@@ -13,6 +13,7 @@ export const acceptedExtensions = [
 	'gif',
 	'png',
 	'bmp',
+	'svg',
 	'tiff', 'tif',
 	'ico',
 	'asf',


jblz:
Change is simple enough.  Just making sure we don't need to do any additional validation on SVG files as they can contain scripts, etc.

https://core.trac.wordpress.org/ticket/24251

mjangda:
Yes, this was discussed internally (https://wp.me/pbHonS-fa#comment-899) and SVGs are considered a customer responsibility (similar to database content and any other code they push).

jblz (approved):
👍

PR #629, opened 2021-01-12
Import SQL: Fix Tracks properties

jblz:
They use `snake_case`

## Description

Our events are getting rejected b/c the prop names are not formatted correctly. This corrects that.

## Steps to Test

* Run an import with `DEBUG=*`
* Inspect tracks events to make sure they look legit
* Make sure the events are no longer rejected


pandah3 (dismissed):
It looks like some of the events are still failing with the same message:
`eventname or eventprops failed "^[a-z_][a-z0-9_]*$" pattern`. 

Will look into it further.

jblz:
Looks good. Thanks for finding and fixing the additional cases.

PR #630, opened 2021-01-12
Import SQL: Add and clean up progress logs

pandah3:
## Description

Our logs for imports are quite sparse right now.
We've gotten requests to separate/better organize logs, especially for imports that complete really fast.

Current Logs|
|---|
|logs main covlogs main 2|

- Adding logs for search and replace progress
- Require confirmation from users to verify search-replace values + send Track event if user bails
- Note the start & stop of each piece for better clarity (S-R, SQL validation, etc.)
- When we exit on error, log a message instead of just exiting suddenly

Also some fixes:
- Upon cancelling the confirmation prompt when the in-place flag is set, exit instead of continuing on
- Ensure we're returning the `outputFileName` when the in-place flag is set

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run an import:
```
VIP_PROXY="<proxy>" node ./dist/bin/vip-import-sql @2909.develop /path/to/file/dump.sql --search-replace="platform-import-test.go-vip.net,test-import.com"
```

4. Verify you see the correct progress logs:

Revised Logs|
|---|
|import logs 1 covimport logs 2|

5. Try the command again and answer `no/false` to some of the prompts. Verify that the program exits
6. This time, try the command with the `--in-place` flag and verify the appropriate logs

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -139,6 +139,7 @@ command( {
 			await trackEventWithEnv( 'import_sql_command_queued' );
 
 			console.log( '🚧 🚧 🚧 Your sql file import is queued 🚧 🚧 🚧' );
+			console.log( `You can check the status of your import via the \`import_progress\` site meta in the VIP Go Admin Console!` );
 		} catch ( e ) {
 			err( e );
 		}
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -48,7 +48,7 @@ command( {
 	envContext: true,
 	module: 'import-sql',
 	requiredArgs: 1,
-	requireConfirm: 'Are you sure you want to import the contents of the provided SQL file?',
+	requireConfirm: 'Are you sure you want to proceed?',
 } )
 	.example( 'vip import sql <file>', 'Import SQL provided in <file> to your site' )
 	.option( 'search-replace', 'Specify the <from> and <to> pairs to be replaced' )
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -80,6 +80,7 @@ command( {
 
 		let fileNameToUpload = fileName;
 
+		// Run Search and Replace if the --search-replace flag was provided
 		if ( searchReplace && searchReplace.length ) {
 			const { outputFileName } = await searchAndReplace( fileName, searchReplace, {
 				isImport: true,
@@ -90,8 +91,10 @@ command( {
 			fileNameToUpload = outputFileName;
 		}
 
+		// Run SQL validation
 		await validate( fileNameToUpload, true );
 
+		// Call the Public API
 		const api = await API();
 
 		try {
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -48,7 +48,7 @@ command( {
 	envContext: true,
 	module: 'import-sql',
 	requiredArgs: 1,
-	requireConfirm: 'Are you sure you want to proceed?',
+	requireConfirm: 'Are you sure you want to import the contents of the provided SQL file?',
 } )
 	.example( 'vip import sql <file>', 'Import SQL provided in <file> to your site' )
 	.option( 'search-replace', 'Specify the <from> and <to> pairs to be replaced' )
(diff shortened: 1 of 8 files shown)


src/lib/search-and-replace.js
```diff
@@ -123,6 +134,8 @@ export const searchAndReplace = async (
 	{ isImport = true, inPlace = false, output = process.stdout }: SearchReplaceOptions,
 	binary: string | null = null
 ): Promise<SearchReplaceOutput> => {
+	console.log( `${ 'Starting Search and Replace...' }` );
```
elazzabi:
```suggestion
	console.log( 'Starting Search and Replace...' );
```

src/lib/search-and-replace.js
```diff
@@ -143,11 +156,41 @@ export const searchAndReplace = async (
...
+	const yes = await confirm( [
+		{
+			key: 'From',
+			value: `${ chalk.cyan( replacements[ 0 ] ) }`,
+		},
+		{
+			key: 'To',
+			value: `${ chalk.cyan( replacements[ 1 ] ) }`,
+		},
+	], 'Proceed with the following Search and Replace values?' );
```
elazzabi:
```suggestion
	], 'Proceed with the following values?' );
```

src/lib/search-and-replace.js
```diff
@@ -143,11 +156,41 @@ export const searchAndReplace = async (
...
+		console.log( `${ chalk.red( 'Cancelling' ) }` );
+
+		await trackEvent( 'search_replace_cancelled', { is_import: isImport, in_place: inPlace } );
+
+		process.exit();
+	}
+
 	if ( inPlace ) {
-		await confirm(
+		const approved = await confirm(
```
elazzabi, line 179: Isn't this supposed to be a confirmation step for running search-replace? Is the one added is to confirm values and this one to confirm the search-replace?

If yes, I guess it would be better to have both in the same step (printing the `from:` `to:` + a message to confirm the from/to and start the search-replace)

src/bin/vip-import-sql.js
```diff
@@ -48,7 +48,7 @@ command( {
 	envContext: true,
 	module: 'import-sql',
 	requiredArgs: 1,
-	requireConfirm: 'Are you sure you want to import the contents of the provided SQL file?',
+	requireConfirm: 'Are you sure you want to proceed?',
```
elazzabi: I think the message we had before is more descriptive than this one ("proceed" vs "import the content of the provided SQL file")

src/bin/vip-import-sql.js
```diff
@@ -135,7 +138,9 @@ command( {
 
 			await trackEventWithEnv( 'import_sql_command_queued' );
 
-			console.log( '🚧 🚧 🚧 Your sql file import is queued 🚧 🚧 🚧' );
+			console.log( '\n🚧 🚧 🚧 Your sql file import is queued 🚧 🚧 🚧' );
+			console.log( '--> Check the status of your import via the \`import_progress\` site meta in the VIP Go Admin Console!\n' );
```
elazzabi:
```suggestion
			console.log( '--> Check the status of your import via the \`import_progress\` site meta in the VIP internal console!\n' );
```

Avoiding using internal names outside. Also, we should remove this once we deploy a public version (outside of `-dev`)

elazzabi (changes requested):
Left some comments on this. Let me know what do you think about them!

src/lib/search-and-replace.js
pandah3: Oop, I initially added a `chalk.cyan()` there but took it out, and forgot to change this back! Fixed!

src/lib/search-and-replace.js
pandah3: Removed!

src/bin/vip-import-sql.js
pandah3: Oops, got it! I added a comment as well to remind ourselves to remove it later on.

src/bin/vip-import-sql.js
pandah3: Yeah, I changed the wording because I added more confirmation prompts to the logging, so this initial one seemed a bit misleading (we're not importing the file right away- we do some checks first, and then we have a final prompt asking if you want to continue with the import).

Not super important, but I thought leaving it simpler was better. I changed it back to the original message for now- we can always play around with the wording!

src/lib/search-and-replace.js
pandah3: This confirmation prompt is if the `--in-place` flag is set. So for regular search-replace, this prompt won't show up (but the prompt displaying the `from/to` values will), but if `search-replace` and `in-place` are both set, then both prompts will show up. Hope that clarifies things!

pandah3:
Thank you for the review! Tests were also failing, but I finally got them working 😌
Should be good to go now 👍

elazzabi (approved):
Yay thanks for the fixes @pandah3 ! Looks good ❤️

PR #631, opened 2021-01-12
Tracks: Show an error via debug when a tracks event name or prop is invalid

jblz:
## Description

As shown in #629, there's currently no visibility into events that will be rejected by the tracks service due to naming issues.  This adds a check that the event name and all property names conform to the proper pattern & shows a debug error if not.

## Steps to Test

1. Check out PR.
2. Add some events like: `await trackEvent( 'ThisEventWillBeRejected' );` & `await trackEventWithEnv( 'thiseventwillberejectedtoo', { SnakeCase: 'nope' } );`
1. Run `npm run build`
1. Run the code path that fires the events with `DEBUG=@automattic/vip:analytics:clients:tracks`
1. Verify the error is shown

### DIFF
--- a/src/lib/analytics/clients/tracks.js
+++ b/src/lib/analytics/clients/tracks.js
@@ -12,6 +12,8 @@ const debug = require( 'debug' )( '@automattic/vip:analytics:clients:tracks' );
  */
 import type { AnalyticsClient } from './client';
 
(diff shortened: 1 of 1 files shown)


mjangda (approved):

mjangda:
We should get this upstream in https://github.com/Automattic/node-tracks as well.

PR #632, opened 2021-01-12
Setup dev release 2 dev1

markowsiak:
Setting up a development release of v2.0.0-dev1

### DIFF
--- a/.publishrc
+++ b/.publishrc
@@ -5,11 +5,11 @@
     "untrackedFiles": true,
     "sensitiveData": true,
     "branch": "master",
-    "gitTag": true
+    "gitTag": false
   },
   "confirm": true,
   "publishCommand": "npm publish",
-  "publishTag": "beta",
+  "publishTag": "next",
   "prePublishScript": "npm test",
   "postPublishScript": false
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@automattic/vip",
-  "version": "1.13.0-0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -102,7 +102,8 @@
     "socks-proxy-agent": "^5.0.0",
     "update-notifier": "4.1.0",
     "uuid": "3.4.0",
-    "xml2js": "^0.4.23"
+    "xml2js": "^0.4.23",
+    "keytar": "5.4.0"
```
jblz:
```suggestion
    "xml2js": "^0.4.23"
```

This change seems unrelated (unless I've missed something)

jblz (commented):
Looks good other than the stray keytar package inclusion.

package.json
markowsiak: Yeah I'm not sure why that's in there tbh, I'll check it out.

.publishrc
```diff
@@ -5,11 +5,11 @@
     "untrackedFiles": true,
     "sensitiveData": true,
     "branch": "master",
-    "gitTag": true
+    "gitTag": false
   },
   "confirm": true,
   "publishCommand": "npm publish",
-  "publishTag": "beta",
+  "publishTag": "next",
```
jblz, line 12: `Check that git tag matches version specified in the package.json`

Wouldn't we still want this?

markowsiak: Just following what is in the CONTRIBUTING.md.  `next` for .publishrc and `-dev1` in the version property.

PR #633, opened 2021-01-12
SQL Search & Replace: Don't specify stream encoding

jblz:
## Description

SQL files can contain [sections of binary](https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html).  When we specify `utf8` encoding, those binary sections are being erroneously encoded as text.

This change specifies no encoding for any of the streams in the search & replace lib to access the raw streams. Hopefully that suffices for our purposes :)

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run imports with and without binary columns.


elazzabi (approved):
Looks good 🚢

PR #634, opened 2021-01-15
SQL Import: Error out if the file is too large

jblz:
## Description

We want to limit the size of the file we permit to be uploaded and imported.

Currently, the limit is arbitrarily set at 10 GiB (10 * 1024 * 1024 * 1024 );

## Steps to Test

1. Generate a large empty file: `dd if=/dev/zero of=/tmp/zero-filled-file-11gb.sql bs=64m count=176` (Linux users probably have to capitalize the `M` in the `bs` flag)
1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js @0000.production /tmp/zero-filled-file-11gb.sql`
1. Verify the error is thrown and copy is good and actionable
1. Repeat with a file smaller than the threshold and confirm it does not error out.

### DIFF
--- a/__tests__/lib/client-file-uploader.js
+++ b/__tests__/lib/client-file-uploader.js
@@ -12,14 +12,6 @@ import { getFileMD5Hash, getFileMeta, getPartBoundaries } from 'lib/client-file-
 
 describe( 'client-file-uploader', () => {
 	describe( 'getFileMeta()', () => {
-		it( 'should fail on empty file', async () => {
-			const fileName = '__fixtures__/client-file-uploader/emptyfile.txt';
-			expect.assertions( 1 );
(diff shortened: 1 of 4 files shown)


__tests__/lib/client-file-uploader.js
```diff
@@ -12,14 +12,6 @@ import { getFileMD5Hash, getFileMeta, getPartBoundaries } from 'lib/client-file-
 
 describe( 'client-file-uploader', () => {
 	describe( 'getFileMeta()', () => {
-		it( 'should fail on empty file', async () => {
```
jblz, line 15: This responsibility has been shifted to the caller.

dchymko (approved):
LGTM. Tested locally with
- 10GB+ file: error
- 0-length file: error
- 100k file: success

PR #635, opened 2021-01-18
SQL Import: Move some debug output to debug command

jblz:
## Description

In order to decrease the default verbosity, change some `console.log`s to `debug`. That way, we can still see the debug output as needed, but customers are not subjected to it.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run an import (with `./dist/bin/vip-import-sql.js` ... etc.)
    * Verify only information that is pertinent to the customer is displayed
1. Repeat with the `DEBUG=*` prefix
    * Verify debug data are included


pandah3 (approved):
Tested and working as expected 👍 
:shipit:

PR #636, opened 2021-01-18
Add auto formatting

joshbetz:
Fixes all lint errors and adds auto formatting via wp-prettier.


package.json
```diff
@@ -92,6 +92,7 @@
     "ini": "1.3.6",
     "json2csv": "3.11.5",
     "jwt-decode": "2.2.0",
+    "keytar": "5.4.0",
```
jblz: Where is the `keytar` dependency coming from?  Happened to @markowsiak the other day as well.

joshbetz: Hmm. It's an optional dependency already https://github.com/Automattic/vip/blob/master/package.json#L108

It seems that `npm` is automatically moving it into dependencies for some reason?

markowsiak: I didn't see explicitly why either.
I just thought if npm was going to move it to be non-optional, then it would remove it from the optional dep section as well 🤷  (unless to preserve some backwards compatibility with other versions of npm?)

jblz (approved):
💥

PR #637, opened 2021-01-19
SQL Import: Fix help prompts for imports and S-R

pandah3:
## Description

- Add examples for the SQL import command
- Add examples for standalone Search and Replace command
- Expose the `output` flag for S-R when it's run during an import (we're currently only exposing it for standalone S-R commands)
- Fix wording of descriptions to be more user-friendly + add warning about `no space` for S-R params

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `VIP_PROXY="<proxy>" node ./dist/bin/vip-import-sql -h`
1. Verify help output looks good:
help examples

5. Run `VIP_PROXY="<proxy>" node ./dist/bin/vip-search-replace -h` and verify help output looks good


src/bin/vip-import-sql.js
```diff
@@ -42,6 +42,32 @@ const err = message => {
...
+		'       * Ensure there are no spaces between your search-replace parameters',
+	},
+	// `in-place` flag
+	{
+		usage: 'vip import sql @mysite.develop <file.sql> --search-replace="<from,to>" --in-place',
+		description: 'Search and Replace on the input <file.sql>, then import the replaced file to your site',
+	},
+	// `output` flag
+	{
+		usage: 'vip import sql @mysite.develop <file.sql> --search-replace="<from,to>" --output="<output.sql>"',
```
elazzabi: I think we don't need the `<` and `>` in the `<from,to>` right?

If that's correct, I wonder if we should remove them from the example in order to avoid any confusion that it may cause to users 🤔

elazzabi (approved):
Looks good to me, I left a comment about the `<` and `>`.

Also, is the `/bin/go-search-replace` binary file is needed? Or was it wrongly committed to this PR?

elazzabi (dismissed):
Looks good to me, I left a comment about the `<` and `>`.

Also, is the `/bin/go-search-replace` binary file is needed? Or was it wrongly committed to this PR?

**Heh double comment for some reason**

jblz:
+1 we should not commit the search and replace binary -- mostly since it's OS-specific

src/bin/vip-import-sql.js
pandah3: Makes sense! Removed!

pandah3:
Argh, that pesky file! I excluded it from the commit, but it somehow made its way!
Removed! Merging :shipit:

PR #614, opened 2020-12-08
SQL Import: Poll for import status after kick off & in separate command

jblz:
## Description

Uses the job status from GraphQL to inform the user about the status of the import as it works through the system.

It leverages both the jobs system and the import_status site meta to relay the progress of the current import and to report failures to the user.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run an import e.g. `./dist/bin/vip-import-sql.js @1000.production --search-replace="from.example.com,to.example.com" ~/Downloads/dbdump.sql`
    * Confirm job status polls appropriately
    * Confirm a successful import is reported correctly
1. Repeat the above, and exit the CLI prior to job completion (or proceed in a separate terminal session)
1. Run `./dist/bin/vip-import-sql-status.js @1000.production`
    * Confirm job status polls appropriately as above
1. Attempt to import various files that you know will fail server-side. (e.g. manually alter a db dump file to insert a row with a duplicate unique key)
    * Confirm the error is provided to you and is actionable.

PR #639, opened 2021-01-19
SQL Imports: Simplify import update progress logs

pandah3:
## Description

Simplifying the user output logs for the SQL import progress.
We also want the output to visually match our other processes (e.g.- data sync).

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql @103.staging /path/to/file.sql --search-replace="hello.com,goodbye.com"`
1. Verify update progress logs look good
1. Try without the S-R flag, and verify that you don't see any of the S-R-related output

### DIFF
--- a/src/lib/cli/progress.js
+++ b/src/lib/cli/progress.js
@@ -30,7 +30,7 @@ const marks = {
 	running: chalk.blueBright( sprite.next().value ),
 	success: chalk.green( '✓' ),
 	failed: chalk.red( '✕' ),
-	unknown: chalk.yellow( '✕' ),
+	skipped: chalk.green( '✕' ),
 };
 
 // Various action steps for SQL imports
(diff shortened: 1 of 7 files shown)


dchymko:
Tested locally and LGTM so far. What else do we need to add here?

pandah3:
Thank you for testing!
Added lots more changes to simplify the output and have the progress logs match data sync's.
Needs a bit more fine-tuning, and then it should be good to go!

src/lib/cli/progress.js
```diff
@@ -0,0 +1,83 @@
+// @flow
+
+/**
+ * External dependencies
+ */
+import chalk from 'chalk';
+import { stdout as progressLog } from 'single-line-log';
+
+// Progress spinner
+const sprite = {
```
elazzabi, line 10: This `progress` component is useful. I think it makes sense to use the same component for sync progress too (in a future PR)

src/lib/cli/progress.js
```diff
@@ -0,0 +1,83 @@
...
+			done: false,
+		};
+	},
+};
+
+// Icons for different action states/status
+const marks = {
+	pending: '○',
+	running: chalk.blueBright( sprite.next().value ),
+	success: chalk.green( '✓' ),
```
elazzabi, line 31: Should we have an extra mark for `skipped`? I guess not all imports will run with search-replace so it may be a good idea to mark that step as `skipped` for those imports.

elazzabi (commented):
Looks good to me so far. I didn't run the test locally yet, but I left a few non-blocking comments.

Let us know when you're confident with the change and you want us to run it end-to-end.

src/lib/cli/progress.js
pandah3: Yup, definitely!

src/lib/cli/progress.js
pandah3: Good idea! Added 👍

pandah3:
I'm having trouble getting the spinner for the `running` status to run. I've tried iterating through the array in different ways, but it doesn't seem to be incrementing each time it gets called 😞

Besides that, I think this is ready!

jblz:
> I'm having trouble getting the spinner for the `running` status to run. 

Take a look at #614. I've reworked the sprite into a class that's pretty easy to work with. It progresses the pointer each time it's printed. As the comment says, we will probably want to throttle the `next` function to control how quickly the animation runs, but it works pretty well.

src/lib/cli/progress.js
```diff
@@ -0,0 +1,96 @@
...
+
+// Various action steps for SQL imports
+const steps = {
+	replace: 'Performing Search and Replace',
+	validate: 'Validating SQL',
+	upload: 'Uploading file to S3',
+	startImport: 'Starting import',
+	import: 'Importing...',
+	// Add more as needed
+};
```
jblz, line 44: These strings are specific to imports, but the file otherwise seems generalized.

pandah3: Yup, I'll generalize these and consolidate the sprite stuff from master + this branch + progress logic from data sync) in another PR.

pandah3:
Marking for retro-review.

PR #640, opened 2021-01-19
Block import of multisite into single site and vice versa

markowsiak:
## Description

We should fail in the instances that a:

* multisite sql dump file was provided for a single site
* single site sql dump file was provided for a multisite

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Execute an import that would pass -- script continues with import
1. Execute an import that would fail because the SQL dump is multi and the site is not -- scripts exits
1. Execute an import that would fail because the SQL dump is single site and the site is multisite -- script exits

Example commands coming!

---

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -8,7 +8,6 @@
 /**
  * External dependencies
  */
-import chalk from 'chalk';
 import gql from 'graphql-tag';
 import debugLib from 'debug';
 
@@ -18,10 +17,13 @@ import debugLib from 'debug';
 import command from 'lib/cli/command';
 import { currentUserCanImportForApp, isSupportedApp, SQL_IMPORT_FILE_SIZE_LIMIT } from 'lib/site-import/db-file-import';
 import { getFileSize, uploadImportSqlFileToS3 } from 'lib/client-file-uploader';
-import { trackEvent } from 'lib/tracker';
-import { validate, getReadInterface } from 'lib/validations/sql';
+import { trackEventWithEnv } from 'lib/tracker';
+import { validate } from 'lib/validations/sql';
+import { isMultiSiteInSiteMeta } from 'lib/validations/is-multi-site';
+import { isMultiSiteDumpFile } from 'lib/validations/is-multi-site-sql-dump'
 import { searchAndReplace } from 'lib/search-and-replace';
 import API from 'lib/api';
+import * as exit from 'lib/cli/exit';
 
 /**
  * - Include `import_in_progress` state & error out if appropriate (this likely needs to be exposed in the data graph)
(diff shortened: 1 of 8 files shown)


dchymko:
Just thinking out loud here, I wonder if we could get the multisite value first and then pass that into the validate function. It would keep the checks all in there and only scan through the file once?

src/bin/vip-import-sql.js
```diff
@@ -35,13 +37,35 @@ const appQuery = `
...
+
+	const fileSize = await getFileSize( fileName );
+
+	if ( ! fileSize ) {
+		exit.withError( `File '${ fileName }' is empty.` );
+	}
+
+	if ( fileSize > SQL_IMPORT_FILE_SIZE_LIMIT ) {
+		exit.withError(
+			`The sql import file size (${ fileSize } bytes) exceeds the limit the limit (${ SQL_IMPORT_FILE_SIZE_LIMIT } bytes).` +
```
elazzabi:
```suggestion
			`The sql import file size (${ fileSize } bytes) exceeds the limit (${ SQL_IMPORT_FILE_SIZE_LIMIT } bytes).` +
```
Fixing a small typo here

src/bin/vip-import-sql.js
```diff
@@ -35,13 +37,35 @@ const appQuery = `
...
+				'Please split it into multiple files or contact support for assistance.' );
```
elazzabi, line 66: > Please split it into multiple files

Is this an opportunity to provide a link to the documentation on how you can do it? Or we can just leave "Contact support for assistance" now.

src/lib/cli/exit.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * External dependencies
+ */
+import chalk from 'chalk';
+
+/**
+ * Internal dependencies
+ */
+
+export function withError( message: string ) {
```
elazzabi, line 15: That's a clever way! I see that `exit.withError()` is much better than always calling `process.exit()`... Thanks for that! ❤️

src/lib/validations/line-by-line.js
```diff
@@ -0,0 +1,84 @@
...
+	} );
+}
+
+export async function getReadInterface( filename: string ) {
+	let fd;
+	try {
+		fd = await openFile( filename );
+	} catch ( e ) {
+		console.log( chalk.red( 'Error: ' ) + 'The file at the provided path is either missing or not readable.' );
+		console.log( 'Please check the input and try again.' );
```
elazzabi: Is this an opportunity to use the `exit.withError()` ?

elazzabi (dismissed):
Left a few non-blocking comments. Otherwise, looks good to me!

src/bin/vip-import-sql.js
markowsiak: TBH I'm open to either.  This code was moved in this PR but not fabricated.  Happy to adjust as necessary.

src/lib/validations/line-by-line.js
markowsiak: I will check this out, but I believe you're correct 👍

elazzabi (approved):
Gave this another look, it looks good 🎉

PR #644, opened 2021-01-25
Forno 493/lowercase environment name

chrean:
## Description

Since our internal API automatically converts mixed case environment names to lowercase, we want to convert them 
## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-wp.js @3203.Sit -- option get home` (note the uppercase `Sit`)
1. Verify the command is working. Without this PR, the above command would return `command not found` .


elazzabi (approved):
Looks good <3

PR #645, opened 2021-01-25
Clean up search and replace pairs when importing

dchymko:
## Description
Removes spaces and other whitespace/non-printable chars from the `--search-replace` pairs when running an import.  This will prevent problems such as cut and paste errors, spaces between pairs( `www.domain1.com, www.domain2.com`) , etc

Note: there is some redundancy in the regex pattern between the low (1-32) chars and the "\s" pattern but I thought it was best to have both. 

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql @1000.dev --search-replace="www.dom ain1.com,    www.domain2.com" ./db.sq`
1. Verify that repairs displayed in confirmation are free from extraneous chars

### DIFF
--- a/src/lib/search-and-replace.js
+++ b/src/lib/search-and-replace.js
@@ -150,7 +150,10 @@ export const searchAndReplace = async (
 	}
 
 	// determine all the replacements required
-	const replacementsArr = pairs.map( str => str.split( ',' ) );
+	const replacementsArr = pairs.map( str => {
+		const filtered = str.replace( /[\s\x00-\x1F\x7F]/g, '' ); //whitespace & char codes 0-32, 127.
+		return filtered.split( ',' );
+	}	);
 	const replacements = flatten( replacementsArr );
 	debug( 'Pairs: ', pairs, 'Replacements: ', replacements );
 
--- a/src/lib/search-and-replace.js
+++ b/src/lib/search-and-replace.js
@@ -150,10 +150,7 @@ export const searchAndReplace = async (
 	}
 
 	// determine all the replacements required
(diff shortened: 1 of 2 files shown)


src/lib/search-and-replace.js
```diff
@@ -150,7 +150,10 @@ export const searchAndReplace = async (
 	}
 
 	// determine all the replacements required
-	const replacementsArr = pairs.map( str => str.split( ',' ) );
+	const replacementsArr = pairs.map( str => {
+		const filtered = str.replace( /[\s\x00-\x1F\x7F]/g, '' ); //whitespace & char codes 0-32, 127.
+		return filtered.split( ',' );
```
elazzabi: Is this better than using `trim()` directly on the strings? Something like:

```js
replacementsArr.map( str => str.split( ',' ).map( s => s.trim() ) );
```

elazzabi (dismissed):
Left a note, but otherwise looks good.

src/lib/search-and-replace.js
dchymko: Good suggestions. Two reasons I didn't go forward for that:
1) It won't remove spaces or special chars from _within_ the string
2) trim only removes whitespace and not the other non-printables. 

But it feels like I'm overthinking this, so I'll probably refactor a bit along the lines of your suggestion.

elazzabi: oh, thank you for working on that! Looks good to me

elazzabi (approved):
Looks good 🎉

PR #646, opened 2021-01-27
Fix SQL validation for imports

pandah3:
## Description

With the changes to SQL validation in #640, we need to identify if the SQL validation is running via an import or as a standalone command.

## Steps to Test=

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql @103.staging /Users/joannelee/Downloads/test.sql --search-replace="hello.com,goodbye.com"`
1. This mainly works in conjuction with #639, where we're changing the output logs for imports

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -232,9 +232,9 @@ export const postValidation = async ( filename: string, isImport: boolean = fals
 		'https://docs.wpvip.com/how-tos/prepare-for-site-launch/migrate-content-databases/' );
 };
 
-const perLineValidations = ( line: string ) => {
+const perLineValidations = ( line: string, runAsImport: boolean ) => {
 	if ( lineNum % 500 === 0 ) {
-		log( `Reading line ${ lineNum } ` );
+		runAsImport ? '' : log( `Reading line ${ lineNum } ` );
 	}
 
 	const checkValues: any = Object.values( checks );
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -247,8 +247,8 @@ const perLineValidations = ( line: string, runAsImport: boolean ) => {
 	lineNum += 1;
 };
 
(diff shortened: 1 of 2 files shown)


src/lib/validations/sql.js
```diff
@@ -247,24 +247,29 @@ const perLineValidations = ( line: string ) => {
 	lineNum += 1;
 };
 
+const execute = ( line: string, isImport: boolean = true ) => {
+	perLineValidations( line, isImport );
+};
+
+const postLineExecutionProcessing = async ( { fileName, isImport }: PostLineExecutionProcessingParams ) => {
```
jblz: Do we need this function any more? It's a 1:1 proxy with `postValidation` as far as I can see.

src/lib/validations/sql.js
```diff
@@ -247,24 +247,29 @@ const perLineValidations = ( line: string ) => {
...
+
 export const staticSqlValidations = {
-	execute: ( line: string ) => {
-		perLineValidations( line );
-	},
-	postLineExecutionProcessing: async ( { fileName, isImport }: PostLineExecutionProcessingParams ) => {
-		await postValidation( fileName, isImport );
-	},
+	execute,
+	postLineExecutionProcessing,
```
jblz: same as above, can we just change this to:
```suggestion
	postLineExecutionProcessing: postValidation,
```

...?

src/lib/validations/sql.js
pandah3: Ah yes, removed!

src/lib/validations/sql.js
pandah3: Fixed!

src/lib/validations/sql.js
pandah3: It looks like we actually need this to pass down `isImport`. Reverted in #639

PR #648, opened 2021-01-27
More robust check for import status object

dchymko:
## Description

When an import is kicked off, we might not have an import progress object, so we need to short-circuit before checking any properties of the object or doing any additional processing.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vipvip-import-sql @siteid ./db123.sql`
1. Verify import doesn't error out with `TypeError: Cannot read property 'started_at' of null`

### DIFF
--- a/src/lib/site-import/status.js
+++ b/src/lib/site-import/status.js
@@ -162,6 +162,7 @@ export async function importSqlCheckStatus( { app, env }: ImportSqlCheckStatusIn
 				debug( { createdAt, dbOperationInProgress, importInProgress, importStepProgress } );
 
 				if (
+					importStepProgress &&
 					importStepProgress.started_at &&
 					importStepProgress.started_at * 1000 > new Date( createdAt ).getTime()
 				) {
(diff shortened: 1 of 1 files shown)


src/lib/site-import/status.js
```diff
@@ -162,6 +162,7 @@ export async function importSqlCheckStatus( { app, env }: ImportSqlCheckStatusIn
 				debug( { createdAt, dbOperationInProgress, importInProgress, importStepProgress } );
 
 				if (
+					importProgress &&
 					importStepProgress.started_at &&
```
dchymko, line 166:
```suggestion
					importStepProgress?.started_at &&
```
I suppose we could use the null chaining operator here? It seems we don't make much use of this elsewhere.

src/lib/site-import/status.js
```diff
@@ -162,6 +162,7 @@ export async function importSqlCheckStatus( { app, env }: ImportSqlCheckStatusIn
 				debug( { createdAt, dbOperationInProgress, importInProgress, importStepProgress } );
 
 				if (
+					importStepProgress &&
 					importStepProgress.started_at &&
```
dchymko:
```suggestion
					importStepProgress?.started_at &&
```
I suppose we could use the null chaining operator here? It seems we don't make much use of this elsewhere.

jblz: done

jblz (approved):
used optional chaining & added a bit more guarding...tested well w/ missing meta

PR #650, opened 2021-01-27
SQL Import: Print the status every .1 seconds

jblz:
## Description

Currently, we only update the "spinner" glyph whenever we get a result from the server. Since our interval is ~ 5 seconds, that means the spinner is not very dynamic.

This change decouples the status printing from the fetch results and prints on a 1 second interval.

The interval is cleared when the promise resolves (or is rejected).

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js`
    * The spinner should update every second (when a task is "running")
    * The command should exit correctly on success & error
1. Run `./dist/bin/vip-import-sql-status.js` while there is an ongoing import
    * The spinner should update every second (when a task is "running")
    * The command should exit correctly on success & error


pandah3 (changes requested):
Tested both commands and they both work as expected!

I think we can make the spinner go even faster so it's a smooth run through of the glyphs rather than cutting off in-between glyphs.

Also, after the first step is complete, the next one just stays `pending` (the circle icon), rather than switching to `running`.

mjangda:
So much much smoother now.

I would put the ^C comment at the end just for better visibility. We could also mention the `vip import sql status` command there.

I'm noticing that some lines are still getting "eaten" by single-line-log though. I can't spot an obvious culprit though.

pandah3 (commented):
The spinner still felt a bit slow for me so I sped it up to 1/10th of a second.

|Before|After|
|---|---|
|spinner|spinner revised|

Haven't touched too much of the other stuff since we'll probably need to adjust them anyways after #655 lands 👍

jblz:
Incorporated into #660

PR #653, opened 2021-01-28
SQL Imports: Reinstitute check for file access in command gate fn

jblz:
## Description

At some point, we lost graceful handling of missing or unreadable files. This brings it back and uses `fs.promises` functions since we've dropped node 8 support.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js` with paths to empty, non-existent, & unreadable files
    * Failure should be graceful & not point to contact support
1. Repeat with a readable file
    * It should not fail


dchymko (approved):
Tested locally, LGTM!
:+1: on the promise cleanup bit as well!

PR #655, opened 2021-01-28
Streamline import progress output

pandah3:
## Description

Changes from #639 +
Flipped progress logs to use the sprite/progress logic in `/lib/cli/format.js` added in #614

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql --search-replace="hello.com,goodbye.com"`

### DIFF
--- a/src/lib/cli/progress.js
+++ b/src/lib/cli/progress.js
@@ -32,9 +32,9 @@ export const setStatusForCurrentAction = ( status, action ) => {
 const completedSteps = [];
 
 export function progress( steps: Object[], runningSprite: RunningSprite ) {
-	const logs = steps.reduce(
-		( carry, step ) => {
-			let statusOfAction;
+	// Iterate over each step and collect the logs to output
+ const reducer = ( accumulator, step ) => {
+		let statusOfAction;
 		statusOfAction = step.status;
 
 		const skipped = `${ step.id }-skipped`;
@@ -63,12 +63,13 @@ export function progress( steps: Object[], runningSprite: RunningSprite ) {
 			statusOfAction = 'pending';
 		}
 
-		const outputStep = carry + `${ getGlyphForStatus( statusOfAction, runningSprite ) } ${ step.name }\n`;
+		const outputStep = accumulator + `${ statusIcon } ${ step.name }\n`;
 
   return outputStep;
-	},
-		''
-	);
- 
+	}
+	
+	const logs = steps.reduce( reducer, '' );
+
+	// Output the logs
  return progressLog( logs );
 }
--- a/src/lib/cli/progress.js
+++ b/src/lib/cli/progress.js
@@ -3,7 +3,6 @@
 /**
  * External dependencies
  */
-import chalk from 'chalk';
 import { stdout as progressLog } from 'single-line-log';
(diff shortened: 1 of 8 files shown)


src/lib/validations/sql.js
```diff
@@ -201,38 +216,27 @@ export const postValidation = async ( filename: string, isImport: boolean = fals
...
-
-			process.exit();
-		}
-
-		console.log( `\n${ chalk.underline( 'Starting the import process...' ) }` );
-		return;
-	}
-
-	console.log( '\n🎉 You can now submit for import, see here for more details: ' +
+	isImport ? '' : console.log( '\n🎉 You can now submit for import, see here for more details: ' +
```
elazzabi: I think this messaging is outdated and we can skip it

src/lib/validations/sql.js
```diff
@@ -177,17 +188,21 @@ const checks: Checks = {
 	},
 };
 
-export const postValidation = async ( filename: string, isImport: boolean = false ) => {
```
elazzabi, line 180: Is removing the default value for `isImport` expected here? (just to be sure we're not getting `undefined` somewhere)

src/lib/validations/sql.js
```diff
@@ -43,9 +52,11 @@ const requiredCheckFormatter = ( check, type ) => {
 	}
 };
 
-const infoCheckFormatter = check => {
+const infoCheckFormatter = ( isImport, check ) => {
```
elazzabi: For this, the `requiredCheckFormatter` function, and the `errorCheckFormatter`, I guess it's better for the `isImport` flag to be the last argument. You also get the benefit that all prior calls will still work as expected (vs this implementation which will break all prior implementations)

elazzabi (dismissed):
Left a few non-blocking comments. Other than that, it looks we can ship this and iterate

src/lib/cli/progress.js
```diff
@@ -0,0 +1,78 @@
...
+	} );
+
+	return currentProgressSteps;
+};
+
+// Instantiate a new instance of the RunningSprite class
+const runningSprite = new RunningSprite();
+
+// Keep track of completed and skipped steps
+const completedSteps = [];
```
jblz, line 34: This value will be shared everyplace that the progress lib module is imported. It would be better to pass around an instance, store the state in a property, & call methods on the instance, that sort of thing.

...so we can use the progress tracking code for the preflight as well as the steps from the server (& even other places in the future).

But I'm ok if that happens in a future PR.

src/lib/validations/sql.js
pandah3: Ah that makes sense! Fixed!

src/lib/validations/sql.js
pandah3: I took it out thinking it was breaking something, but it seems to be okay now. Reinstated!

jblz (approved):

src/lib/cli/progress.js
pandah3: Duly noted!

PR #657, opened 2021-01-28
Add environment launched status to the import sql command

pandah3:
## Description

Exposing the `launched` query field of an environment and displaying that when the `import sql` command is run so users know if their site is launched or not before running an import.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql`
1. Verify you see the launch details before the confirmation prompt

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -348,10 +348,21 @@ args.argv = async function( argv, cb ): Promise<any> {
 
 		switch ( _opts.module ) {
 			case 'import-sql':
-				if ( options.env && options.env.primaryDomain ) {
-					const primaryDomainName = options.env.primaryDomain.name;
+				const site = options.env;
+				if ( site && site.primaryDomain ) {
+					const primaryDomainName = site.primaryDomain.name;
 					info.push( { key: 'Primary Domain Name', value: primaryDomainName } );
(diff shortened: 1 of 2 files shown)


src/lib/cli/command.js
```diff
@@ -348,10 +348,21 @@ args.argv = async function( argv, cb ): Promise<any> {
...
+				const site = options.env;
+				if ( site && site.primaryDomain ) {
+					const primaryDomainName = site.primaryDomain.name;
 					info.push( { key: 'Primary Domain Name', value: primaryDomainName } );
 				}
+
+				// Site launched details
+				const launchedQuery = site.hasOwnProperty( 'launched' );
+
+				if ( launchedQuery ) {
```
mjangda:
```suggestion
				const haveLaunchedField = site.hasOwnProperty( 'launched' );

				if ( haveLaunchedField ) {
```

mjangda (dismissed):
Minor suggestion to improve a var name, but otherwise 👍

src/lib/cli/command.js
pandah3: Changed 👍

PR #658, opened 2021-01-28
Fix search replace output for SQL import command

pandah3:
## Description

#655 adds the logic to display the Search and Replace values if the S-R flag is passed to the import sql command.
We need to account for multiple pairs of S-R values coming in, and not just one pair.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Test w/ one S-R pair and verify output in both the initial summary and the one after the confirmation prompt:
`node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql --search-replace="http://hi.com,https://bye.com"`
1. Test w/ two S-R pairs and verify output: `node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql -s="hi.com,bye.com" -s="hello.com,bye.com"`
1. Test w/ no S-R flag: `node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql`

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -210,9 +210,25 @@ command( {
 
 		// Run Search and Replace if the --search-replace flag was provided
 		if ( searchReplace && searchReplace.length ) {
-			const params = searchReplace.split( ',' );
+			const message = param => {
+				console.log( `        s-r: ${ chalk.blue( param[ 0 ] ) } -> ${ chalk.blue( param[ 1 ] ) }` );
+			}
 
-			console.log( `        s-r: ${ chalk.blue( params[ 0 ] ) } -> ${ chalk.blue( params [ 1 ] ) }\n` );
+			// Only one pair of S-R values specified
+			if ( typeof searchReplace === 'string' ) {
+				const urlPair = searchReplace.split( ',' );
+
+				message( urlPair );
+				console.log();
+			} else {
+				// Multiple pairs of S-R values specified
+				searchReplace.forEach( pair => {
+					const urls = pair.split( ',' )
+
+					message( urls );
+				} )
+				console.log();
+			}
 
 			const { outputFileName } = await searchAndReplace( fileName, searchReplace, {
 				isImport: true,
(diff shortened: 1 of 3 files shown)


elazzabi (dismissed):
Looks good and tested locally 👍

markowsiak (dismissed):
LGTM 🚢

src/bin/vip-import-sql.js
```diff
@@ -214,12 +214,13 @@ command( {
 		console.log( `         to: ${ chalk.cyan( domain ) }` );
 		console.log( `       site: ${ app.name } (${ formatEnvironment( opts.env.type ) })` );
 		if ( searchReplace && searchReplace.length ) {
-			const searchAndReplaceParams = searchReplace.split( ',' );
-			console.log(
-				`        s-r: ${ chalk.blue( searchAndReplaceParams[ 0 ] ) } -> ${ chalk.blue(
-					searchAndReplaceParams[ 1 ]
-				) }`
-			);
+			const searchReplaceList =
```
jblz: It would be nice to consolidate this logic in the command portion below and just pass any derived strings we want to print to the args handler here.

jblz: That said, I'm curious if we really need both.

pandah3: Refactored! I left both in since the initial summary is a bit hard to see once the progress logs get long!

src/lib/cli/format.js
```diff
@@ -167,3 +167,18 @@ export function getGlyphForStatus( status: string, runningSprite: RunningSprite
...
 	}
 }
+
+// Format Search and Replace values to output
+export const formatSearchReplaceValues = ( values, message ) => {
+	// Convert single pair S-R values to arrays
+	const searchReplaceValues = typeof values === 'string' ? [ values ] : values;
+
+	const formattedOutput = searchReplaceValues.map( pairs => {
+		const [ from, to ] = pairs.split( ',' );
```
jblz: What do you think about trimming whitespace proper to output, like so?
```suggestion
		const [ from, to ] = pairs.split( ',' ).map( s => s.trim() );
```

jblz (dismissed):
Awesome!  Left a suggestion about trimmng, but this is looking good.

src/lib/cli/format.js
pandah3: Oop, good catch! Added!

PR #659, opened 2021-01-28
Fix import file path reference to work cross-platform

dchymko:
## Description
The file path concatenation was working incorrectly in Windows causing the import to silently fail. This changes the file basename reference to one that works cross-platform and also adds error output in the case of failure on the file operation.

Tested on
- [x] MacOS
- [x] Windows (v10 Enterprise)
- [x] Linux (Ubuntu 20.10)

## Steps to Reproduce
1. Check out `master` on a Windows machine.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql.js @<siteid> c:\db.sql`
1. Import should halt after the "Compressing the file to <path> prior to transfer..." step

## Steps to Test
1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql.js @<siteid> c:\db.sql`
1. Verify that import file is uploaded, compressed and started successfully.

windows

### DIFF
--- a/src/lib/client-file-uploader.js
+++ b/src/lib/client-file-uploader.js
@@ -101,7 +101,7 @@ export async function getFileMeta( fileName: string ): Promise<FileMeta> {
 	return new Promise( async ( resolve, reject ) => {
 		const fileSize = await getFileSize( fileName );
 
-		const basename = path.posix.basename( fileName );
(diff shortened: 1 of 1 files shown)


jblz (approved):
🚀

PR #660, opened 2021-01-29
SQL Imports: Another iteration on the UX and underlying API

jblz:
## Description

Building on #655, This change accomplishes a few things:

* Move the responsibility for updating the step state to the upper level (e.g. no progress setting scattered throughout client-file-uploader, search & replace, & sql validation functionality)
* Simplify the code needed in the upper level (e.g. only needing to set the success and failure status with the lib handling more of the mechanics)
* Combine the output of the "preflight" steps that are hard-coded in the client with the importing status that comes from the server. (Try as I might, I could not get the single-log-line lib to play nicely unless all the input was combined prior to writing to stdout.)
    * Utilize a `prefix` and `suffix` to enable printing lines before and after the single-log-line writes.
* Make the `ProgressTracker` an object instance and consolidate state inside it
* Pull in enhancements from https://github.com/Automattic/vip/pull/650

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js` (with required fields)
    * Verify the output is an improved experience
1. While an import is running run `./dist/bin/vip-import-sql-status.js`
    * Verify the output is an improved experience
1. Attempt to cause failures
    * They should be handled gracefully and attempt to show actionable steps

### DIFF
--- a/src/bin/vip-import-sql-status.js
+++ b/src/bin/vip-import-sql-status.js
@@ -17,6 +17,7 @@ import * as exit from 'lib/cli/exit';
 import { isSupportedApp } from 'lib/site-import/db-file-import';
 import { importSqlCheckStatus } from 'lib/site-import/status';
 import command from 'lib/cli/command';
+import { ProgressTracker } from 'lib/cli/progress';
 
 const appQuery = `
 id,
@@ -47,7 +48,9 @@ command( {
 
 	await track( 'import_sql_check_status_command_execute' );
 
-	console.log( `Checking the sql import status for env ID: ${ env.id }, app ID: ${ env.appId }` );
+	const progressTracker = new ProgressTracker( [] );
+	progressTracker.prefix = `=============================================================
+Checking the sql import status for env ID: ${ env.id }, app ID: ${ env.appId }:\n`;
 
-	await importSqlCheckStatus( { app, env } );
+	await importSqlCheckStatus( { app, env, progressTracker } );
 } );
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -222,12 +222,15 @@ command( {
 		let fileNameToUpload = fileName;
 
 		const progressTracker = new ProgressTracker( SQL_IMPORT_PREFLIGHT_PROGRESS_STEPS );
-		progressTracker.prefix = `=============================================================
+		const setProgressTrackerPrefixAndSuffix = () => {
+			progressTracker.prefix = `=============================================================
 Processing your sql import for env ID: ${ env.id }, app ID: ${ env.appId }:\n`;
-		progressTracker.suffix = `\n${ getGlyphForStatus(
-			'running',
-			progressTracker.runningSprite
-		) } Loading remaining steps`;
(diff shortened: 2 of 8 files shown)


src/bin/vip-import-sql-status.js
```diff
@@ -47,7 +48,9 @@ command( {
 
 	await track( 'import_sql_check_status_command_execute' );
 
-	console.log( `Checking the sql import status for env ID: ${ env.id }, app ID: ${ env.appId }` );
+	const progressTracker = new ProgressTracker( [] );
+	progressTracker.prefix = `=============================================================
+Checking the sql import status for env ID: ${ env.id }, app ID: ${ env.appId }:\n`;
```
pandah3, line 53: I think we should format and print these details to match the `import sql` command so it looks cleaner.

We can add that in a separate PR and just remove these lines here.

```suggestion
```

src/bin/vip-import-sql.js
```diff
@@ -204,15 +207,31 @@ command( {
...
+		}
+		console.log();
+
+		// NO `console.log` after this point! It will break the progress printing.
 
 		let fileNameToUpload = fileName;
 
+		const progressTracker = new ProgressTracker( SQL_IMPORT_PREFLIGHT_PROGRESS_STEPS );
+		progressTracker.prefix = `=============================================================
+Processing your sql import for env ID: ${ env.id }, app ID: ${ env.appId }:\n`;
```
pandah3: Same thing as above- I think we should add these details to the summary printed initially when you start the command, or to the summary after the confirmation prompt so it looks cleaner.

(We can add that in a separate PR)

```suggestion
```

src/bin/vip-import-sql.js
```diff
@@ -204,15 +207,31 @@ command( {
...
+		progressTracker.suffix = `\n${ getGlyphForStatus(
+			'running',
+			progressTracker.runningSprite
+		) } Loading remaining steps`;
```
pandah3: Noting that I missed spotting this log the first time around, so it may not be too noticeable (although that may be because we didn't bring in the fast spinner changes yet).

pandah3 (commented):
This is amazinggg 🎉 thanks for reworking this into one giant class!!

A couple of things:
- We wanted a final confirmation prompt before starting the import, right 🤦? Fun times with the single line log output again 😭 

- The current output is a bit harsh(?) when the server side progress logs get displayed b/c a giant block of text just appears all of a sudden. I wonder if we should separate the first four steps as `Import preflight` and the rest as `Import status` for better readability? If we add the confirmation step in as well, that'll help break things up too.

- When I pass in a directory as the SQL file, part of the progress still loads:
directory error

We'll probably want to handle that error more gracefully too (in another PR).

dchymko:
> We'll probably want to handle that error more gracefully too (in another PR).
We could also check for that case as well `! fs.statSync(filePath).isDirectory()`

jblz:
Directory ("not a file") gating added in PR #660

elazzabi (approved):
Tested locally! Much better UX overall, thank you ❤️

PR #661, opened 2021-01-29
SQL Import: Add a gate to check if path is a file

jblz:
## Description

If the passed path is not a file (e.g. a directory), throw an error.  See bug report [here](https://github.com/Automattic/vip/pull/660#pullrequestreview-578871140).

directory error

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-sql-import.js` with a readable directory path
    * It should fail with a meaningful error message
1. Run `./dist/bin/vip-sql-import.js` with a readable file path
    * It should not fail due to this change


elazzabi (approved):
Looks good 🎉

PR #664, opened 2021-01-31
Append app & env ids to output

pandah3:
## Description

Display the app id and the env id for a site so users can easily reference the site's ID.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql @103 /Users/joannelee/Downloads/test.sql`
1. Verify app & env ids show


jblz (approved):
Haven't tested this, but lgtm if you're confident in the change.

PR #665, opened 2021-01-31
SQL Import: UX Enhancements including restoring functionality to show upload percentage & API call optimization

jblz:
## Description

* Tweak a pre-flight step from "Uploading file to S3" to simply "Uploading"
* Check the Importing gate prior to the dbOperationInProgress for better messaging
* Include the (approximated) percentage of the upload as a suffix to the `uploading` step
    * Uses a `progressCallback` function in the uploader functionality to call with the percentage
    * The `ProgressTracker` instance uses that value to set a `percentage` property on the step
    * The print formatting functionality appends the `percentage` as a `suffix` to the appropriate `upload` step when it's present and incomplete
* Optimize the GraphQL query for import status to only include the target env & job id
* Only show job creation and completion times if there's job data to report (e.g. if there's no job record, just show that message)
* Improve error handling

|Example of Upload Percentage|
|---|
|Screen Shot 2021-01-31 at 9 01 13 AM|

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js` and `./dist/bin/vip-import-sql-status.js` with various inputs
1. Verify UX is improved and no regressions are discovered

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -125,16 +125,16 @@ const gates = async ( app, env, fileName ) => {
 		importStatus: { dbOperationInProgress, importInProgress },
 	} = env;
 
-	if ( dbOperationInProgress ) {
-		await track( 'import_sql_command_error', { errorType: 'existing-dbop' } );
-		exit.withError( 'There is already a database operation in progress. Please try again later.' );
-	}
-
 	if ( importInProgress ) {
 		await track( 'import_sql_command_error', { errorType: 'existing-import' } );
 		exit.withError(
-			'There is already an import in progress. You can view the status with the `vip import sql status` command.'
+			'There is already an import in progress.\n\nYou can view the status with command:\n    vip import sql status'
 		);
+	}
+
+	if ( dbOperationInProgress ) {
+		await track( 'import_sql_command_error', { errorType: 'existing-dbop' } );
+		exit.withError( 'There is already a database operation in progress. Please try again later.' );
 	}
 };
 
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -277,12 +277,20 @@ Processing the SQL import for your environment...
 
 		const startImportVariables = {};
 
-		// Uploading the SQL file to AWS S3
+		const progressCallback = percentage => {
+			progressTracker.setUploadPercentage( percentage );
+		};
+
 		try {
 			const {
 				fileMeta: { basename, md5 },
 				result,
-			} = await uploadImportSqlFileToS3( { app, env, fileName: fileNameToUpload } );
+			} = await uploadImportSqlFileToS3( {
+				app,
+				env,
+				fileName: fileNameToUpload,
+				progressCallback,
+			} );
 
 			startImportVariables.input = {
 				id: app.id,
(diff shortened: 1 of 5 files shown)


src/bin/vip-import-sql.js
```diff
@@ -72,7 +72,7 @@ const debug = debugLib( 'vip:vip-import-sql' );
 const SQL_IMPORT_PREFLIGHT_PROGRESS_STEPS = [
 	{ id: 'replace', name: 'Performing Search and Replace' },
 	{ id: 'validate', name: 'Validating SQL' },
-	{ id: 'upload', name: 'Uploading file to S3' },
+	{ id: 'upload', name: 'Uploading' },
```
pandah3: Minor suggestion to append `file` or `SQL file`:
```suggestion
	{ id: 'upload', name: 'Uploading file' },
```

src/lib/client-file-uploader.js
```diff
@@ -208,7 +216,11 @@ export async function uploadUsingPutObject( {
 	const progressPassThrough = new PassThrough();
 	progressPassThrough.on( 'data', data => {
 		readBytes += data.length;
-		debug( `${ Math.floor( ( 100 * readBytes ) / fileSize ) }%...` );
+		const percentage = Math.floor( ( 100 * readBytes ) / fileSize ) + '%';
+		debug( percentage );
+		if ( typeof progressCallback === 'function' ) {
+			progressCallback( percentage );
+		}
```
pandah3, line 223: Minor- it'd be nice to wrap this logic into a function since it's used in multiple places!

src/lib/site-import/status.js
```diff
@@ -70,16 +70,25 @@ export type ImportSqlCheckStatusInput = {
...
 			app: { environments },
 		},
 	} = response;
-	const { importStatus, jobs } = environments.find( e => e.id === envId );
-	const importJob = jobs.find( ( { type } ) => type === 'sql_import' );
+	if ( ! environments?.length ) {
+		throw new Error( 'Unable to determine import status from environment' );
+	}
+	const [ environment ] = environments;
+	const { importStatus, jobs } = environment;
```
pandah3, line 85: Will `importStatus` be `undefined` if there's no import currently in progress?

pandah3 (commented):
Love the percentage for the `upload` step!
Thanks for the improvements!

One thing I noticed that might be a bit confusing to users- towards the end of the import, all the progress steps show that they're completed (the green checkmark), but the status is still `running` in the footer summary.

Do we need a `Finishing import` or `Getting site ready` type of final step at the end?
Screen Shot 2021-01-31 at 12 45 12 PM

src/lib/site-import/status.js
jblz: No, it would only be undefined if there's never been an import for the site. Also it looks like it's not referenced in that case in a way that errors:
Screen Shot 2021-01-31 at 6 57 49 PM

I _did_ notice a "Bad Request" error in some cases, but that's not unique to this branch.

src/lib/client-file-uploader.js
jblz: Perhaps in a follow up :)  Thanks for the review!

PR #668, opened 2021-02-01
SQL Import: Show 'Finishing up' when all steps succeed & tidy statusMessage

jblz:
## Description

There's a bit of a lag between the job steps completing and the job actually completing.  This adds a "Finishing Up" message like so:

Screen Shot 2021-02-01 at 1 20 41 AM

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js` with various imputs
1. Verify UX is improved

### DIFF
--- a/src/lib/cli/format.js
+++ b/src/lib/cli/format.js
@@ -128,6 +128,13 @@ export function requoteArgs( args: Array<string> ): Array<string> {
 	} );
 }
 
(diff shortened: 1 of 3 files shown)


elazzabi (approved):
Looks good ❤️

PR #669, opened 2021-02-01
Test for SET @@SESSION.sql_log_bin in sql static validations

elazzabi:
## Description

This adds a test for `SET @@SESSION.sql_log_bin` usually added by `mysqldump` to SQL files. More info https://dev.mysql.com/doc/refman/8.0/en/set-sql-log-bin.html

## Steps to Test

An automated test case was added.

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -124,7 +124,7 @@ const checks: Checks = {
 		outputFormatter: errorCheckFormatter,
 		results: [],
 		message: 'SET @@SESSION.sql_log_bin statement',
-		excerpt: '\'SET @@SESSION.sql_log_bin=0\' statement should not be present (case-insensitive)',
+		excerpt: '\'SET @@SESSION.sql_log_bin\' statement should not be present (case-insensitive)',
 		recommendation: 'Remove these lines',
 	},
 	trigger: {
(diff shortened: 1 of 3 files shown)


src/lib/validations/sql.js
```diff
@@ -118,6 +118,15 @@ const checks: Checks = {
 		excerpt: '\'CREATE DATABASE\' statement should not  be present (case-insensitive)',
 		recommendation: 'Remove these lines',
 	},
+	binaryLogging: {
+		matcher: /SET @@SESSION.sql_log_bin/i,
+		matchHandler: lineNumber => lineNumber,
+		outputFormatter: errorCheckFormatter,
+		results: [],
+		message: 'SET @@SESSION.sql_log_bin statement',
+		excerpt: '\'SET @@SESSION.sql_log_bin=0\' statement should not be present (case-insensitive)',
```
dchymko: Super minor, but we probably don't need the `=0` here.

dchymko (dismissed):
LGTM!

src/lib/validations/sql.js
elazzabi: Oh yeah good catch!

elazzabi: Fixed!

elazzabi:
@markowsiak @jblz forgot to merge this before, so going to merge it now.

PR #673, opened 2021-02-01
2.0.0

jblz:
## Description

🎉 

* Bump version in package.json & package-lock.json
* Change npm publish tag to "latest"
* Remove some missing & unused imports and variables in the S&R lib discovered by the flow test

### DIFF
--- a/.publishrc
+++ b/.publishrc
@@ -9,7 +9,7 @@
   },
   "confirm": true,
   "publishCommand": "npm publish",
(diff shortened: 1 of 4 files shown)


elazzabi:
🎉 
I wonder if we need to change this too https://github.com/Automattic/vip/blob/master/.publishrc#L12? Or will it be part of another PR?

elazzabi (approved):

PR #674, opened 2021-02-01
Fix: SQL file validation output disappearing

elazzabi:
## Description

The SQL file validation logs were disappearing when running an import. This tries to fix it by moving up the check (before initiating the progress)

## Steps to Test

Example:

1. Get a problematic SQL. A file with `ENGINE=MyISAM` should do.
1. Try to run an import with the SQL file. It should fail but nothing is displayed to your screen.
1. Check out PR.
1. Run `npm link`
1. Run the import another time. It should fail but the error is displayed this time.
1. Run an import with a correct SQL file. Everything should work as expected.

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -223,9 +223,16 @@ command( {
 			formatSearchReplaceValues( searchReplace, output );
 		}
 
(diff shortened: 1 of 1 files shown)


jblz (approved):
Lgtm

PR #676, opened 2021-02-01
SQL Import: Fix validation error output

mjangda:
Wrap import validation in try-catch we can stop the progress tracker and properly exit on error.

TODO: return errors instead of console logging them.

## Steps to Test

1. `npm run build`
2. `./dist/bin/vip-import-sql.js @309 --search-replace=example.com,also.example.com __fixtures__/validations/bad-sql-dump.sql`

Should cleanly error out.

Try a successful import as well.

### DIFF
--- a/src/lib/cli/exit.js
+++ b/src/lib/cli/exit.js
@@ -13,8 +13,6 @@ import chalk from 'chalk';
  */
 
 export function withError( message: string ) {
-	let updatedMessage = message;
-
 	console.log( message.toString().replace( /^(Error: )*/, chalk.red( 'Error: ' ) ) );
 	process.exit( 1 );
 }
--- a/src/lib/cli/exit.js
+++ b/src/lib/cli/exit.js
@@ -13,6 +13,8 @@ import chalk from 'chalk';
  */
 
 export function withError( message: string ) {
-	console.log( chalk.red( message.toString().replace( /^(Error: )*/, 'Error: ' ) ) );
+	let updatedMessage = message;
+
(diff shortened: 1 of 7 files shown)


src/lib/cli/exit.js
```diff
@@ -13,6 +13,8 @@ import chalk from 'chalk';
  */
 
 export function withError( message: string ) {
-	console.log( chalk.red( message.toString().replace( /^(Error: )*/, 'Error: ' ) ) );
+	let updatedMessage = message;
```
dchymko: Whats the purpose of this assignment here?

pandah3 (commented):
Woohoo, thank you!

This may have broken the spinner for the progress steps if you skip S-R:
Screen Shot 2021-02-01 at 5 17 05 PM

src/lib/cli/exit.js
mjangda: Oh, I changed my mind halfway through :)

pandah3:
On a valid test file, it's double logging the progress steps for some reason 😭:

Screen Shot 2021-02-01 at 5 26 58 PM

pandah3 (approved):
Tested with:

- Good SQL file
- Good SQL file w/ S-R
- Bad SQL file
- Bad SQL file w/ S-R

+

Tested:
- Standalone SQL validation command w/ both good & bad files

All looking good 🎉

mjangda:
Noting that the approach here isn't ideal since we mix content and presentation, but there are larger problems caused by single-line-log that would take a long time to fix. This should be good enough for now.

PR #677, opened 2021-02-01
Guard against empty importStatus data

dchymko:
## Description
Provides a better error message for import commands run against that don't return an import status (no longer active, etc)

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vipvip-import-sql @siteid ./db123.sql
1. Verify the `Error: Unable to check for existing import or database operations on the site.` error is displayed

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -121,6 +121,12 @@ const gates = async ( app, env, fileName ) => {
 		);
 	}
 
+	if ( ! env?.importStatus ) {
+		await track( 'import_sql_command_error', { errorType: 'empty-import-status' } );
+		exit.withError(
+			'Unable to check for existing import or database operations on the site.'
+		);
+	}
 	const {
 		importStatus: { dbOperationInProgress, importInProgress },
 	} = env;
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -124,7 +124,7 @@ const gates = async ( app, env, fileName ) => {
 	if ( ! env?.importStatus ) {
 		await track( 'import_sql_command_error', { error_type: 'empty-import-status' } );
 		exit.withError(
-			'Unable to check for existing import or database operations on the site.'
+			'Could not determine the import status for this environment.'
 		);
 	}
 	const {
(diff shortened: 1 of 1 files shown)


src/bin/vip-import-sql.js
```diff
@@ -121,6 +121,12 @@ const gates = async ( app, env, fileName ) => {
 		);
 	}
 
+	if ( ! env?.importStatus ) {
+		await track( 'import_sql_command_error', { errorType: 'empty-import-status' } );
```
jblz:
```suggestion
		await track( 'import_sql_command_error', { error_type: 'empty-import-status' } );
```

src/bin/vip-import-sql.js
```diff
@@ -121,6 +121,12 @@ const gates = async ( app, env, fileName ) => {
 		);
 	}
 
+	if ( ! env?.importStatus ) {
+		await track( 'import_sql_command_error', { error_type: 'empty-import-status' } );
+		exit.withError(
+			'Unable to check for existing import or database operations on the site.'
```
jblz: How about:
```suggestion
			'Could not determine the import status for this environment.'
```

jblz (dismissed):

src/bin/vip-import-sql.js
mjangda: Not a blocker, but I find it's helpful to include a "what to do next" style suggestion for errors like this to avoid leaving users in a confused state. Should they run it again? Should they check the environment ID? Should they contact support?

dchymko: > How about:

Looks good. I was trying to stay somehat consistent with the messaging for the other two related errors:
https://github.com/Automattic/vip/blob/1ced866721c9e63b5553e34eb0149239c41fad71/src/bin/vip-import-sql.js#L143

This is better overall though.

>Not a blocker, but I find it's helpful to include a "what to do next" style suggestion for errors like this to avoid leaving users in a confused state. Should they run it again? Should they check the environment ID? Should they contact support?

Good point, I was trying to think of something here, but those are good suggestions, especially around checking the siteID.

src/bin/vip-import-sql.js
dchymko: Ahhm weird, I grabbed it from another spot where they were like that so fixed those as well. (819a49fc881b1c5094275ed8c2a08fe555074144) Good catch!

PR #680, opened 2021-02-10
SQL Validation: Add stricter checking for the TRIGGER statement

pandah3:
## Description

Adding stricter SQL validation for the `TRIGGER` statement to avoid false positives.

Before, we were checking for:
`TRIGGER` (case sensitive)

This PR:
`CREATE TRIGGER` (case sensitive)
or
```
CREATE DEFINER=`root`@`host` TRIGGER`
```
(case sensitive)

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-validate-sql.js <sql file>`
1. Should only catch instances stated above

Tried testing w/:

```
CREATE TRIGGER
CREATE DEFINER=`root`@`%` TRIGGER
CREATE DEFINER=`root`@`host` TRIGGER
CREATE DEFINER=`root`@`127.0.0.1` TRIGGER
CREATE DEFINER=`youruser`@`host` TRIGGER
```

Please let me know if there's any instances I missed! Thank you!

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -156,7 +156,8 @@ const checks: Checks = {
 		recommendation: 'Remove these lines',
 	},
 	trigger: {
-		matcher: /TRIGGER/,
+		// Match `CREATE (DEFINER=`root`@`host`) TRIGGER`
+		matcher: /^CREATE (DEFINER=`?(\w*)(`@`)?(\w*\.*%?)*`?)?(| )TRIGGER/i,
 		matchHandler: lineNumber => lineNumber,
 		outputFormatter: errorCheckFormatter,
 		results: [],
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -157,7 +157,7 @@ const checks: Checks = {
 	},
 	trigger: {
 		// Match `CREATE (DEFINER=`root`@`host`) TRIGGER`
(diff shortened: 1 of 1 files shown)


src/lib/validations/sql.js
```diff
@@ -156,7 +156,8 @@ const checks: Checks = {
 		recommendation: 'Remove these lines',
 	},
 	trigger: {
-		matcher: /TRIGGER/,
+		// Match `CREATE (DEFINER=`root`@`host`) TRIGGER`
+		matcher: /^CREATE (DEFINER=`?(\w*)(`@`)?(\w*\.*%?)*`?)?(| )TRIGGER/,
```
jblz: Now that it's more defined, we can probably make this case-insensitive.  Especially if we ship #681 alongside.

jblz: This doesn't seem to match the string:

```
CREATE (DEFINER=`root`@`host.fdsa.com`) TRIGGER
```

...unless I'm missing something.
https://regex101.com/r/9pCSWm/1

pandah3: Oops, I was missing the optional parentheses around the definer. Fixed the regex to be case-insensitive + match the example string 🙇‍♀️

pandah3:
This catches any `TRIGGER` SQL statements, but it doesn't catch any matching DB values if for some reason the statement was defined as a value (i.e:
```
INSERT INTO `table` VALUES (1, CREATE DEFINER=`root`@`host` TRIGGER`, 'exampleValue)
```

jblz (approved):
lgtm.  Thanks.

PR #681, opened 2021-02-10
SQL Import: Add ability to skip local validation

jblz:
## Description

Add a new option to the import sql command:
`--skip-validate`

When truthy or passed without a value, this option foregoes the line-by-line validation prior to uploading the file and submitting for import.

See discussion: https://a8c.slack.com/archives/CKUJZT08Y/p1612980186308600

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql` with and without `--skip-validate`
    * If not present or a falsey value is provided, file is validated as before
    * If present with an empty or truthy value (e.g. `--skip-validate=true`) file is not validated and is submitted as-is
   
Search & Replace options should still work as before regardless of the new param.

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -189,6 +189,10 @@ command( {
 	requireConfirm: 'Are you sure you want to import the contents of the provided SQL file?',
 } )
 	.command( 'status', 'Check the status of the current running import' )
+	.option(
+		'skip-validate',
+		'Do not perform pre-upload file validation. If unsupported entries are present, the import is likely to fail'
+	)
 	.option( 'search-replace', 'Perform Search and Replace on the specified SQL file' )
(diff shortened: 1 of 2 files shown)


elazzabi (approved):
I got the following error when I was trying this, so I fixed it. Also added some `chalk.yellow` to the `--skip-validate` flag. Looks good 🎉 

```
  Error: Step name validate is already completed.
    at ProgressTracker.setStatusForStepId (/Users/ahmedelazzabi/a8c/vip/dist/lib/cli/progress.js:145:13)
    at ProgressTracker.stepSuccess (/Users/ahmedelazzabi/a8c/vip/dist/lib/cli/progress.js:120:10)
    at command.option.option.option.option.examples.argv (/Users/ahmedelazzabi/a8c/vip/dist/bin/vip-import-sql.js:324:19)
```

PR #699, opened 2021-02-25
Subsite SQL Import into Launched Sites

### LINKED ISSUE #691: [SQL Import] Remove any limitations on importing to and retrieving the import status of sites marked “Launched”

### LINKED ISSUE #693: [SQL Import] Place abundant emphasis on the fact that the site is launched prior to kicking off an import
- Make clear that other sites that share resources could have their performance affected during an import (TODO: enumerate which cases this applies)

### LINKED ISSUE #695: [SQL Import] Show list of `blog_id`s that are being written to in the SQL file
* Collect as we validate the file
* Show to the customer

### LINKED ISSUE #696: [SQL Import] Show a “playbook” of what our validation interprets is going to happen in the import
* Tables created
* Number of INSERTS to x table, etc.
* Should add a disclaimer that this is approximate

markowsiak:
## Description

* get isVIP from the public API so we can gate feature availability to staff during development
* displays the tables that will be imported as part of the SQL import process
* Prompts the user once, and adds friction by requiring the domain of the import

image

#### Jest

`npm t` for automated tests.

#### Manual Test plan

1. Checkout this branch
1. Execute an import for an unlaunched non-multisite app `npm run build && ./dist/bin/vip-import-sql.js @{id}.develop ~/your-sql-dump.sql`
1. Note that the list of tables that will be imported are listed and that you must type the domain name of the app as confirmation.
2. Execute an import for a launched multisite app `npm run build && ./dist/bin/vip-import-sql.js @{id}.production ~/multisite-dump.sql`
3. Note that the list of tables that will be imported are listed and separated by blog_id and that confirmation to proceed requires you to type the domain name.
4. Logout of the CLI
5. Login as a non isVIP user
6. Retry the imported into a launched multisite and note that you cannot proceed 

Closes #691 
Closes #693 
Closes #695 
Closes #696 

---

### DIFF
--- a/__tests__/bin/vip-import-sql.js
+++ b/__tests__/bin/vip-import-sql.js
@@ -1,15 +1,11 @@
 /**
- * @format
- */
-
-/**
  * External dependencies
  */
 
 /**
  * Internal dependencies
  */
-import { validationsAndGetTableNames, gates } from 'bin/vip-import-sql';
+import { validateAndGetTableNames, gates } from 'bin/vip-import-sql';
 import * as isMultiSite from 'lib/validations/is-multi-site';
 import * as featureFlags from 'lib/api/feature-flags';
 import * as exit from 'lib/cli/exit';
@@ -22,7 +18,7 @@ const mockExit = jest.spyOn( process, 'exit' ).mockImplementation( () => {} );
 const exitWithErrorSpy = jest.spyOn( exit, 'withError' );
 
 describe( 'vip-import-sql', () => {
-	describe( 'validationsAndGetTableNames', () => {
+	describe( 'validateAndGetTableNames', () => {
 		it( 'returns an empty array when skipValidate is true', async () => {
 			const params = {
 				skipValidate: true,
@@ -30,7 +26,7 @@ describe( 'vip-import-sql', () => {
 				envId: 1,
 				fileNameToUpload: '__fixtures__/client-file-uploader/db-dump-ipsum-67mb.sql',
 			};
-			const result = await validationsAndGetTableNames( params );
+			const result = await validateAndGetTableNames( params );
 			expect( result ).toEqual( [] );
 		} );
 		it( 'returns an array of table names that are contained within the input file', async () => {
@@ -40,7 +36,7 @@ describe( 'vip-import-sql', () => {
 				envId: 1,
 				fileNameToUpload: '__fixtures__/client-file-uploader/db-dump-ipsum-67mb.sql',
 			};
-			const result = await validationsAndGetTableNames( params );
+			const result = await validateAndGetTableNames( params );
 			const expected = [
 				'wp_commentmeta',
 				'wp_comments',
--- a/__tests__/bin/vip-import-sql.js
+++ b/__tests__/bin/vip-import-sql.js
@@ -1,0 +1,89 @@
+/**
+ * @format
+ */
+
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import { validationsAndGetTableNames, gates } from 'bin/vip-import-sql';
+import * as isMultiSite from 'lib/validations/is-multi-site';
+import * as featureFlags from 'lib/api/feature-flags';
+import * as exit from 'lib/cli/exit';
+
+jest.mock( 'lib/tracker' );
+jest.mock( 'lib/validations/site-type' );
+jest.mock( 'lib/validations/is-multi-site' );
+jest.mock( 'lib/api/feature-flags' );
+const mockExit = jest.spyOn( process, 'exit' ).mockImplementation( () => {} );
+const exitWithErrorSpy = jest.spyOn( exit, 'withError' );
+
+describe( 'vip-import-sql', () => {
+	describe( 'validationsAndGetTableNames', () => {
+		it( 'returns an empty array when skipValidate is true', async () => {
+			const params = {
+				skipValidate: true,
+				appId: 1,
+				envId: 1,
+				fileNameToUpload: '__fixtures__/client-file-uploader/db-dump-ipsum-67mb.sql',
+			};
+			const result = await validationsAndGetTableNames( params );
+			expect( result ).toEqual( [] );
+		} );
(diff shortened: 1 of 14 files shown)


src/lib/validations/site-type.js
```diff
@@ -28,25 +28,59 @@ export const siteTypeValidations = {
...
+				);
+			}
 
-			throw new Error( 'You have provided a multisite SQL dump file for import into a single site (non-multisite).' );
+			if ( ! isMultiSiteSqlDump ) {
+				await track( 'import_sql_command_error', {
+					error_type: 'subsite-import-without-subsite-sql-dump',
+				} );
+				throw new Error(
+					'You have requested a subsite SQL import but have no provided a subsite compatiable SQL dump.'
```
jblz:
```suggestion
					'You have requested a subsite SQL import but have not provided a subsite compatiable SQL dump.'
```

mjangda:
> Prompts the user once, and adds friction by requiring the name of the environment

A thought: It might be better to prompt the user for the domain / placeholder name (`example-com.go-vip.net` or `example-com.production`) rather than just the environment name (`production`).

We want to prevent someone from accidentally importing into the wrong application as well (e.g. agency partner types `vip import sql 124` instead of `123`).

elazzabi (commented):
Code looks good 🎉 I'll test with an app next. Do you have an app and SQL dump you use for testing?

__tests__/bin/vip-import-sql.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * @format
+ */
+
```
jblz:
```suggestion
```

We can ditch the pragma now -- it's not required.

src/bin/vip-import-sql.js
```diff
@@ -54,6 +65,21 @@ const appQuery = `
...
+const GET_WP_SITE_LIST = gql`
+	query getWpSites($appId: Int) {
+		app(id: $appId) {
+			environments {
+				wpSites {
+					nodes {
+						homeUrl
+						id
+					}
+				}
```
jblz: Curious why this isn't just part of the app query...

Does it error out if the site isn't multisite or something?

src/bin/vip-import-sql.js
```diff
@@ -180,13 +212,187 @@ const examples = [
...
+	let siteArray = [];
+	if ( isMultiSite ) {
+		// eslint-disable-next-line no-multi-spaces
+		console.log( `  multisite: ${ isMultiSite.toString() }` );
+		siteArray = await getMultiSiteList( { env, track } );
+	}
+
+	if ( ! tableNames.length ) {
+		console.log();
+		console.log( 'Since validation was skipped, no playbook information could be displayed' );
```
jblz: I'd probably just not print anything in this case.  Maybe change these to `debug`s?

src/bin/vip-import-sql.js
```diff
@@ -180,13 +212,187 @@ const examples = [
...
+};
+
+export type validationsAndGetTableNamesInputType = {
+	skipValidate: boolean,
+	appId: number,
+	envId: number,
+	fileNameToUpload: string,
+};
+
+export async function validationsAndGetTableNames( {
```
jblz: Super minor, but this reads odd to me.  I'd suggest either:

```suggestion
export async function validateAndGetTableNames( {
```

or something like `applyValidationsAndGetTableNames`

src/lib/validations/is-multi-site.js
```diff
@@ -15,27 +15,37 @@ import API from 'lib/api';
 import { trackEventWithEnv } from 'lib/tracker';
 import * as exit from 'lib/cli/exit';
 
+let isMultiSite;
```
jblz: This will be invalid if `isMultiSiteInSiteMeta` is called multiple times with different input.  If we're going to cache the results, we should do it in a map of the input.

jblz (commented):
This is looking good... left a few code comments and I'm giving it a run through now.

src/bin/vip-import-sql.js
markowsiak: tbh I didn't consider adding it to the app query.
I'll test with moving it there to see the result.

markowsiak: Making that change worked out nicely 👍

src/bin/vip-import-sql.js
```diff
@@ -192,13 +214,154 @@ const examples = [
...
+};
+
+export async function validateAndGetTableNames( {
+	skipValidate,
+	appId,
+	envId,
+	fileNameToUpload,
+}: validateAndGetTableNamesInputType ): Promise<Array<string>> {
+	const validations = [ staticSqlValidations, siteTypeValidations ];
+	let tableNamesInSqlFile = [];
```
jblz: Super minor, but we're not really reassigning this, so I'd drop the variable and just return the appropriate values below:

```suggestion
```

src/bin/vip-import-sql.js
```diff
@@ -192,13 +214,154 @@ const examples = [
...
+
+If you are confident the file does not contain unsupported statements, you can retry the command with the ${ chalk.yellow(
+		'--skip-validate'
+	) } option.
+` );
+		}
+		// this can only be called after static validation of the SQL file
+		tableNamesInSqlFile = getTableNames();
+	}
+	return tableNamesInSqlFile;
```
jblz: ... & we can drop an indention level because the `return` obviates the `else` case:

```suggestion
	if ( skipValidate ) {
		console.log( 'Skipping SQL file validation.' );
		return [];
	}
	try {
		await fileLineValidations( appId, envId, fileNameToUpload, validations );
	} catch ( validateErr ) {
		console.log( '' );
		exit.withError( `${ validateErr.message }
If you are confident the file does not contain unsupported statements, you can retry the command with the ${ chalk.yellow( '--skip-validate' ) } option.
` );
	}
	// this can only be called after static validation of the SQL file
	return getTableNames();
```

jblz (dismissed):
This is looking great to me.  I left some minor code style preferences, but feel free to ignore them if you'd like.

jblz (approved):

PR #706, opened 2021-03-15
Add/local dev env

pschoffer:
## Description

Adds the first step towards integrating local dev environment into `vip` command.

This adds `vip dev-environent start` command. Which will create (if non-existing) and start the environment.

It is a slightly reduced version of our internal tool. Things I dropped to make it easier for the first step (we can add them later):
- lookup of config by site id
- auto cloning of clients code (client can still clone their code manually and point to the path of that)  
- more operations (stop/destroy) - TBA in follow-up work

Maybe the command should be first hidden behind a isVIP flag for the first iteration untill we get support for all operations. For that we could use https://github.com/Automattic/vip/pull/699 after merged. 

## Steps to Test

Try variants of command. Use `--help` to learn about a possible parameter. Please let me know if help doesn't make sense or is incomplete in some sense.

It's usefull to have `export DEBUG="@automattic/vip:bin:dev-environment,@automattic/vip:bin:vip"` to see more info like instance path so you can remove that folder to start fresh.


rinatkhaziev:
Should we consider shortening this to `dev-env` or even just `dev`?

assets/dev-environment.lando.template.yml.ejs
```diff
@@ -0,0 +1,251 @@
+name: vipdev<%= siteSlug %>
```
pschoffer, line 1: Not sure about where to put this file... I created an `assets` folder, but I am happy to move it around.

src/bin/vip.js
```diff
@@ -48,8 +49,12 @@ const rootCmd = async function() {
 			.command( 'import', 'Check the validity of an import source' )
 			.command( 'search-replace', 'Perform Search and Replace tasks on files' )
 			.command( 'sync', 'Sync production to a development environment' )
-			.command( 'wp', 'Run WP CLI commands against an environment' )
-			.argv( process.argv );
+			.command( 'wp', 'Run WP CLI commands against an environment' );
+
+		// TODO hide behind flag
```
pschoffer: TODO here

src/lib/dev-environment.js
```diff
@@ -0,0 +1,188 @@
...
+ * External dependencies
+ */
+import debugLib from 'debug';
+import xdgBasedir from 'xdg-basedir';
+import os from 'os';
+import fs from 'fs';
+import ejs from 'ejs';
+import path from 'path';
+import Lando from 'lando/lib/lando';
+import landoUtils from 'lando/plugins/lando-core/lib/utils';
```
pschoffer, line 16: I don't know if these paths will keep working once it gets packaged and installed ??

src/lib/dev-environment.js
```diff
@@ -0,0 +1,188 @@
...
+		debug( 'Instance data to create a new environment:', instanceData );
+
+		await prepareLandoEnv( instanceData, instancePath );
+	}
+
+	await landoStart( instancePath );
+}
+
+function getLandoConfig() {
+	const landoPath = path.join( __dirname, '..', '..', 'node_modules', 'lando' );
```
pschoffer, line 103: Also this path might break when the command gets packaged and installed ??

src/lib/dev-environment.js
```diff
@@ -0,0 +1,188 @@
...
+
+[ 'wordpress', 'muPlugins', 'clientCode' ].forEach( type => {
+	defaults[ type ] = {
+		mode: 'image',
+		image: containerImages[ type ].image,
+		tag: containerImages[ type ].tag,
+	};
+} );
+
+const landoFileTemplatePath = 'assets/dev-environment.lando.template.yml.ejs';
```
pschoffer, line 63: This path is also kinda relative... so not sure if it will work.

pschoffer (commented):
Some open question to more experienced npm package peops.

pschoffer:
@rinatkhaziev I am thinking it is OK to be more verbose especially as this is end customer-facing. I would be OK to shortening it to `dev-env` (although I think `dev-environment` is better as that leads to less confusion about what the env means :) ).

assets/dev-environment.lando.template.yml.ejs
netsuso: Fair enough... Maybe it could be `dev-env` if we need to add more files in the future, but for the moment I think it's OK

src/lib/dev-environment.js
```diff
@@ -0,0 +1,188 @@
...
+async function landoStart( instancePath ) {
+	debug( 'Will start lando app on path:', instancePath );
+
+	const lando = new Lando( getLandoConfig() );
+	await lando.bootstrap();
+
+	const app = lando.getApp( instancePath );
+	await app.init();
+
+	await app.start();
```
netsuso, line 131: It's so great that this works as a library! And we don't get the fancy messages, so it's an easy white labeling 😄

netsuso (changes requested):
In general this looks great! I'm especially amazed at how nice the Lando library integration works. This will make the VIP dev environment a much more integrated and clean solution. Good job!

Although for the moment we only have the `start` command, I think it's a good moment to think about what we want. In my first implementation I only had a `create` command that also started the environment. I did this because I wanted the script to provide a running dev environment without ever requiring to run `lando` (although you needed `lando` to destroy it, but well...).

Now we have full control of the commands we want to provide, and I think it would be great if we split it and provide two commands:

- create: it receives all parameters to configure the environment, but it doesn't start it
- start: simply starts an existing environment. So you can start it, then stop it, and start it again, without needing to recreate it.

So I can envision two sets of commands:

- create, list, status, destroy, update: manage how environments are created, destroyed, updated (i.e, if new software versions are available), etc.

- start, stop, restart: these are just operations on the existing environments. You simply provide the slug, so you can have `vip dev-environment start <slug>`. The command `list` above would show all your environments and their status (running, stopped), so you can easily operate on them.

This doesn't mean it has to be done for this MVP, but this way we can make progress according to that general plan. What do you think?

src/lib/dev-environment.js
rinatkhaziev: Also probably need to bump phpVersion to 7.4 right away.

src/lib/dev-environment.js
```diff
@@ -0,0 +1,188 @@
...
+		image: 'wpvipdev/skeleton',
+		tag: '181a17d9aedf7da73730d65ccef3d8dbf172a5c5',
+	},
+};
+
+export const defaults = {
+	environmentSlug: 'vip-local',
+	title: 'VIP Dev',
+	multisite: false,
+	phpVersion: '7.3',
```
rinatkhaziev:
```suggestion
	phpVersion: '7.4',
```

pschoffer:
Thanks for the feedback - @netsuso ! 

In my mind, most of the end users won't really work with multiple dev-environments (I know I rarely am, and that is only because I have one multisite and one single site). Which makes commands related to the management of environments kinda less important (like list, stat) imho. So you know I would wait for a feedback before building those up.

```
Now we have full control of the commands we want to provide, and I think it would be great if we split it and provide two commands:

create: it receives all parameters to configure the environment, but it doesn't start it
start: simply starts an existing environment. So you can start it, then stop it, and start it again, without needing to recreate it.
```
The current `start` command does both create and start. If you have already an existing environment it will just work as `start` in your example. If you don't it will create it and then start as our sandboxes do. I kinda like this simplicity. What drawbacks do you see to this approach. On the other hand, I don't have a too strong opinion on this, so I am fine with splitting it as you suggest.

The one thing I would push against is making the `slug` positional and therefore required argument in the dev-env operation. Again I am kinda expecting that in most cases the user will only really have one environment.

netsuso:
> In my mind, most of the end users won't really work with multiple dev-environments

I'm not sure about this... There are quite a few customers that run multiple sites, and I bet they will want to have multiple dev environments for them. But maybe @parkcityj has a more informed opinion on this :)

netsuso:
> > In my mind, most of the end users won't really work with multiple dev-environments
> 
> I'm not sure about this... There are quite a few customers that run multiple sites, and I bet they will want to have multiple dev environments for them. But maybe @parkcityj has a more informed opinion on this :)

Actually, we could even consider that the slug becomes the siteid+environment, which is the usual way of handling the `vip` tool. In that case. For the most cases, we want to allow customers to simply "create an local dev env for this siteId/environment". The only problem is that we wouldn't allow multiple dev envs for a single site, but that might be not so important...

pschoffer:
> > > In my mind, most of the end users won't really work with multiple dev-environments
> > 
> > 
> > I'm not sure about this... There are quite a few customers that run multiple sites, and I bet they will want to have multiple dev environments for them. But maybe @parkcityj has a more informed opinion on this :)
> 
> Actually, we could even consider that the slug becomes the siteid+environment, which is the usual way of handling the `vip` tool. In that case. For the most cases, we want to allow customers to simply "create an local dev env for this siteId/environment". The only problem is that we wouldn't allow multiple dev envs for a single site, but that might be not so important...

I like that idea. We can always leave the option to put a custom name for the environment (and therefore have more of them), as it is now. We would just change the default name to be more dynamic (if they provide site id that is).

rinatkhaziev:
There definitely will be customers using this for many sites (think partner agencies' developers or the likes of NCA which have dozens of sites)

> The only problem is that we wouldn't allow multiple dev envs for a single site, but that might be not so important...

Agree on this one.

parkcityj:
Yes, I think a lot of customers will spin up multiple dev environments for different applications - likely including their nonproduction apps.   The way the CLI currently works is vip @appname.environment - so that should still work fine, as long as running multiple apps is supported.

pschoffer:
I split the `start` to `create` and `start` commands. I also added some more usefull error messages (e.g. when you try to start environment that doesnt exist or create one that already exists).

netsuso:
> I split the `start` to `create` and `start` commands. I also added some more usefull error messages (e.g. when you try to start environment that doesnt exist or create one that already exists).

I think we can go ahead with the current branch, to avoid making it much bigger, so we can merge it and keep working on it. E.g, I'd suggest adding `stop` and `destroy`/`remove` commands, as they should be easy and give a lot of value.

netsuso (dismissed):
:ship:

src/bin/vip-dev-environment-create.js
```diff
@@ -0,0 +1,68 @@
...
+		try {
+			await createEnvironment( slug, opt );
+
+			const message = chalk.green( '✓' ) + ` environemnt created.\n\n\nTo start it please run:\n\n${ startCommand }\n`;
+			console.log( message );
+		} catch ( e ) {
+			let messageToShow = chalk.red( 'Error:' );
+			if ( 'Environment already exists.' === e.message ) {
+				messageToShow += `Environemnt already exists\n\n\nTo start the environment run:\n\n${ startCommand }\n\n` +
+				`To create another environemnt use ${ chalk.bold( '--slug' ) } option with a unique name.\n`;
```
nejclovrencic: Typo in `Environemnt` and `environemnt`

nejclovrencic:
Should we at some point add more commands, such as `list`, `stop` and `delete`? I think this would make this even more powerful, we could then maybe even remove the defaultSlug and replace it with a random string, similar as Docker does for containers.

netsuso:
> Should we at some point add more commands, such as `list`, `stop` and `delete`? I think this would make this even more powerful, we could then maybe even remove the defaultSlug and replace it with a random string, similar as Docker does for containers.

Yep, this is just the initial branch, but there will be more to come. You can also read [my comment](https://github.com/Automattic/vip/pull/706#pullrequestreview-619963964) about what you mention :)

src/bin/vip-dev-environment-create.js
pschoffer: Nice catch, It was on more places actually :nervous:

package.json
```diff
@@ -14,7 +14,9 @@
     "vip-import-validate-sql": "dist/bin/vip-import-validate-sql.js",
     "vip-search-replace": "dist/bin/vip-search-replace.js",
     "vip-sync": "dist/bin/vip-sync.js",
-    "vip-wp": "dist/bin/vip-wp.js"
+    "vip-wp": "dist/bin/vip-wp.js",
+    "vip-dev-environment": "dist/bin/vip-dev-environment.js",
+    "vip-dev-environment-start": "dist/bin/vip-dev-environment-start.js"
```
nejclovrencic: vip-dev-environment-create missing here

src/lib/constants/dev-environment.js
```diff
@@ -0,0 +1 @@
+export const DEV_ENVIRONMENT_COMMAND = 'vip dev-environmnet';
```
nejclovrencic: typo

nejclovrencic (approved):

PR #707, opened 2021-03-15
SQL Import: Apply a lower file size limit for launched envs

jblz:
## Description

Apply a different file size limit for SQL files for launched sites (to maintain site stability).

When a file is deemed too large for safely importing into a launched site (currently 350 MiB), show an error:

Screen Shot 2021-03-16 at 12 48 30 PM

Unlaunched sites will continue to be subject to the previous file limit and will show the same error (minus the bit about launched sites).

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js` with a known good import file smaller than the new limit
1. Verify it is not incorrectly gated by this change
1. Copy the known good import file to a temporary location
1. Expand the file size to be greater than the limit (`truncate -s 400M /tmp/your/test/file.sql`)
1. Run `./dist/bin/vip-import-sql.js` with the expanded file against a **launched** test site
1. Verify it **is blocked** by this change and an appropriate error shown
1. Run `./dist/bin/vip-import-sql.js` with the expanded file against an **unlaunched** test site
1. Verify it **is not blocked** by this change
1. Expand the file size to be greater than the unlaunched limit (`truncate -s 11000M /tmp/your/test/file.sql`)
1. Run `./dist/bin/vip-import-sql.js` with the expanded file against an **unlaunched** test site
1. Verify it **continues to be blocked** with this change in place

### DIFF
--- a/src/lib/client-file-uploader.js
+++ b/src/lib/client-file-uploader.js
@@ -22,10 +22,6 @@ import debugLib from 'debug';
  */
 import API from 'lib/api';
 import { MB_IN_BYTES } from 'lib/constants/file-size';
-import {
-	SQL_IMPORT_FILE_SIZE_LIMIT,
-	SQL_IMPORT_FILE_SIZE_LIMIT_LAUNCHED,
-} from 'lib/site-import/db-file-import';
 
 const debug = debugLib( 'vip:lib/client-file-uploader' );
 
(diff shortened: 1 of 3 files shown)


src/lib/client-file-uploader.js
```diff
@@ -173,6 +177,17 @@ export async function uploadImportSqlFileToS3( {
 		debug( `** Compression resulted in a ${ calculation } smaller file 📦 **\n` );
 	}
 
+	const maxFileSize = env.launched
```
jblz: I'm ok if we choose to check this in the gates rather than here, but we should be aware it's not the same check we're performing server-side.

src/lib/site-import/db-file-import.js
```diff
@@ -6,9 +6,10 @@
 /**
  * Internal dependencies
  */
-import { GB_IN_BYTES } from 'lib/constants/file-size';
+import { GB_IN_BYTES, MB_IN_BYTES } from 'lib/constants/file-size';
 
 export const SQL_IMPORT_FILE_SIZE_LIMIT = 10 * GB_IN_BYTES;
+export const SQL_IMPORT_FILE_SIZE_LIMIT_LAUNCHED = 500 * MB_IN_BYTES;
```
markowsiak: Sounds like from chat that we want to lower this

markowsiak (dismissed):
LGTM other than the limit change, 🚢

jblz:
@markowsiak re-requested a review since this is materially different than before.

markowsiak (approved):
Change looks good and tests work out :shipit:

PR #709, opened 2021-03-25
PIE-2632: fix site-type validation for sql file

markowsiak:
## Description

Cleans up site-type validation where site type might mismatch from the SQL dump file provided.

## Steps to Test

Example:

1. Check out PR.
1. Execute an import into a single site with a multisite table SQL dump --> result is failure
2. Execute an import into a multisite with a single site SQL dump --> result is failure

---


src/lib/validations/site-type.js
```diff
@@ -42,45 +39,22 @@ export const siteTypeValidations = {
...
 
-			if ( ! isMultiSiteSqlDump ) {
-				await track( 'import_sql_command_error', {
-					error_type: 'subsite-import-without-subsite-sql-dump',
-				} );
-				throw new Error(
-					'You have requested a subsite SQL import but have not provided a subsite compatiable SQL dump.'
-				);
-			}
+		if ( isMultiSite && ! isMultiSiteSqlDump ) {
```
jblz, line 50: Is there a reason we don't want to allow importing a sql file that just has site 1 data and / or just top-level `wp_*` tables?  Seems like there could be a few use cases for this.

markowsiak: Actually, we need to beef up our validation here.  There are columns in the `wp_users` table for multisites that don't exist in single-site installations.  Thinking about it now, we might want to use that difference as the check for `isMultiSiteSqlDump` instead of if more than one site exists in the SQL.

nejclovrencic (approved):

PR #711, opened 2021-03-27
FORNO-556: Add media Import progress command

chrean:
## Description

Adds `vip media-import status` command, similar to `vip site-import progress`.

## Steps to Test

Make sure you have a running local copy of GOOP, Parker, MMS.

1. Check out PR.
1. Run `npm run build`
1. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media-status.js @1000.production`
1. Verify the command is showing the status of the running import - although no import is really running there.
2. To simulate the import progressing, try connecting to the DB and changing the status manually to all statues until `COMPLETED` and see how the commands behaves.
3. Also try increasing/changing the number of processed files and total files and see if the CLI reflects the changes.


src/bin/vip-import.js
```diff
@@ -18,3 +18,12 @@ command( )
 	.argv( process.argv, async () => {
 		await trackEvent( 'vip_import_command_execute' );
 	} );
+
+command( )
+	.command( 'media', 'Import media files to your application from a compressed archive' )
+	.command( 'media-progress', 'Import media files to your application from a compressed archive' )
+	.command( 'validate-files', 'Validate your media file library' )
```
saroshaga: So I think this will have to be regression tested and removed from the sql command above:
https://github.com/Automattic/vip/blob/9e1ebb3d0c2047f6382043a9a84bc0716bfd8c4c/src/bin/vip-import.js#L16-L16

saroshaga: Maybe even just `validate` ? Because it now lives under the `media` subcommand

src/bin/vip-import.js
```diff
@@ -18,3 +18,12 @@ command( )
 	.argv( process.argv, async () => {
 		await trackEvent( 'vip_import_command_execute' );
 	} );
+
+command( )
+	.command( 'media', 'Import media files to your application from a compressed archive' )
+	.command( 'media-progress', 'Import media files to your application from a compressed archive' )
```
saroshaga: I think we should follow the convention started by the SQL import here. Instead of having a separate command, let's have a nested command like so : https://github.com/Automattic/vip/blob/527a615b661c70ecdd8ba3f6122476767c2d9420/src/bin/vip-import-sql.js#L361-L361

Also, maybe just `status` , again to align with SQL imports.

```
vip import media status @mysite.develop
```

Same goes for the validate files command below.

src/bin/vip-import.js
saroshaga: Or for the sake of conserving time, let's remove it from here and leave it where it was before. We can punt this into a story for a future sprint maybe.

src/lib/media-import/status.js
```diff
@@ -0,0 +1,230 @@
...
+			id
+			name
+			type
+			repo
+			mediaImportStatus {
+				importId
+				siteId
+				status
+            	filesTotal
+            	filesProcessed
```
saroshaga:
```suggestion
				filesTotal
				filesProcessed
```

src/lib/media-import/status.js
```diff
@@ -0,0 +1,230 @@
...
+
+	if ( ! currentUserCanImportForApp( app ) ) {
+		throw new Error(
+			'The currently authenticated account does not have permission to view Media import status.'
+		);
+	}
+	let overallStatus = 'Checking...';
+
+	const setProgressTrackerSuffix = () => {
+		const sprite = getGlyphForStatus( overallStatus, progressTracker.runningSprite );
```
saroshaga, line 134: Maybe we should add appropriate glyphs for our Import statuses?

src/lib/media-import/progress.js
```diff
@@ -0,0 +1,78 @@
...
+	print( { clearAfter = false }: { clearAfter?: boolean } = {} ) {
+		if ( ! this.hasPrinted ) {
+			this.hasPrinted = true;
+			singleLogLine.clear();
+		}
+
+		const progressPercentage = Math.floor( this.status.filesProcessed / this.status.filesTotal * 100 );
+		const logs = `
+=============================================================
+Site ID: ${ this.status.siteId }
```
saroshaga:
```suggestion
App ID: 
Environment ID: ${ this.status.siteId }
```

Because Site ID is a GOOP specific term

src/lib/media-import/status.js
```diff
@@ -0,0 +1,230 @@
...
+			default:
+				statusMessage = `${ capitalize( overallStatus ) }`;
+		}
+
+		const maybeExitPrompt = `${ overallStatus === 'COMPLETING' ? exitPrompt : '' }`;
+
+		const suffix = `
+=============================================================
+Status: ${ statusMessage }
+Site: ${ app.name } (${ formatEnvironment( env.type ) })
```
saroshaga:
```suggestion
App: ${ app.name } (${ formatEnvironment( env.type ) })
```

src/lib/media-import/status.js
```diff
@@ -0,0 +1,230 @@
...
+	if ( ! environments?.length ) {
+		throw new Error( 'Unable to determine import status from environment' );
+	}
+	const [ environment ] = environments;
+	const { mediaImportStatus } = environment;
+
+	return mediaImportStatus;
+}
+
+function buildErrorMessage( importFailed ) {
```
saroshaga, line 102: This is the log on the FAILED status
```
Checking the Media import status for your environment...

Site ID: 1000
Current status: FAILED
Processed Files: 100/300 - 33%
Status: Checking...
Site: test1-go-vip (PRODUCTION)


Error: FAILED
An error occurred: 100/300 files imported.

If this error persists, please contact support.
```

1. I'm not sure if the direct response from Parker should be displayed. Maybe just leave a generic `Import #1 failed` for now?
2. The processed files percentage was already displayed , so maybe it is redundant
3. There are two status fields and one of them is `Checking...` . Also should we get rid of all fields that are already getting displayed?

src/bin/vip-import.js
chrean: I agree, I guess this is an oversight because that was the initial intention.

src/lib/media-import/progress.js
saroshaga: Ah I think, we can remove the Environment ID actually, just keep the App name . (Env will always be prod)

But like I mentioned elsewhere, this information is being also displayed later. Maybe we want to cut down on the duplication?

src/bin/vip-import.js
chrean: Removed via a4be731

src/lib/media-import/status.js
chrean: > 1. I'm not sure if the direct response from Parker should be displayed. Maybe just leave a generic Import #1 failed for now?

Probably not, although I think that errors returned from Parker should not contain any private or sensitive data, but should instead be of help to users to understand what the underlying problem is.

> 2. The processed files percentage was already displayed , so maybe it is redundant

Where it is already displayed? I think it appears only once, or? 🤔 

> 3. There are two status fields and one of them is Checking... . Also should we get rid of all fields that are already getting displayed?

Yeah, that may be confusing. `Current Status` was intended to display the current import state i.e. `INITIALIZING`, `COMPLETING`,  etc. I guess we could just rename it to something else.

chrean: For the last point, I have for now reworded to `Current step` although I don't like it because it's more of state... a79abdf
Perhaps I should just rename the other `Status`. 🤔

src/lib/media-import/status.js
chrean: Done via 34eec57.

src/lib/media-import/status.js
saroshaga: So, if you see my original post above, we have two fields with the same info (Site and Site ID) and _had_ two `Status`s...

chrean: Yup, got rid of that.
Now fixing the repeated output.

chrean:
> How about a ⚠️ for ABORTING and ABORTED ?

There may be terminals that do not support emojis, but perhaps that's a minor concern?

chrean:
However, changed it here: b38fd07

src/lib/media-import/status.js
chrean: Repeated output fixed.

src/lib/media-import/status.js
```diff
@@ -0,0 +1,238 @@
...
+				break;
+			case 'COMPLETING':
+				statusMessage = `Finishing up... ${ sprite } `;
+				break;
+			// Intentionally no break to get default case:
+			default:
+				statusMessage = `${ capitalize( overallStatus ) } ${ sprite }`;
+		}
+
+		const maybeExitPrompt = `${ overallStatus === 'COMPLETED' ? exitPrompt : '' }`;
```
saroshaga:
```suggestion
		const maybeExitPrompt = `${ overallStatus !== 'COMPLETED' ? exitPrompt : '' }`;
```

saroshaga:
@chrean , for non-prod environments, we do return a `null` response for the `mediaImportStatus`. This causes the CLI to 

```
→ VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media-status.js @1000.staging                                               (forno-556/add-progress-command|✚1)

  TypeError: Cannot destructure property 'status' of 'mediaImportStatus' as it is null.
    at checkStatus (/Users/saroshaga/Projects/Automattic/vip/dist/lib/media-import/status.js:200:9)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

Checking the Media import status for your environment...
Processed Files: undefined/undefined - NaN% 
Status: Checking... 
App: test1-go-vip (staging)

^C
```

Should we handle in the CLI itself? Saying that No Media Import was found ?

The reason Parker is returning `null` because the SQL Imports follow the same procedure - not entirely sure why though - but I went with it.

src/lib/media-import/status.js
saroshaga: Because we need the prompt for imports that are still ongoing.

saroshaga:
For the Failure status

```
→ VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media-status.js @1000                                                       (forno-556/add-progress-command|✚1)
✔ Which environment? · production

Checking the Media import status for your environment...
Processed Files: 290/300 - 96% ✕
Status: FAILED ✕
App: test1-go-vip (PRODUCTION)


Error: Import failed
An error occurred: 290/300 files imported.

If this error persists, please contact support.
```

Minor : I think we can do without the redundant `An error occurred: 290/300 files imported.` for now.

src/lib/media-import/status.js
```diff
@@ -0,0 +1,238 @@
...
+			};
+
+			// Kick off the check
+			checkStatus();
+		} );
+
+	try {
+		const results = await getResults();
+
+		if ( typeof results === 'string' ) {
```
saroshaga: I'm not quite sure how this matches the Media Import response 🤔 

Maybe we can convert this if/else null check the results (because they might always be returned) and construct an appropriate error response ?

src/lib/media-import/status.js
```diff
@@ -0,0 +1,238 @@
...
+
+	const setProgressTrackerSuffix = () => {
+		const sprite = getGlyphForStatus( overallStatus, progressTracker.runningSprite );
+
+		const exitPrompt = '(Press ^C to hide progress. The import will continue in the background.)';
+
+		let statusMessage;
+		switch ( overallStatus ) {
+			case 'COMPLETED':
+				statusMessage = `Success ${ sprite } imported data should be visible on your site ${ env.primaryDomain.name }.`;
```
saroshaga:
```suggestion
				statusMessage = `COMPLETED ${ sprite } : The Imported files should be visible on your site ${ env.primaryDomain.name }`;
```

src/lib/media-import/status.js
```diff
@@ -0,0 +1,238 @@
...
+				break;
+			case 'COMPLETING':
+				statusMessage = `Finishing up... ${ sprite } `;
```
saroshaga:
```suggestion
				statusMessage = `COMPLETING ${ sprite } `;
```

saroshaga (dismissed):
thanks for taking time to getting this ready @chrean , truly appreciate it!

I've left a couple of questions and a few minor suggestions (if you agree)

src/lib/media-import/status.js
```diff
@@ -0,0 +1,238 @@
...
+					setSuffixAndPrint();
+					/* TODO: Here we will add failure details */
+					return reject( { ...mediaImportStatus, error: 'Import failed' } );
+				}
+
+				progressTracker.setStatus( mediaImportStatus );
+
+				setSuffixAndPrint();
+
+				if ( [ 'COMPLETED', 'ABORTED' ] === status ) {
```
hanifn: This doesn't seem to be working. If the import is `COMPLETED`, the `checkStatus` loop still runs. Seems to work as expected if I changed it like so:
```suggestion
				if ( [ 'COMPLETED', 'ABORTED' ].includes( status ) ) {
```

hanifn (dismissed):

src/lib/media-import/status.js
chrean: Done in aefa2e2.
Also fixed a few other things and added `VALIDATED` and `VALIDATING` statuses to glyph considerations, considering those like `RUNNING` states.

src/lib/media-import/status.js
chrean: Thanks @hanifn , I had just noticed that and fixed that in a just-submitted commit 😓

chrean:
@saroshaga now the output on error looks a bit cleaner, as follows.
I have decided to keep the information on how many files were processed at it seemed a good place for a recap, even though that data is already present on the screen (but may require scrolling on smaller terminals if the output becomes longer in a future, not sure).

```
DEBUG="*media*" VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media-status.js  @1000.production

Checking the Media import status for your environment...
Processed Files: 100/300 - 33% ✕
Status: FAILED ✕
App: test1-go-vip (PRODUCTION)
(Press ^C to hide progress. The import will continue in the background.)

Error: Import failed
100/300 files imported.

If this error persists, please contact support.
```

chrean:
Now fixing this:

> Should we handle in the CLI itself? Saying that No Media Import was found ?

> The reason Parker is returning null because the SQL Imports follow the same procedure - not entirely sure why though - but I went with it.

chrean:
All feedback has been addressed. There are still things that may be improved compared to the original implementation of the ProgressTracker, which I already had to fix to handle cases where Parker response is `null`, etc.

saroshaga (approved):
I think the changes look good... 

🚀 🙇

PR #712, opened 2021-04-01
FORNO-555: Add start media import command

hanifn:
## Description

Adds a new `vip media-import` command to start a new media import

## Steps to Test

Make sure the docker containers for GOOP, Parker and MMS are running locally on your computer.

1. Check out PR.
1. Run `npm run build`
1. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js @1000.production --url https://beruberuberu-test.go-vip.co/wp-content/uploads/2021/03/wp-content.zip`
1. Verify the command is showing showing the correct environment and URL
1. Check MMS logs to confirm that files are being uploaded via the mock file service
1. Verify the command stops checking status after the import is `COMPLETED`


src/bin/vip-import-media.js
```diff
@@ -66,15 +92,66 @@ command( {
...
 	appQuery,
 	envContext: true,
-	requiredArgs: 1,
 	module: 'import-media',
-	requireConfirm: 'Are you sure you want to import the contents of the provided file?',
-	skipConfirmPrompt: true,
+	requireConfirm: 'Are you sure you want to import the contents of the url?',
 } )
-	.command( 'status', 'Check the status of the current running import' )
+	// .command( 'status', 'Check the status of the current running import' )
```
saroshaga: Should this be removed?

src/bin/vip-import-media.js
```diff
@@ -66,15 +92,66 @@ command( {
...
+		await track( 'import_media_start_execute' );
+
+		const progressTracker = new MediaImportProgressTracker( [] );
+		progressTracker.prefix = `
+=============================================================
+Processing the files import for your environment...
+`;
+
+		console.log();
+		console.log( `importing: ${ url }` );
```
saroshaga:
```suggestion
		console.log( `Importing archive from: ${ url }` );
```

src/bin/vip-import-media.js
```diff
@@ -66,15 +92,66 @@ command( {
...
+		console.log( `to: ${ env.primaryDomain.name }` );
```
saroshaga: I think maybe, this line and the one after it can be somehow combined? they offer the same information

saroshaga (approved):
Thanks for this @hanifn !! Left a few suggestions and tested this.

I was able to successfully upload a file:

```
→ VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js @70.production --url https://beruberuberu-test.go-vip.co/wp-content/uploads/wp-content.tar.gz
+ App: test-ms-subdir-01-vipv2-net (id: 70)
+ Environment: production (id: 70)
✔ Are you sure you want to import the contents of the url? (y/N) · true

importing: https://beruberuberu-test.go-vip.co/wp-content/uploads/wp-content.tar.gz
to: testingmapped.com
site: test-ms-subdir-01-vipv2-net (production)


Processing the files import for your environment...
Processed Files: 2/2 - 100% ✓
Status: COMPLETED ✓ : The Imported files should be visible on your site testingmapped.com
App: test-ms-subdir-01-vipv2-net (PRODUCTION)
```

and I also tested the failure condition:

```
→ VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js @1000.production --url https://beruberuberu-test.go-vip.co/wp-content/uploads/2021/03/wp-content.zip
+ App: test1-go-vip (id: 1000)
+ Environment: production (id: 1000)
✔ Are you sure you want to import the contents of the url? (y/N) · true

importing: https://beruberuberu-test.go-vip.co/wp-content/uploads/2021/03/wp-content.zip
to: testingmapped.com
site: test1-go-vip (production)


Processing the files import for your environment...

Status: FAILED ✕
App: test1-go-vip (PRODUCTION)
(Press ^C to hide progress. The import will continue in the background.)
```

_small thing but I noticed that when I use the URL that you provided, MMS is not reading the total file count properly 🤔 _

src/bin/vip-import-media.js
```diff
@@ -66,15 +92,66 @@ command( {
...
+			if ( e.graphQLErrors ) {
+				for ( const err of e.graphQLErrors ) {
+					console.log( chalk.red( 'Error:' ), err.message );
+				}
+
+				return;
+			}
+
+			await track( 'import_media_start_execute_error', {
+				error: `Error starting: ${ e.message }`,
```
saroshaga:
```suggestion
				error: `Error: ${ e.message }`,
```

saroshaga:
@hanifn , actually I was thinking, we need to add a disclaimer that if the directory structure inside the archive begins with `/wp-content/uploads` then we will extract all the files after that root path and import it. If the archive does not begin with `/wp-content/uploads` (and only that - we are not considering only `/uploads`) then we will just take the content as is.

https://github.com/Automattic/vip-go-media-management-service/blob/94e20929f19d05210fe14f105225078f62d1727f/src/uploads/uploads.service.ts#L248-L248

src/bin/vip-import-media.js
hanifn: Maybe just a `to: ${ env.primaryDomain.name } (${ env.type})`?
We are already showing the env name in the status checks so I'm thinking we can just remove that. What do you think?

src/bin/vip-import-media.js
hanifn: Removed in 6b3578c6e8ef023273486bbfdd60a8dab98000bd

hanifn:
>  we need to add a disclaimer

@saroshaga  Hmm.. Should we add this as part of the confirmation message? I'm thinking that this sort of disclaimer should be shown before the confirmation check

saroshaga:
> > we need to add a disclaimer
> 
> @saroshaga Hmm.. Should we add this as part of the confirmation message? 

Makes sense, thanks!

> I'm thinking that this sort of disclaimer should be shown before the confirmation check

Then it will have to be another confirmation check , wouldn't it?

hanifn:
> Then it will have to be another confirmation check , wouldn't it?

Yeah I've thought about that. I'm just not sure about how to do that in `args` 🤔

saroshaga:
> > Then it will have to be another confirmation check , wouldn't it?
> 
> Yeah I've thought about that. I'm just not sure about how to do that in `args` 🤔

I think having it as part of the same prompt would also be fine, we just to HIGHLIGHT it very well :D

hanifn:
> I think having it as part of the same prompt would also be fine, we just to HIGHLIGHT it very well :D

@saroshaga Added in f399935b1c5bfeebe551e6e36bfce13ee322e927
What do you think?

src/bin/vip-import-media.js
```diff
@@ -66,15 +92,68 @@ command( {
 	appContext: true,
 	appQuery,
 	envContext: true,
-	requiredArgs: 1,
 	module: 'import-media',
-	requireConfirm: 'Are you sure you want to import the contents of the provided file?',
-	skipConfirmPrompt: true,
+	requireConfirm: `
+${ chalk.bold( 'NOTE: If the provided archive\'s directory structure begins with `/wp-content/uploads`,' ) }
+${ chalk.bold( 'then we will extract only the files after that path and import it. Otherwise, we will' ) }
```
saroshaga:
```suggestion
${ chalk.red( chalk.bold( 'we will extract only the files after that path and import it. Otherwise, we will' ) ) }
```

saroshaga: @hanifn , looks good, left a suggestion. we can wrap all three lines in red chalk

hanifn: >  we can wrap all three lines in red chalk

Done in 09c0b37659ba06e90a9b2209e26cb6c1baf5edf9

src/bin/vip-import-media.js
saroshaga: thanks! looks good!

saroshaga:
🚢

src/bin/vip-import-media.js
```diff
@@ -66,15 +92,68 @@ command( {
...
+	requireConfirm: `
+${ chalk.red.bold( 'NOTE: If the provided archive\'s directory structure begins with `/wp-content/uploads`,' ) }
+${ chalk.red.bold( 'we will extract only the files after that path and import it. Otherwise, we will' ) }
+${ chalk.red.bold( 'import all files and preserve the directory structure as is.' ) }
+
+Are you sure you want to import the contents of the url?
+`,
 } )
-	.command( 'status', 'Check the status of the current running import' )
+	.option( 'url', 'Valid URL to download a file archive from', '' )
```
saroshaga, line 104: @hanifn , actually I think this was incorrect. We have essentially removed the `import media status` command here

saroshaga: Sorry for not detecting it earlier!

saroshaga: Fixing here: https://github.com/Automattic/vip/pull/715

PR #714, opened 2021-04-01
destroy environment

pschoffer:
## Description

Adds `vip dev-environment destroy` subcommand.

## Steps to Test
Run `vip dev-environment destroy`

### DIFF
--- a/src/lib/dev-environment.js
+++ b/src/lib/dev-environment.js
@@ -115,6 +115,22 @@ export async function createEnvironment( slug, options ) {
 	await prepareLandoEnv( instanceData, instancePath );
 }
 
+export async function destroyEnvironment( slug ) {
+	debug( 'Will destroy an environment', slug );
+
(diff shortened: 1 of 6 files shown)


src/lib/dev-environment.js
```diff
@@ -99,6 +99,22 @@ export async function createEnvironment( slug, options ) {
...
+
+	debug( 'Instance path for', slug, 'is:', instancePath );
+
+	const environmentExists = fs.existsSync( instancePath );
+
+	if ( ! environmentExists ) {
+		throw new Error( 'Environment not found.' );
+	}
+
+	fs.rmdirSync( instancePath, { recursive: true } );
```
nejclovrencic, line 132: I think we should probably run `lando destroy` here as well, to remove all the docker stuff that gets created by lando

pschoffer: Great point! I added destroy.

nejclovrencic (approved):

PR #715, opened 2021-04-02
FORNO-555: Fix `import media status` sub command

saroshaga:
## Description

This PR re-adds the `import media status` sub-command. It converts the `--url` option for the `import media` command to a required argument.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js -h`
1. Should show the subcommand

```
  Usage: vip import-media.js [options] [command]
  
  Commands:
    status  Check the status of the current running import
```
5. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js`
6. Should show the help menu again
7. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js @70.production  https://beruberuberu-test.go-vip.co/wp-content/uploads/wp-content.tar.gz`

```
+ App: test-ms-subdir-01-vipv2-net (id: 70)
+ Environment: production (id: 70)
+ Archive URL: https://beruberuberu-test.go-vip.co/wp-content/uploads/wp-content.tar.gz
✔ 
NOTE: If the provided archive's directory structure begins with `/wp-content/uploads`,
we will extract only the files after that path and import it. Otherwise, we will
import all files and preserve the directory structure as is.

Are you sure you want to import the contents of the url?
 (y/N) · false
```
8. If you provide a local file the following error should be displayed
```
Error: 
	Invalid URL provided: ${ url }
	Please make sure that it is a publicly accessible web URL containing an archive of the media files to import
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,8 @@
     "vip-import": "dist/bin/vip-import.js",
     "vip-import-sql": "dist/bin/vip-import-sql.js",
     "vip-import-sql-status": "dist/bin/vip-import-sql-status.js",
+    "vip-import-media": "dist/bin/vip-import-media.js",
+    "vip-import-media-status": "dist/bin/vip-import-media-status.js",
     "vip-import-validate-files": "dist/bin/vip-import-validate-files.js",
     "vip-import-validate-sql": "dist/bin/vip-import-validate-sql.js",
     "vip-search-replace": "dist/bin/vip-search-replace.js",
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -82,9 +82,10 @@ ${ chalk.red.bold( 'import all files and preserve the directory structure as is.
 
 Are you sure you want to import the contents of the url?
 `,
+	skipConfirmPrompt: true,
 } )
-	.option( 'url', 'Valid URL to download a file archive from', '' )
 	.command( 'status', 'Check the status of the current running import' )
+	.option( 'url', 'Valid URL to download a file archive from', '' )
 	.examples( examples )
 	.argv( process.argv, async ( args: string[], opts ) => {
 		const { app, env, url } = opts;
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -46,8 +46,6 @@ const START_IMPORT_MUTATION = gql`
 				importId
 				siteId
 				status
-				filesTotal
-				filesProcessed
 			}
 		}
 	}
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -17,22 +17,11 @@ import chalk from 'chalk';
  */
 import command from 'lib/cli/command';
 import API from 'lib/api';
-import {
-	currentUserCanImportForApp,
-	isSupportedApp,
-	SQL_IMPORT_FILE_SIZE_LIMIT,
-	SQL_IMPORT_FILE_SIZE_LIMIT_LAUNCHED,
-} from 'lib/site-import/db-file-import';
 // eslint-disable-next-line no-duplicate-imports
 import { trackEventWithEnv } from 'lib/tracker';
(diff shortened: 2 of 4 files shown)


hanifn:
@saroshaga I think this is breaking the import start command

```
❯ VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip-import-media.js @1000.production --url https://beruberuberu-test.go-vip.co/wp-content/uploads/2021/03/wp-content.zip
  Usage: vip import-media.js [options] [command]
  
  Commands:
    status  Check the status of the current running import
  
  Options:
    -a, --app      Specify the app
    -e, --env      Specify the environment
    -f, --force    Skip confirmation (disabled by default)
    -h, --help     Output the help for the (sub)command
    -u, --url      Valid URL to download a file archive from (defaults to "")
  
  Examples:
    - Start a media import with the contents of the archive file in the URL
    $ vip import media @mysite.develop --url https://domain.to/archive.zip

    - Check the status of the most recent import. If an import is running, this will poll until it is complete.
    $ vip import media status @mysite.develop
```

saroshaga:
> @saroshaga I think this is breaking the import start command

🤔 I'm not sure why... 

Without the subcommand though, it won't show up on the actual CLI, right?

saroshaga:
> > @saroshaga I think this is breaking the import start command
> 
> 🤔 I'm not sure why...
> 
> Without the subcommand though, it won't show up on the actual CLI, right?

@hanifn , the problem lies here

https://github.com/Automattic/vip/blob/68dc0f60fb44e1e3bc08e2823346467a55bf8613/src/lib/cli/command.js#L95-L95

The custom command expects at least one argument. I'm going to by-pass this for our case.

Fixed here:  a40eb5c

hanifn:
> The custom command expects at least one argument. I'm going to by-pass this for our case.

I wonder if there' s a proper way to do this? 🤔

saroshaga:
> > The custom command expects at least one argument. I'm going to by-pass this for our case.
> 
> I wonder if there' s a proper way to do this? 🤔

https://a8c.slack.com/archives/CKUJZT08Y/p1617604020215400

pandah3 (commented):
Are we specifying which site's media import status we're checking on? That should be the args, ie:

`vip import status @1000` or
`vip import status @1000.production`

saroshaga:
> Are we specifying which site's media import status we're checking on? That should be the args, ie:
> 
> `vip import status @1000` or
> `vip import status @1000.production`

hey @pandah3 !! The command that we're looking for is `vip import media @1000.production` - which starts a media import.

The problem is that if we add the `status` sub command, then the aforementioned start-import command errors out

pandah3:
Hmm I don't think we have a conditional for the `import sql status` subcommand so this should work without having to put in any special rules.. I tried digging through the code a bit, but Docker isn't cooperating with me, so I'm unable to test this out 💢

saroshaga:
> Hmm I don't think we have a conditional for the `import sql status` subcommand so this should work without having to put in any special rules.. I tried digging through the code a bit, but Docker isn't cooperating with me, so I'm unable to test this out 💢

The sql and media import commands are a little different. There are no arguments required for the media import command. And by the looks of it, [this code](https://github.com/Automattic/vip/blob/68dc0f60fb44e1e3bc08e2823346467a55bf8613/src/lib/cli/command.js#L94-L94) does not allow a command being called without one.

saroshaga:
@chrean , @hanifn : as per our discussion, I have converted the option to a required argument for the `import media` command. This has fixed the original problem as well. A few cosmetic improvements have also been made.

@pandah3 : we decided to get rid of the option that was causing the problem and decided to convert it into an argument. Since there is one argument for the command, the aforementioned check is passing now.

chrean (approved):
👍

PR #718, opened 2021-04-07
Dev-environment adds info

pschoffer:
## Description

Adds dev-environment info command

## Steps to Test

run `vip dev-environment info`

### DIFF
--- a/src/bin/vip-dev-environment-info.js
+++ b/src/bin/vip-dev-environment-info.js
@@ -1,0 +1,53 @@
+#!/usr/bin/env node
+
+/**
+ * @flow
+ * @fomat
+ */
+
+/**
+ * External dependencies
+ */
+import debugLib from 'debug';
+import chalk from 'chalk';
+
+/**
+ * Internal dependencies
+ */
+import command from 'lib/cli/command';
+import { defaults, printEnvironmentInfo } from 'lib/dev-environment';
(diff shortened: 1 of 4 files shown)


src/bin/vip-dev-environment-info.js
```diff
@@ -0,0 +1,53 @@
...
+	.examples( examples )
+	.argv( process.argv, async ( arg, opt ) => {
+		const slug = opt.slug || defaults.environmentSlug;
+
+		debug( 'Args: ', arg, 'Options: ', opt );
+
+		try {
+			await printEnvironmentInfo( slug );
+		} catch ( e ) {
+			let messageToShow = chalk.red( 'Error:' );
```
nejclovrencic, line 42: This seems to be repeating for every command now, maybe it would make sense to move this error handling part to a common function?

nejclovrencic (dismissed):

src/bin/vip-dev-environment-info.js
pschoffer: Yeah that's a good point but then I check and the error message I return on stop fail is different by design (no prompt to create environment). So I figured a little bit of copy and paste never heard nobody:) (I was not able to figure out a way to do it in a nice way)

nejclovrencic (approved):

PR #726, opened 2021-04-19
Add/dev env wp

pschoffer:
## Description

Adding the abilty to run wp cli commands from dev-environment through the vip command

## Steps to Test

Try `vip dev-environment wp -- post list`


nejclovrencic:
@pschoffer I think we're missing the command in package.json `bin`.

Can we also somehow display help when you don't provide `--`? Right now this produces an error:

```
> vip dev-environment wp -- post list
Error: '/Users/nejclovrencic/.nvm/versions/node/v14.16.0/bin/node' is not a registered wp command. See 'wp help' for available commands.

  Error: Error: '/Users/nejclovrencic/.nvm/versions/node/v14.16.0/bin/node' is not a registered wp command. See 'wp help' for available commands.
  ```

pschoffer:
Nice catch. 

I also noticed that if you do `vip dev-environment wp post` the "post" is ignored as it is before "--" so I tried to add a validation for that and return an error instead.

nejclovrencic (approved):

rinatkhaziev:
Just noting the following  also works: `vip dev-environment --slug es-upgrade -- wp post list` 🎉

PR #730, opened 2021-04-21
Dev-environment Support @naming

pschoffer:
## Description

This adds support for `@<siteId>.<envName>` notation for dev-environment subcommands. For now, we only infer the name of the environment. The next step is to use provided data to fetch some information on creation to init some defaults correctly.

## Steps to Test

1. `vip @2891.develop dev-environment create`

### DIFF
--- a/src/lib/dev-environment.js
+++ b/src/lib/dev-environment.js
@@ -20,7 +20,7 @@ import chalk from 'chalk';
 /**
  * Internal dependencies
  */
-import { DEV_ENVIRONMENT_COMMAND } from './constants/dev-environment';
+import { DEV_ENVIRONMENT_FULL_COMMAND } from './constants/dev-environment';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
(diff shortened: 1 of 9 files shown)


src/lib/dev-environment.js
```diff
@@ -401,15 +401,34 @@ export function getEnvironmentPath( name: string ) {
...
 }
+
+type EnvironmentNameOptions = {
+	slug: string,
+	app: string,
+	env: string,
+}
+
+export function getEnvironmentName( options: EnvironmentNameOptions ) {
+	if ( options.slug ) {
```
nejclovrencic, line 423: Just a thought, does slug have priority if I provide both @app.env and --slug?

nejclovrencic (approved):

src/lib/dev-environment.js
pschoffer: It does, right? I have test case for it as well. The idea being say you would like to use an alias or have multiple dev-envs per single VIP environment.

PR #731, opened 2021-04-22
Add custom WP config for dev environment

pauarge:
## Description

When spinning up a new dev environment from the CLI, lando would use the default WP configuration that would come with the image. That was a bit problematic, because if we needed to do some tweaks on that configuration, we could not do that.

This was the case with the `WPCOM_IS_VIP_ENV` variable, which should be false as this is a local environment, but was true by default.

This PR adds the `wp-config-defaults.php` file, which can be customized further to fit a local development environment. In this case, the only change from the default behavior is `WPCOM_IS_VIP_ENV`.

## Steps to Test

1. Check out PR
2. Run `npm link` to install the latest version
3. Spin up a dev environment and create/edit a post (doing that before wouldn't work due to the incorrect value of `WPCOM_IS_VIP_ENV`)


.gitignore
```diff
@@ -17,4 +17,8 @@ node_modules/
 # jest coverage
 coverage
 
+# Mac
 .DS_Store
+
+# Webstorm
+.idea
```
pauarge: Added this to avoid committing a bunch of Webstorm-derived configuration files

src/lib/dev-environment.js
```diff
@@ -64,6 +64,7 @@ export const defaults = {
 } );
 
 const landoFileTemplatePath = path.join( __dirname, '..', '..', 'assets', 'dev-environment.lando.template.yml.ejs' );
+const configDefaultsFilePath = path.join( __dirname, '..', '..', 'assets', 'config', 'wp-config-defaults.php' );
```
pschoffer: I am not sure about the config subfolder. It is a config but not for vip cli, which might be confusing. I think I would just place that under assets.

In both cases I would prefix it with `dev-environment` so that it is more clear what this asset is for.

So I would suggest:
```suggestion
const configDefaultsFilePath = path.join( __dirname, '..', '..', 'assets', 'dev-environment.wp-config-defaults.php' );
```

This means you would need to change the import as well.

What do you think?

pschoffer (changes requested):
Perfect, thanks @pauarge for catching/fixing that! I left one minor note.

src/lib/dev-environment.js
pauarge: I wasn't sure either, I think what you are suggesting is a better solution. Committing that.

pschoffer (dismissed):
Nice works great! I think we might still need a pattisery sign-off.

nejclovrencic (approved):

pschoffer (approved):
Great!

pschoffer:
I will merge this as I am doing some refactoring in the same place so it might be easier to merge now to avoid conflicts:)

rinatkhaziev:
r1602-stacks

PR #732, opened 2021-04-22
Refactors dev-env to split responsibilities

pschoffer:
## Description

Having all the dev-env logic in one file started to be too much. So I broke it up to 3 - lando specific, core and cli helpers.

## Steps to Test

1. I guess running through some of the dev-env subcommands verifying they still work


src/bin/vip-dev-environment-create.js
```diff
@@ -15,7 +15,8 @@ import chalk from 'chalk';
  * Internal dependencies
  */
 import command from 'lib/cli/command';
-import { createEnvironment, printEnvironmentInfo, getEnvironmentName } from 'lib/dev-environment';
+import { createEnvironment, printEnvironmentInfo } from 'lib/dev-env/dev-env-core';
```
pauarge: I know it might be a bit longer to write, but what do you think about naming the folder and the files `dev-environment` instead of `dev-env`? I think it's a good option to maintain consistency and call it the same everywhere.

src/lib/dev-env/dev-env-core.js
```diff
@@ -296,44 +205,6 @@ async function prepareLandoEnv( instanceData, instancePath ) {
...
-
-	const argv = {
-		_: args,
-	};
-
-	wpTask.run( argv );
-}
-
 export function generateInstanceData( slug: string, options: NewInstanceOptions ) {
 	const instanceData = {
```
pauarge: I know this was not introduced in the PR, but we could return the object directly and not declare the variable.

src/lib/dev-env/dev-env-core.js
```diff
@@ -43,7 +41,6 @@ const containerImages = {
 };
 
 export const defaults = {
```
pauarge: Maybe this constants (same for the above) could be extracted into a constants file?

pauarge (changes requested):
Great refactor, left some comments regarding minor changes.

src/bin/vip-dev-environment-create.js
pschoffer: Yeah I was on the fence with that one. I wanted to have it in the file name as well, so that would mean `lib/dev-environment/dev-environment-lando.js` for instance. I guess that is fine. Will change it ot that.

src/lib/dev-env/dev-env-core.js
pschoffer: That function will be totally rewritten in the prompt work I am doing now.

src/lib/dev-env/dev-env-core.js
pschoffer: Yep good idea.

pauarge (approved):
LGTM

nejclovrencic (approved):
Awesome! 😎  🚢

PR #733, opened 2021-04-23
Add/dev env propmts

pschoffer:
## Description

Adds awesome propmts to guide users to setup their dev-environment. All the provided parameters will be skipped and we would ask for the rest.

## Steps to Test

1. `vip dev-environment create`

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -117,14 +117,13 @@ export async function promptForArguments( providedOptions: NewInstanceOptions ) 
 		wpTitle: providedOptions.title || await promptForText( 'WordPress site title', DEV_ENVIRONMENT_DEFAULTS.title ),
 		phpVersion: providedOptions.phpVersion || await promptForText( 'Php version', DEV_ENVIRONMENT_DEFAULTS.phpVersion ),
 		multisite: providedOptions.multisite || await promptForBoolean( 'Multisite', DEV_ENVIRONMENT_DEFAULTS.multisite ),
-		// wordpress: providedOptions.wordpress ? processComponentOptionInput( providedOptions.wordpress, 'wordpress' ) : await promptForCode( 'wordpress' ),
-		// muPlugins: processComponentOptionInput( options.muPlugins, 'muPlugins' ),
-		// jetpack: processComponentOptionInput( options.jetpack, 'jetpack' ),
-		// clientCode: processComponentOptionInput( options.clientCode, 'clientCode' ),
 		wordpress: {},
+		muPlugins: {},
+		jetpack: {},
+		clientCode: {},
 	};
 
-	const components = [ 'wordpress' ];
+	const components = [ 'wordpress', 'muPlugins', 'jetpack', 'clientCode' ];
 	for ( const component of components ) {
 		const option = providedOptions[ component ];
 		instanceData[ component ] = option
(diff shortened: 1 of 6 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -53,3 +54,153 @@ export function printTable( data: Object ) {
...
+	phpVersion: string,
+	wordpress: string,
+	muPlugins: string,
+	jetpack: string,
+	clientCode: string
+}
+
+const promptIntro = 'This is a wizard to help you set up you local dev environment.\n\n' +
+	'Sensible defaualt values were pre-selected for convinience. ' +
+	'You can also choose to create multiple different environments with different settings using the --slug option.\n\n';
```
pauarge: Not sure if we should move this into the constant files.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -53,3 +54,153 @@ export function printTable( data: Object ) {
...
+		wpTitle: providedOptions.title || await promptForText( 'WordPress site title', DEV_ENVIRONMENT_DEFAULTS.title ),
+		phpVersion: providedOptions.phpVersion || await promptForText( 'Php version', DEV_ENVIRONMENT_DEFAULTS.phpVersion ),
+		multisite: providedOptions.multisite || await promptForBoolean( 'Multisite', DEV_ENVIRONMENT_DEFAULTS.multisite ),
+		wordpress: {},
+		muPlugins: {},
+		jetpack: {},
+		clientCode: {},
+	};
+
+	const components = [ 'wordpress', 'muPlugins', 'jetpack', 'clientCode' ];
```
pauarge, line 104: If this doesn't change, we can extract it from the function so we don't evaluate it each time. As the question above, not sure if leave it in this file or even in the constants file.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -53,3 +54,153 @@ export function printTable( data: Object ) {
...
+const promptIntro = 'This is a wizard to help you set up you local dev environment.\n\n' +
+	'Sensible defaualt values were pre-selected for convinience. ' +
+	'You can also choose to create multiple different environments with different settings using the --slug option.\n\n';
+
+export async function promptForArguments( providedOptions: NewInstanceOptions ) {
+	console.log( promptIntro );
+
+	const instanceData = {
+		wpTitle: providedOptions.title || await promptForText( 'WordPress site title', DEV_ENVIRONMENT_DEFAULTS.title ),
+		phpVersion: providedOptions.phpVersion || await promptForText( 'Php version', DEV_ENVIRONMENT_DEFAULTS.phpVersion ),
```
pauarge:
```suggestion
		phpVersion: providedOptions.phpVersion || await promptForText( 'PHP version', DEV_ENVIRONMENT_DEFAULTS.phpVersion ),
```

pauarge (changes requested):
Looks good, left a couple of suggestions in terms of code structure.

src/lib/dev-environment/dev-environment-cli.js
pschoffer: I am not sure about this one. The values in that array are kinda referenced in the `promptForComponent` function a lot so I kinda like that they are here close to each other.

What would be the benefit of moving it out of the function? I guess it would only get evaluated at a slightly different time (when the script requires this file as oppose to when it runs the function). But this is not something that would impact performance I assume, or?

What we could do is to move the individual components as a constants and constract an array of constants and use that in the code bellow as well. What do you think?

src/lib/dev-environment/dev-environment-cli.js
pschoffer: Done

src/lib/dev-environment/dev-environment-cli.js
pauarge: I think the performance benefit would be minimal (especially given that this is a CLI and will only be executed once anyway), so feel free to dismiss my point.

pauarge (approved):
LGTM

PR #734, opened 2021-04-23
Fetch image tags from Docker Hub in dev environment

pauarge:
## Description

This PR adds the ability for the `dev-environment create` command to fetch the latest available version of the [WP VIP WordPress image available in Docker Hub](https://hub.docker.com/r/wpvipdev/wordpress).

We are updating the command by doing a `GET` request to Docker Hub each time the user creates a new environment to fetch the latest available tag. 

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-dev-environment-create --slug test`
4. You should see a `.lando.yml` file created with the latest WP version (at the moment of this PR, `5.7.1`.


__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -17,11 +18,15 @@ import { getEnvironmentPath,
 	startEnvironment,
 	generateInstanceData,
 	destroyEnvironment,
-	getParamInstanceData,
-	getEnvironmentName } from 'lib/dev-environment/dev-environment-core';
+	getParamInstanceData } from 'lib/dev-environment/dev-environment-core';
```
pauarge: Removed `getEnvironmentName` because it wasn't being used.

pschoffer:
Nice! This looks great. However with the prompt work the underlying code change a lot. Would you mind rebasing it on https://github.com/Automattic/vip/pull/733?

Also, in the new code you will find we are using a new 'allTags' hardcodded array to offer these as options to the user. Do you think we could fetch the list of all of the tags instead of the latest one?

pauarge:
@pschoffer I've reviewed your PR, I will leave this one in draft and when yours gets merged, I can tackle this one fixing the conflicts (I don't think they will be too bad). 

As you point out, we can very easily fetch all the available tags. I wouldn't fetch _all_ of them but a subset, because the list could eventually be very long. If the user wants an (old) version that is not suggested, I'd leave the option to specify it manually.

nejclovrencic (approved):

PR #737, opened 2021-04-27
Fetching base info

pschoffer:
## Description

This change adds some basic app data lookup to guide user some more when setting up dev environment.

So far only name and isMultisite is supported.

When running `vip @app.env dev-environment create` we will prefill the the name in the following propts and set the correct default for multiste in the prompt according to the environment mentioned in `@app.env`.

## Steps to Test

Run `vip @<app>.<env> dev-environment create`


src/bin/vip-dev-environment-create.js
```diff
@@ -60,7 +60,19 @@ command()
 
 		debug( 'Args: ', arg, 'Options: ', opt );
 
-		const instanceData = await promptForArguments( opt );
+		let appInfo: any = {};
```
nejclovrencic, line 63: This is actually not needed, because `command()` already has an option to fetch an app and env. You just have to pass the GraphQL query to it, and it will return app and env info in `opt`.

Example:

```
command( { 	
  appContext: true,
  appQuery,
  envContext: true, 
} )
```

https://github.com/Automattic/vip/blob/master/src/bin/vip-import-sql.js#L362

pschoffer: Yeah, I looked into it, but if I would use it it would require you to provide both app AND environment on every command interaction. Which is not something I wanted. And marking it optional in the command seems like quite a hussle. But that is for sure a way of doing it.

nejclovrencic (approved):

PR #742, opened 2021-05-06
Prevent calling isVip when logout command is being executed

nejclovrencic:
## Description

Currently, we always call `await checkIsVIP()` when using `logout or help` command. If you are using an invalid token, and try to logout, it will fail as the checkIsVIP check fails with 401. In this case, there is no way to logout successfully. This PR fixes this by preventing call to `await checkIsVIP()` when logout command is ran.


pschoffer (approved):
Works great. Thank you!

PR #746, opened 2021-05-07
Update/site exists

pschoffer:
## Description

When we try to create dev env using `vip dev-environment create` we will check if the environment already exist BEFORE we start the whole wizard asking you about your environment rather than after.

## Steps to Test

Run `vip dev-environment create` (2x if you dont have default environment)


nejclovrencic (approved):

nejclovrencic:
🚢

PR #761, opened 2021-05-17
Dev-Env: Adds ability to do soft destroy

pschoffer:
## Description

Adds an ability to do soft destroy. This helps fix the broken docker state after a crash or unexpected termination of docker without the need to recreate the environment (and go through the creation wizzard again)

## Steps to Test

`vip dev-environment destroy --soft`
`vip dev-environment start` - should work


pauarge (approved):
Works fantastically, very appreciated addition!

PR #771, opened 2021-05-27
Refactor Import status reporting to support k8s

dchymko:
## Description
This updates the status query/reporting logic to work for sites on k8s tenancy. We no longer have the `jobs` property available to us on the GraphQL query so we need to adapt this to use `importStatus`  

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run an import: `/dist/bin/vip-import-sql.js @<site_id>.<env> <file.sql>`
1. Verify the import reports the progress correctly

### DIFF
--- a/src/lib/site-import/status.js
+++ b/src/lib/site-import/status.js
@@ -27,6 +27,7 @@ const IMPORT_SQL_PROGRESS_QUERY = gql`
 		app(id: $appId) {
 			environments(id: $envId) {
 				id
+				isK8sResident
 				jobs(types: "sql_import") {
 					id
 					type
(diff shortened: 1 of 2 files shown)


src/lib/site-import/status.js
```diff
@@ -236,17 +237,52 @@ ${ maybeExitPrompt }
...
 				}
-				const { importStatus, importJob } = status;
+				const { importStatus } = status;
+				let { importJob } = status;
+
+				let jobStatus, jobSteps = [];
+				if ( env.isK8sResident ) {
+					// in the future the API may provide this in k8s jobs so account for that.
+					// Until then we need to create the importJob from the status object.
+					if ( ! importJob ) {
```
dchymko, line 247: This _should_ allow us to send this object from the server in the future without requiring an update on the client.

netsuso (approved):
Looks great to me! This adapts quite nicely to the new k8s jobs, good work! :)

PR #780, opened 2021-06-02
Make sure failed analytics tracking request don't make the whole command fail

brunobasto:
## Description

Issues have been reported when users have their network block outgoing requests to our analytics endpoint. The analytics feature can be disabled via an [environment variable](https://github.com/Automattic/vip#analytics), but our default behavior shouldn't be to throw errors when that fails.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Block outgoing requests to `https://public-api.wordpress.com/rest/v1.1/tracks/record`
4. Run `DEBUG=* ./dist/bin/vip-app-list.js`
5. Verify that the command didn't error out


dchymko (approved):
This is great, thanks for jumping on this so quickly!

PR #789, opened 2021-06-08
Disable enterprise search by default

pschoffer:
## Description

Changes the default config to not enable enterprise search.

## Steps to Test

Create new environment, notice it is not using search.


rinatkhaziev:
LGTM but we mention the "new" (VIP_ENABLE_VIP_SEARCH) constant in our docs (https://docs.wpvip.com/how-tos/vip-search/enable/), let's match that here.

rinatkhaziev (approved):

rogertheriault:
I've tested locally - works as designed.

PR #790, opened 2021-06-09
Remove dual-path logic for checking import status

dchymko:
## Description
This restores the original single-path logic for checking the status of an import job. (Which is now possible due to the API now returning the job data consistently, regardless of site type

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js @<id> file.sql` for both site types.
1. Verify import progress is shown and that the imports work.


robersongomes (dismissed):
LGTM

joshbetz (dismissed):

joshbetz:
@dchymko should we merge this?

dchymko:
> @dchymko should we merge this?

We should deploy/release #812 first to provide a new version that won't be broken by changes related to jobSteps being provided by the API.

netsuso:
@dchymko Does this still apply? I see the branch #812 you mentioned is already merged, but I'm not sure if changes here are still needed

dchymko:
> @dchymko Does this still apply? I see the branch #812 you mentioned is already merged, but I'm not sure if changes here are still needed

~~Yes, this should be good now. I was wanting to be sure we had a few releases to minimize it breaking the CLI for users on previous versions.~~

Edit: Something isn't quite right here, its not outputting the steps correctly. I'll have to check the logic.

netsuso (dismissed):

PR #795, opened 2021-06-17
Handle parameter validation in a consistent way

pschoffer:
## Description
If we are not logged in (or not help or logout subcommand) we would go to `args.parse( process.argv );`

However as we have never registered subcommands `args.parse` will basically evaluate any argument as invalid.

This change uses the same validation logic in both flows.

## Steps to Test
1. Logout `vip logout`
1. Try to run command with argument e.g. `vip dev-env create --slug test`


nejclovrencic (approved):
LGTM! 🚢

PR #801, opened 2021-06-29
FORNO-759: Add logged in user details to all Tracks events

saroshaga:
## Description

This PR adds the logged in user information - specifically the user's VIP flag status - to all Tracks events

### DIFF
--- a/__tests__/lib/search-and-replace.js
+++ b/__tests__/lib/search-and-replace.js
@@ -17,7 +17,6 @@ import { searchAndReplace } from 'lib/search-and-replace';
 // Import prompt as a module since that's how we implement it in lib/search-and-replace.js,
 // as opposed to importing prompt.confirm on its own
 import * as prompt from 'lib/cli/prompt';
-import * as apiConfig from 'lib/cli/apiConfig';
 
 global.console = { log: jest.fn(), error: jest.fn() };
 
@@ -26,8 +25,6 @@ const testFilePath = path.resolve( fixtureDir, 'client-file-uploader', 'tinyfile
 
 jest.mock( 'node-fetch' );
 fetch.mockReturnValue( Promise.resolve( new Response( 'ok' ) ) );
-
-jest.spyOn( apiConfig, 'checkIfUserIsVip' ).mockResolvedValue( true );
 
 const binary = path.resolve(
 	fixtureDir,
--- a/__tests__/lib/search-and-replace.js
+++ b/__tests__/lib/search-and-replace.js
@@ -17,6 +17,7 @@ import { searchAndReplace } from 'lib/search-and-replace';
 // Import prompt as a module since that's how we implement it in lib/search-and-replace.js,
 // as opposed to importing prompt.confirm on its own
 import * as prompt from 'lib/cli/prompt';
+import * as apiConfig from 'lib/cli/apiConfig';
 
 global.console = { log: jest.fn(), error: jest.fn() };
 
@@ -25,6 +26,8 @@ const testFilePath = path.resolve( fixtureDir, 'client-file-uploader', 'tinyfile
 
 jest.mock( 'node-fetch' );
 fetch.mockReturnValue( Promise.resolve( new Response( 'ok' ) ) );
+
+jest.spyOn( apiConfig, 'checkIfUserIsVip' ).mockResolvedValue( true );
 
 const binary = path.resolve(
 	fixtureDir,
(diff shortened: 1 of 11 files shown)


chrean (dismissed):
👍

saroshaga:
Dismissed the earlier review because I had to fix the tests

src/lib/cli/apiConfig.js
```diff
@@ -64,6 +63,11 @@ export async function checkIsVIP(): Promise<boolean> {
 	return !! isVip;
 }
 
+export async function checkIfUserIsVip() {
+	const res = await featureFlags.get();
```
mjangda, line 67: Is this cached? If not, we'll be adding a lot of requests to every event tracking call which will probably slow things down a fair bit.

mjangda (changes requested):

src/lib/cli/apiConfig.js
saroshaga: Good point! What do you think should be a reasonable TTL for this? 1 day? or maybe even more?

saroshaga: @mjangda , you did mean caching on Parker, right? Do we have a caching mechanism in place for the VIP CLI? 🤔

mjangda: I meant in the CLI.

If I do:

```
await checkIfUserIsVip();
await checkIfUserIsVip();
await checkIfUserIsVip();
await checkIfUserIsVip();
await checkIfUserIsVip();
```

will that make 5 HTTP requests?

mjangda: Looks like the `featureFlags.get()` makes an API request to Parker via the Apollo Client which we have set up to use an in-memory cache (https://github.com/Automattic/vip/blob/4955be2eaae70466e597b3112fb557dcd92a8556/src/lib/api.js#L70), so I think we should be okay.

mjangda: Or, maybe not :)

Without this PR (one call to Parker):

```
❯ DEBUG=agent-base ./dist/bin/vip-app.js 309 > /dev/null
  agent-base Resolving socket for 'https:' request: 'POST /graphql' +0ms
```

With this PR (3 extra calls to Parker, with an added time of ~600ms):

```
❯ DEBUG=agent-base ./dist/bin/vip-app.js 309 > /dev/null
  agent-base Resolving socket for 'https:' request: 'POST /graphql' +0ms
  agent-base Resolving socket for 'https:' request: 'POST /graphql' +336ms
  agent-base Resolving socket for 'https:' request: 'POST /graphql' +158ms
  agent-base Resolving socket for 'https:' request: 'POST /graphql' +295ms
```

mjangda: We probably need a basic cache within the feature flags lib. Can just store in memory without the need to persist across CLI commands (i.e. just store the feature flags data in a variable in the module).

saroshaga: Thanks @mjangda !! Got sidetracked by another higher priority PR. Will get back on this soon!

saroshaga: So the Apollo Client's in memory cache does work. But because of the way we use the Client library in the `feature-flag.js` file (and most other files TBH) - by instantiating a new object of the Parker API library every time we make an API call, we cannot leverage the cached objects. 

Please see https://github.com/Automattic/vip/pull/801/commits/802cf280bd8c1642e0d4fb63640c5d8554df68d8

__tests__/lib/search-and-replace.js
```diff
@@ -26,6 +27,8 @@ const testFilePath = path.resolve( fixtureDir, 'client-file-uploader', 'tinyfile
 jest.mock( 'node-fetch' );
 fetch.mockReturnValue( Promise.resolve( new Response( 'ok' ) ) );
 
+jest.spyOn( apiConfig, 'checkIfUserIsVip' ).mockResolvedValue( true );
```
mjangda: This feels annoying if we have to remember to do it for every test suite (it will lead to inevitably lead to wasted time wondering "why is my test suite failing").

Can we globally mock the module instead?

mjangda (changes requested):
Looks much better! Just left a comment about the apiConfig spy.

src/lib/cli/apiConfig.js
mjangda: Nice fix! Thanks!

__tests__/lib/search-and-replace.js
saroshaga: @mjangda : This is done in 0deca2f

had to create a separate setup file for jest mocks because it was messing up the env variable mocks in the `jest.setup.js` file

saroshaga: thanks! good point!

mjangda (approved):
Looks good! Thanks for all the changes!

PR #803, opened 2021-07-01
Validate Path to a component

pschoffer:
## Description

To make the command a bit more user-friendly I added validations that the local paths provided (for instance for client-code) point to a non-empty directory.

The next step (PR) is to offer a way to clone the code for client-code and or create a folder if it doesn't exist.

## Steps to Test

1. run `vip dev-env create` 
2. Provide incorrect paths/paths to files/paths to an empty directory
3. Same is supported for arguments

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -10,6 +10,7 @@ import chalk from 'chalk';
 import formatters from 'lando/lib/formatters';
 import { prompt, Confirm, Select } from 'enquirer';
 import debugLib from 'debug';
+import fs from 'fs';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


brunobasto (approved):
Tested locally and LGTM!

Just out of curiosity, can you explain why the directory has to be non-empty?

Thank you, @pschoffer!

pschoffer:
Absolutely @brunobasto!

This piece of code is invoked when we are setting up a local dev environment and we would provide a local folder to be mapped as one of the components. The component could be a WordPress core code, the client code, or mu-plugins. All of those are required to set up vip go like environment.

Checking that the folder is non-empty is a very basic sanity check that the path provided points to something at least. Ideally for example we would check that the path to the local folder of WordPress code actually contains core source files, but this is simpler and works for all components.

brunobasto:
> Absolutely @brunobasto!
> 
> This piece of code is invoked when we are setting up a local dev environment and we would provide a local folder to be mapped as one of the components. The component could be a WordPress core code, the client code, or mu-plugins. All of those are required to set up vip go like environment.
> 
> Checking that the folder is non-empty is a very basic sanity check that the path provided points to something at least. Ideally for example we would check that the path to the local folder of WordPress code actually contains core source files, but this is simpler and works for all components.

Got it. Awesome. Thank you!

PR #804, opened 2021-07-01
Make dev-env start more resilient

pschoffer:
## Description

There is this a dev-env issue when the crash of docker or Lando itself somehow based on [this issue](https://github.com/lando/lando/issues/971) can cause docker networks to get deleted. `lando start` then can't correctly spin up some services (most common proxy network gets deleted, causing proxy container not starting).

This change replaces the behavior of `vip dev-env start` to internally run the equivalent of `lando rebuild` instead of `lando start`. This way we always (as far as we know) get the correct setup. The downside is slightly longer spin up time, but nothing too bad.

Another benefit is that rebuilding does NOT release volumes (which are so far suggested workaround with `destroy` did), meaning we get to keep the database intact. 

This also introduces `--skip-rebuild` flag that is used to fallback to previous logic of only starting inactive services.

## Steps to Test

1. Stop dev-env `vip dev-env stop`
1. destroy netwroks `docker network prune`
2. Start dev-env `vip dev-env start` 
3. See the environment still works

### DIFF
--- a/src/bin/vip-dev-env-start.js
+++ b/src/bin/vip-dev-env-start.js
@@ -30,14 +30,18 @@ const examples = [
 
 command()
 	.option( 'slug', 'Custom name of the dev environment' )
+	.option( 'skip-rebuild', 'Only start stopped services' )
 	.examples( examples )
 	.argv( process.argv, async ( arg, opt ) => {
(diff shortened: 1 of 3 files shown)


pauarge (approved):
LGTM

PR #808, opened 2021-07-13
FORNO-779: Throttle request to Parker when fetching media import status

nejclovrencic:
## Description

This PR throttles request to Parker, when fetching media import status. After 2 minutes, we start reducing the poll interval by a second, every minute.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run media import status command
4. Verify polling is reducing by observing network traffic

### DIFF
--- a/src/lib/media-import/status.js
+++ b/src/lib/media-import/status.js
@@ -22,6 +22,8 @@ import { capitalize, formatEnvironment, formatData } from 'lib/cli/format';
 import { RunningSprite } from '../cli/format';
 
 const IMPORT_MEDIA_PROGRESS_POLL_INTERVAL = 1000;
+const ONE_MINUTE_IN_MILLISECONDS = 1000 * 60;
+const TWO_MINUTES_IN_MILLISECONDS = 2 * ONE_MINUTE_IN_MILLISECONDS;
 
 const IMPORT_MEDIA_PROGRESS_QUERY = gql`
 	query App( $appId: Int, $envId: Int ) {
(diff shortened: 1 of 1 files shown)


src/lib/media-import/status.js
```diff
@@ -235,11 +239,22 @@ ${ maybeExitPrompt }
...
 				overallStatus = status;
 
-				setTimeout( checkStatus, IMPORT_MEDIA_PROGRESS_POLL_INTERVAL );
+				// after two minutes, we'll start decreasing the pollInterval
+				pollIntervalDecreasing = startDate < ( Date.now() - TWO_MINUTES_IN_MILLISECONDS );
+
+				// decrease poll interval by a second, every minute
+				if ( pollIntervalDecreasing && startDate < ( Date.now() - ONE_MINUTE_IN_MILLISECONDS ) ) {
+					pollInterval = pollInterval + IMPORT_MEDIA_PROGRESS_POLL_INTERVAL;
+					startDate = Date.now();
```
saroshaga, line 248: Haven't tested this out and just going on the basis of a dry run.. but wouldn't this reset the `startDate` for the two minute mark check?

saroshaga: So I think this logic will actually increment the interval by 1000 milliseconds every two minutes

@nejclovrencic

nejclovrencic: I had an IF first and did some refactoring, but forgot to add the condition for that. Fixed now, thanks!

saroshaga: ah okay thanks!

saroshaga (approved):
Thanks for the changes @nejclovrencic !! 

🚀

PR #809, opened 2021-07-27
[dev-env] handle multisite = false correctly

pschoffer:
## Description

If the `--multisite false` or `--multisite=false` is provided we should interpret it as **not** multisite and not prompt for it in the wizard. 

`--multisite` or `--multisite no` or `--multisite=true` or `--multisite=whatever` should result in true and no prompt.

No `--multisite` at all should prompt,

## Steps to Test

Try the options above when creating dev-env


brunobasto (approved):
LGTM! Tested and works as intended.

Thanks @pschoffer!

PR #811, opened 2021-08-01
Update DB Engine check to reduce false positives

### LINKED ISSUE #805: False Positive when Validating SQL
## Expected/Desired Behavior

1. Expected to validate SQL.

## Actual Behavior

Flagged tables in SQL that have an invalid engine though the table has a valid engine.

1. Import a SQL file with some content that includes an engine query string. Here is some example content that matches the current regex: https://regexr.com/6144d

Example string in content that validates SQL things is not InnoDB.

```
http://www.renditionmovie.com/?engine=adwords!10023
```

dchymko:
## Description
There are some cases where the `ENGINE=InnoDB` check was triggering a false positive, such as querystrings e.g. `?engine=adwords`

Fixes #805 

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Create a test SQL import file containing a previously-flagged string https://github.com/Automattic/vip/blob/master/src/lib/validations/sql.js#L230
1. Run `./dist/bin/vip-import-validate-sql.js <testfile.sql>`
1. Verify the check passes

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -227,7 +227,7 @@ const checks: Checks = {
 		recommendation: '',
 	},
 	engineInnoDB: {
-		matcher: /ENGINE=(?!(InnoDB))/i,
+		matcher: / ENGINE=(?!(InnoDB))/i,
 		matchHandler: lineNumber => lineNumber,
 		outputFormatter: errorCheckFormatter,
 		results: [],


src/lib/validations/sql.js
```diff
@@ -227,7 +227,7 @@ const checks: Checks = {
 		recommendation: '',
 	},
 	engineInnoDB: {
-		matcher: /ENGINE=(?!(InnoDB))/i,
+		matcher: / ENGINE=(?!(InnoDB))/i,
```
brunobasto, line 230: Not sure I got this right, but if the leading space may or may not be there, what about

```js
/\s?ENGINE=(?!(InnoDB))/i
```

dchymko: That still flags it as an error with the example given in the issue. This isn't perfect and of course can be skipped with the `--skip-validate` option, but will alleviate the current issue. There may still be an issue with it not catching this if a newline starts with it, but this is probably the closest we can do for now,

brunobasto: Got it. Thanks for clarifying!

PR #812, opened 2021-08-01
Retrieve the status and steps regardless of the site type.

dchymko:
## Description
This fixes a bug where the import status and steps aren't correctly deconstructed in all cases

This ensure that regardless of site type or if jobsSteps is provided by the API, they will be deconstructed correctly 

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js @<id> file.sql` where id is a k8s-based site
1. Run `./dist/bin/vip-import-sql.js @<id> file.sql` where id is a non-k8s-based site
1. Verify import returns the progress correctly.

### DIFF
--- a/src/lib/site-import/status.js
+++ b/src/lib/site-import/status.js
@@ -275,15 +275,13 @@ ${ maybeExitPrompt }
 
 						importJob.progress = { status: jobStatus, steps: jobSteps };
 					}
-				} else {
-					if ( ! importJob ) {
-						return resolve( 'No import job found' );
-					}
-
-					( {
-						progress: { status: jobStatus, steps: jobSteps },
-					} = importJob );
+				} else if ( ! importJob ) {
+					return resolve( 'No import job found' );
 				}
+
+				( {
+					progress: { status: jobStatus, steps: jobSteps },
+				} = importJob );
 
 				createdAt = importJob.createdAt;
 				completedAt = importJob.completedAt;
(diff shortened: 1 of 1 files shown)


src/lib/site-import/status.js
```diff
@@ -275,16 +275,14 @@ ${ maybeExitPrompt }
...
-					} = importJob );
+				} else if ( ! importJob ) {
+					return resolve( 'No import job found' );
 				}
 
+				( {
+					progress: { status: jobStatus, steps: jobSteps },
+				} = importJob );
+
 				createdAt = importJob.createdAt;
```
brunobasto, line 284: Maybe we could use the same syntax as above to be consistent?

```js
( {
    createdAt,
    completedAt,
    progress: { status: jobStatus, steps: jobSteps },
} = importJob );
```
Or maybe the same syntax as below?

```js
jobStatus = importJob.progress.jobStatus;
jobSteps = importJob.progress.jobSteps;
createdAt = importJob.createdAt;
completedAt = importJob.completedAt;
```

brunobasto (dismissed):
LGTM! Left a minor comment

src/lib/site-import/status.js
dchymko: Good idea, the deconstruction can be somewhat confusing in that format anyway. Changed in 7fc429e

PR #814, opened 2021-08-03
Adds a check for an orphaned proxy container

pschoffer:
## Description

This is another step to make `vip dev-env` more stable. There is a scenario where proxy network is removed without removing the proxy container itself. During a start/rebuild then the process fails to start the proxy container due to missing network.

This change detects the stopped proxy container and removes it as part of the rebuild call. 

## Steps to Test

1. Stop proxy container `landoproxyhyperion5000gandalfedition_proxy_1`
1. Run `vip dev-env start` 
1. Notice the proxy container is removed and later recreated

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -64,6 +64,8 @@ export async function landoRebuild( instancePath: string ) {
 	const app = lando.getApp( instancePath );
 	await app.init();
 
+	await ensureNoOrphantProxyContainer( lando );
+
 	await app.rebuild();
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -165,3 +167,33 @@ export async function landoExec( instancePath: string, toolName: string, args: A
...
+ * can safelly add a network and a new proxy container.
+ *
+ * @param {object} lando Bootstrapped Lando object
+ */
+async function ensureNoOrphantProxyContainer( lando ) {
+	const proxyContainerName = lando.config.proxyContainer;
+
+	const docker = lando.engine.docker;
+	const containers = await docker.listContainers( { all: true } );
+	const proxyContainerExists = containers.some( container => container.Names.includes( `/${ proxyContainerName }` ) );
```
pauarge, line 186: Not sure this will work, but just asking

```suggestion
	const proxyContainerExists = containers.some( container => container.Names.includes( proxyContainerName ) );
```

pauarge (changes requested):
Looks good, just a small question

src/lib/dev-environment/dev-environment-lando.js
pschoffer: Nah, I tried that and it didn't work. The names returned have some sort of namespaces so `/` for no namespace.

pauarge (approved):
LGTM

PR #818, opened 2021-08-10
Using new VIP Docker images for dev-env

pauarge:
## Description

This PR changes the Docker images that are used in the developer environment to the ones in https://github.com/Automattic/vip-container-images. 

Some other relevant changes:

- Removed the option to fetch Jetpack from a remote image.
- Removed the option to run different PHP Versions (fixed to 7.4 for now).
- Hardcoded the supported WordPress versions.

## Steps to Test

1. Check out PR and install `vip` from it.
2. Developer Environment should use the new images and work normally.


pschoffer:
Great work! So much good stuff in this one.

For the jetpack, I would go even further. Dropping the prompt and the ability to use the local folder also. 

In "real life" clients can still use custom jetpack I believe using some constants, but then it is taken from client-code instead of mu-plugins, so no need to introduce some special way of doing it in `dev-env`.

And even for us when testing the new JP we can just change it in mu-plugins that are mapped locally.

I guess there could be an edge case where we failed to publish the newest versions of JP in time and the client would like to test it already. That scenario is very unlikely I think. What do you think?

pauarge:
@pschoffer I really like the simplification approach for Jetpack. At the end of the day, we're introducing a way of adding Jetpack just for dev-env, which goes against the spirit of replicating the production environment. I am pushing a change to do it this way.

pschoffer (approved):
Amazing!

PR #822, opened 2021-08-13
Increasing dev-env PMA upload limit to 4G

pauarge:
## Description

PHPMyAdmin on the development environment would fail to import big SQL files. This PR adds an explicit value to the maximum upload size of the file. By default, that size is 2M and we are bumping it to 4G.

## Steps to Test

Example:

1. Check out PR.
2. Create a new environment
3. Try to upload a big SQL file on PHPMyAdmin

### DIFF
--- a/assets/dev-environment.lando.template.yml.ejs
+++ b/assets/dev-environment.lando.template.yml.ejs
@@ -55,6 +55,9 @@ services:
     type: phpmyadmin
     hosts:
       - database
+    overrides:
+      environment:
+        UPLOAD_LIMIT: 128M
 
   vip-search:
     type: elasticsearch:custom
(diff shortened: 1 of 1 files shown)


pschoffer (dismissed):
Great stuff! We could even make that configurable or maybe just put some huge number like 128G? I don't think we have really a reason to limit the size in the local setup. What do you think?

pauarge:
My approach here would be not to bother our users with yet another configurable bit. This is that kind of number that you only look at when you hit the limit, so I agree that setting a sufficiently big number would be enough.

Regarding what number to set, I'd wouldn't set something ridiculously high because it might not be realistic to expect that kind of files in prod.

pschoffer (approved):
Awesome!

PR #823, opened 2021-08-13
Fixes issue where user is unable to login after logout

brunobasto:
## Description

[v2.0.11](https://github.com/Automattic/vip/releases/tag/v2.0.11) introduced a bug where users would be unable to login after logging out. The root cause was a `checkIfUserIsVip()` call added to `trackEvent()`. When a token wasn't set, the request would return with `500` and the process would immediately exit. 

The solution here checks if a token is set before making a request to check if a user is VIP or not. Also, I noticed that after a successful login, we would also get an error message:

```
Unauthorized: You are unauthorized to perform this request, please logout with `vip logout` then try again.
```

The cause of that was that the request headers were only being set once at startup. I've used Apollo's `setContext` and middleware APIs to set the headers dynamically so that, after a token is set, subsequent requests succeed.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip.js logout`
4. Run `./dist/bin/vip.js`
5. You should be able to login now. Before, you would get an error and would be unable to login.

### DIFF
--- a/src/lib/cli/apiConfig.js
+++ b/src/lib/cli/apiConfig.js
@@ -65,7 +65,7 @@ export async function checkIsVIP(): Promise<boolean> {
 }
 
 export async function checkIfUserIsVip() {
-	let token = await Token.get();
+	const token = await Token.get();
 
 	if ( token && token.valid() ) {
 		const res = await featureFlags.get();
@@ -73,7 +73,7 @@ export async function checkIfUserIsVip() {
 		return res?.data?.me?.isVIP;
 	}
 
-	return false;	
+	return false;
 }
 
 export async function exitWhenFeatureDisabled( featureName: string ): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


src/lib/cli/apiConfig.js
```diff
@@ -64,8 +65,15 @@ export async function checkIsVIP(): Promise<boolean> {
...
 
 export async function checkIfUserIsVip() {
-	const res = await featureFlags.get();
-	return res?.data?.me?.isVIP;
+	const token = await Token.get();
+
+	if ( token && token.valid() ) {
+		const res = await featureFlags.get();
+
+		return res?.data?.me?.isVIP;
```
chriszarate: to ensure boolean
```suggestion
		return !! res?.data?.me?.isVIP;
```

chriszarate (dismissed):

brunobasto:
Merging since got approval from @chriszarate (dismissed because of a test fix). Also, this is kind of critical since it's preventing users from using the tool. Talked to @mjangda about expediting this fix even outside our release window.

PR #825, opened 2021-08-16
Sets up a volume for media files

pschoffer:
## Description

Sets up a volume for uploads folder. This in combination with a [change](https://github.com/Automattic/vip-container-images/pull/3) in dev-tools image enables media uploads.

## Steps to Test

1) Create a new dev environment
2) Verify uploading media files work


pauarge (approved):
LGTM

PR #829, opened 2021-08-19
Path resolving fixes

pschoffer:
## Description

The recent addition of path validation made resolving broken. This change moves up the path resolution closer to the user imput.

## Steps to Test

Try different paths when creating new environment for local components.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -11,6 +11,8 @@ import formatters from 'lando/lib/formatters';
 import { prompt, Confirm, Select } from 'enquirer';
 import debugLib from 'debug';
 import fs from 'fs';
+import path from 'path';
+import os from 'os';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -183,6 +194,13 @@ async function processComponent( component: string, option: string ) {
 	return result;
 }
 
+export function resolvePath( input: string ): string {
+	// Resolve does not do ~ reliably
+	const resolvedPath = input.replace( /^~/, os.homedir() );
```
pauarge, line 199: Is this already handled by some built-in function?

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -273,7 +291,7 @@ export async function promptForComponent( component: string ) {
 	};
 }
 
-async function getLatestImageTags( component: string ): string[] {
+function getLatestImageTags( component: string ): string[] {
```
pauarge, line 294: 👍

src/lib/dev-environment/dev-environment-cli.js
pschoffer: Doesn't seem to from what I have found online - https://stackoverflow.com/a/37112508/11875271

Like there are some tools probably, but it seems kinda simple to introduce a new dependency.

pauarge (approved):
LGTM

PR #831, opened 2021-08-19
Disable statsd by default

pschoffer:
## Description

Statsd is not that useful container for end-users of dev-env. However, it is a bit resource intensive. So we are disabling it by default. Providing `--statsd` during creation will enable statsd.

## Steps to Test

VErify statsd is not included in new dev environments.


pauarge (approved):
That's awesome. Might have conflicts with this one, though: https://github.com/Automattic/vip/pull/828

rinatkhaziev:
@pschoffer My CPU thanks you from the bottom of its silicon heart.

PR #832, opened 2021-08-19
Support ES version option

pschoffer:
## Description

Sets the default version of ElasticSearch to 7.5.1 (to match production environment), also introduces `--elastic-search` option to choose a version during the creation of dev-environment.

## Steps to Test

Try to use `--elastic-search` on dev-env creation.

### DIFF
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -58,6 +58,7 @@ command()
 	.option( 'phpmyadmin', 'Enable PHPMyAdmin component. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'xdebug', 'Enable XDebug. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'elastic-search', 'Explicitly choose ElasticSearch version to use' )
+	.option( 'elasticsearch', 'Explicitly choose Elasticsearch version to use' )
 	.examples( examples )
 	.argv( process.argv, async ( arg, opt ) => {
 		const slug = getEnvironmentName( opt );
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -57,6 +57,7 @@ command()
 	.option( 'statsd', 'Enable statsd component. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'phpmyadmin', 'Enable PHPMyAdmin component. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'xdebug', 'Enable XDebug. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
+	.option( 'elastic-search', 'Explicitly choose ElasticSearch version to use' )
 	.examples( examples )
 	.argv( process.argv, async ( arg, opt ) => {
 		const slug = getEnvironmentName( opt );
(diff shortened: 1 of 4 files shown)


pauarge (changes requested):
That looks great, but I'm working on the ARM version of that image and this might change a bit. Do you mind holding this PR off?

src/bin/vip-dev-env-create.js
```diff
@@ -54,6 +54,7 @@ command()
 	.option( 'wordpress', 'Use a specific WordPress version or local directory (default: last stable)' )
 	.option( 'mu-plugins', 'Use a specific mu-plugins changeset or local directory (default: "auto": last commit in master)' )
 	.option( 'client-code', 'Use the client code from a local directory or VIP skeleton (default: use the VIP skeleton)' )
+	.option( 'elastic-search', 'Explicitly choose ElasticSearch version to use' )
```
rinatkhaziev, line 60:
```suggestion
	.option( 'elasticsearch', 'Explicitly choose Elasticsearch version to use' )
```

src/lib/constants/dev-environment.js
```diff
@@ -19,6 +19,7 @@ export const DEV_ENVIRONMENT_DEFAULTS = {
 	title: 'VIP Dev',
 	multisite: false,
 	phpVersion: '7.4',
+	elasticSearchVersion: '7.5.1',
```
rinatkhaziev:
```suggestion
	elasticsearchVersion: '7.5.1',
```

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -119,7 +119,8 @@ type NewInstanceOptions = {
 	php: string,
 	wordpress: string,
 	muPlugins: string,
-	clientCode: string
+	clientCode: string,
+	elasticSearch: string,
```
rinatkhaziev:
```suggestion
	elasticsearch: string,
```

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -152,6 +153,7 @@ export async function promptForArguments( providedOptions: NewInstanceOptions, a
 	const instanceData = {
 		wpTitle: providedOptions.title || await promptForText( 'WordPress site title', name || DEV_ENVIRONMENT_DEFAULTS.title ),
 		multisite: 'multisite' in providedOptions ? providedOptions.multisite : await promptForBoolean( multisiteText, multisiteDefault ),
+		elasticSearch: providedOptions.elasticSearch || DEV_ENVIRONMENT_DEFAULTS.elasticSearchVersion,
```
rinatkhaziev:
```suggestion
		elasticsearch: providedOptions.elasticsearch || DEV_ENVIRONMENT_DEFAULTS.elasticsearchVersion,
```

rinatkhaziev:
There's no capital `S` in Elasticsearch :) I added suggestions (hoping I didn't miss any). 

Also, thoughts about making an alias of `es` for the argument?

pschoffer:
> There's no capital S in Elasticsearch :) I added suggestions (hoping I didn't miss any).

Oh I do this every time. Thank you!

> Also, thoughts about making an alias of es for the argument?

There is no native support for option aliases in the args parser we use. We could define another option `-es` and kinda make in this function meant the same thing. But that feels unnecessary to me. What do you think? @rinatkhaziev

rinatkhaziev:
> There is no native support for option aliases in the args parser we use. We could define another option `-es` and kinda make in this function meant the same thing. But that feels unnecessary to me. What do you think? @rinatkhaziev

Ah if there isn't one  then let's not worry about it. I may be confusing with _some other_ CLI that has `.alias()` method...

pschoffer:
@rinatkhaziev There are aliases to the command not options though - https://www.npmjs.com/package/args

@pauarge, please let me know when the ES image is ready.

pauarge (dismissed):
We are not there yet with the custom ES image, so it's OK to merge this now.

pschoffer:
Only rebased since reviewed so merging anyway :yolo:

PR #835, opened 2021-08-20
Removing mu-plugins test command

pauarge:
## Description

This PR removes the `test` command from the dev-environment. That command is supposed to run the tests on mu-plugins, but it doesn't work. To run the `mu-plugins` test, please refer to that repository for instructions.


pschoffer (approved):
Great

PR #836, opened 2021-08-20
Making PHPMyAdmin optional on dev-env

pauarge:
## Description

This PR makes the PHPMyAdmin container optional i order to reduce resource consumption when running an environment. That means it will be disabled by default and users can enable it by

`vip dev-env create --phpmyadmin`

## Steps to Test

1. Check out PR.
2. Create an environment with PMA and one without it.
3. Check that in each case, a Docker container with PMA is running (or not).


pschoffer (approved):
Looks great!

PR #837, opened 2021-08-20
Patching Docker for Windows in dev-env

pauarge:
## Description

Docker for Windows with the WSL 2 back-end would fail to boot the Elasticsearch container due to [some known misconfiguration](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#_set_vm_max_map_count_to_at_least_262144). 

This PR tweaks the start script to detect if its running on Windows. If so, it runs a simple Powershell command to modify the virtual memory kernel flag.

## Steps to Test

1. Check out PR on a Windows machine.
2. Create an environment and start it.
3. Check that the Elasticsearch container launches properly.

### DIFF
--- a/assets/docker-es-patch.ps1
+++ b/assets/docker-es-patch.ps1
@@ -1,0 +1,1 @@
+wsl -d docker-desktop bash -c "sysctl -w vm.max_map_count=262144"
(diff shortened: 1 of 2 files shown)


pschoffer (approved):
Looks good. I assume it is ok to run multiple times right (on each start)?

pauarge:
@pschoffer Exactly. It would simply keep overriding that value.

PR #839, opened 2021-08-24
Use official MariaDB image and enable version selection on dev-env

pauarge:
## Description

This PR switches the MariaDB image that's being used in the dev environment, from Bitnami's to the official one. The biggest difference being that the official image supports ARM architectures.

We are also introducing the possibility to select a specific MariaDB version. By default, we'll use 10.3.

## Steps to Test

1. Check out PR.
2. Create a new environment.
3. Verify that the environment and the database work as expected.


src/bin/vip-dev-env-create.js
```diff
@@ -57,8 +57,8 @@ command()
 	.option( 'statsd', 'Enable statsd component. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'phpmyadmin', 'Enable PHPMyAdmin component. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
 	.option( 'xdebug', 'Enable XDebug. By default it is disabled', undefined, value => 'false' !== value?.toLowerCase?.() )
-	.option( 'elastic-search', 'Explicitly choose ElasticSearch version to use' )
```
pauarge, line 60: This is a leftover from a previous PR that we don't need.

rinatkhaziev (approved):
🥇

PR #840, opened 2021-08-24
Use official Elasticsearch image on dev-env

pauarge:
## Description

This PR switches the Elasticsearch image that's being used in the dev environment, from Bitnami's to the official one. The biggest difference being that the official image supports ARM architectures. We are also bumping the default ES version to `7.10.1`.

## Steps to Test

1. Check out PR.
2. Create a new environment.
3. Verify that the environment and vip-search work as expected.


src/lib/constants/dev-environment.js
```diff
@@ -19,7 +19,7 @@ export const DEV_ENVIRONMENT_DEFAULTS = {
 	title: 'VIP Dev',
 	multisite: false,
 	phpVersion: '7.4',
-	elasticsearchVersion: '7.5.1',
+	elasticsearchVersion: '7.10.1',
```
rinatkhaziev, line 21: Is 7.10.1 the earliest available ARM build? I'm asking cause we decided to lock it closer to prod?

pauarge: The first ARM-compatible version is `7.8`. Given that we're not using `7.8` in prod, I thought `7.10` would be a better fit.

rinatkhaziev (dismissed):
🎉

pschoffer (approved):
LGTM. I tested a bit and it works well. The only issue is that we are diverging from prod version as discussed, but  guess that is a good tradeoff for supporting more machines.

PR #842, opened 2021-08-25
Removing PHP parameter from dev environment

pauarge:
## Description

The developer environment on the CLI doesn't support specifying a PHP version for now (it's fixed to 7.4). For that reason, we're getting rid of this parameter in the CLI that is not being used anywhere.


pschoffer (approved):
Nice

PR #844, opened 2021-08-26
Expose DB and expose extra services in info table

pschoffer:
## Description

This PR brings in two changes:
* Exposes random port to connect to database
* Rewrites the way we expose extra services in info table. The previous implementation didn't work for custom services.

## Steps to Test

Setup new dev-env and check the info table (e.g. after start). It should contain database row now:
```
 NAME               test                                                                                      
 LOCATION           /home/pavel/.local/share/vip/dev-environment/test                                         
 SERVICES           devtools, nginx, php, database, memcached, vip-search, wordpress, mu-plugins, client-code 
 NGINX URLS         http://test.vipdev.lndo.site/                                                             
 ENTERPRISE SEARCH  http://127.0.0.1:49167                                                                    
 DATABASE           127.0.0.1:49166                                                                           
 STATUS             UP      
```


rinatkhaziev (changes requested):
Thanks, looks good! 

On thing though, this should be able to expose other services ideally in a non-hardcoded. I.e. for phpMyAdmin, or Mailhog, or whatever extra service us/customers decide to add to the config. 

Perhaps we can switch the logic from the "allow list" (i.e. hardcoded services we want to show) to "deny list" where we specifically not show the mappings for stuff like `wordpress, mu-plugins, client-code`, and show the rest.

pschoffer:
@rinatkhaziev done. Turns out we don't need to skip that many services, as we only list the ones with mapped ports to host anyway.

rinatkhaziev (approved):
Awesome

PR #846, opened 2021-08-26
dev-env: Update error message for directory prompt

rebeccahum:
## Description

Currently, the error message is `Provided path "${ resolvedPath }" does not point to a non-empty directory.`. This reads to be confusing to me because it is a [double-negative](https://en.wikipedia.org/wiki/Double_negative).

I think it would be clearer to state that the directory inputted is invalid or doesn't exist.


rinatkhaziev (approved):
Here was a witty comment but I remembered this is a public repo.

Thanks for the catch!

PR #848, opened 2021-08-30
Not using a prefix to all dev environments

pauarge:
## Description

All environments created by the CLI have the prefix `vipdev`. That was initially designed to avoid conflicts, but what it ends up doing is creating very long and sometimes hard to read names:

Captura de Pantalla 2021-08-30 a les 13 17 23

This PR removes that prefix.

## Steps to Test

1. Check out PR.
2. Create a new environment.
3. Check it the container names don't have the prefix.


pschoffer (approved):
LGTM

PR #849, opened 2021-08-30
Adding SQL import to dev environment

pauarge:
## Description

This PR adds the ability to import SQL files containing a WordPress database into the dev environment. 

The added command relies on `wp db import` and relies in the fact that Lando mounts the `$HOME` folder to `/user` in the Docker machine. Thus, this command is essentially an alias to `vip dev-env exec -- wp db import somefile.sql` but with some path handling for that file.

## Steps to Test

1. Check out PR and build the CLI.
2. Create a new environment.
3. Get a SQL that contains a WordPress site.
4. Import the site by running `vip dev-env import path-to-file.sql`


rinatkhaziev:
Nice! 

Thoughts on adding search-replace on the domain names?  Having an easy import is a good first step, but if we want the experience to be smooth we should also provide an option for search-replace of the domains.

__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -6,8 +6,9 @@
  * External dependencies
```
pschoffer, line 6: Oh tests, how nice!

src/bin/vip-dev-env-import-media.js
```diff
@@ -0,0 +1,23 @@
...
+ * Internal dependencies
+ */
+import command from '../lib/cli/command';
+
+command( {
+	requiredArgs: 1,
+} )
+	.option( 'slug', 'Custom name of the dev environment' )
+	.argv( process.argv, async ( unmatchedArgs: string[], opt ) => {
+		console.log( 'Importing media to a dev environment is not available yet.' );
```
pschoffer, line 22: Maybe we should drop this if not supported? But I guess keeping it is fine as well.

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -264,3 +270,36 @@ export async function getApplicationInformation( appId: number, envType: string
 
 	return appData;
 }
+
+export async function resolveImportPath( slug: string, fileName: string, searchReplace: string, inPlace: boolean ): Promise<SQLImportPaths> {
+	let resolvedPath = path.resolve( fileName );
```
pschoffer: I had issues with path.resolve vs `~`, could you please use [this](https://github.com/Automattic/vip/blob/fcd2cd5084b51305a9d380ff9769d6211c552208/src/lib/dev-environment/dev-environment-cli.js#L201) instead

Some info - https://stackoverflow.com/a/37112508/11875271

pschoffer (commented):
Left a minor note about path resolve. Otherwise I tested it and it imports great.

src/bin/vip-dev-env-import-media.js
pauarge: I'm working on adding this feature right now, so I think it's ok. We won't be releasing a new version of the CLI in the meantime.

pschoffer (dismissed):
Nice!

pschoffer (approved):

PR #850, opened 2021-08-30
Use official memcached image on dev-env

pauarge:
## Description

This PR replaces the memcached Bitnami image that's being used by default by Lando for the official memcached image. The main reason for this change is ARM support of the official image. We are also bumping the version (from 1.15 to 1.16) and moving from the Debian-base image to the Alpine-based one.

## Steps to Test

1. Check out PR.
2. Create a new environment.
3. Check that WordPress works correctly and the new environment spins up.
4. Set and get keys with `wp cache get/set`


pschoffer (approved):
Looks good and works on my linux.

PR #856, opened 2021-09-02
Adding media import command on dev environment

pauarge:
## Description

This PR adds a simple way of importing a media folder into a dev environment. It operates by copying a provided folder (ideally, a `wp-contents/uploads/` folder) into the proper directory of the environment.

This command works well when it's run in conjunction with `vip dev-env import sql`.

## Steps to Test

1. Check out PR and build the CLI.
2. Get an `uploads` folder from a WP install.
3. Run `vip dev-env import media path/to/wp-contents/uploads/`.
4. Check the files are available on the media library.

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -333,5 +333,5 @@ export async function importMediaPath( slug: string, filePath: string ) {
 
 	console.log( `${ chalk.yellow( '-' ) } Started copying files` );
 	copydir.sync( resolvedPath, uploadsPath );
-	console.log( `${ chalk.green( '✓' ) } Files successfuly copied to ${ uploadsPath }.` );
+	console.log( `${ chalk.green( '✓' ) } Files successfully copied to ${ uploadsPath }.` );
 }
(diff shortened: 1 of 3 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -303,3 +307,31 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
...
+	if ( ! fs.existsSync( resolvedPath ) || ! fs.lstatSync( resolvedPath ).isDirectory() ) {
+		throw new Error( 'The provided path does not exist or it is not valid (see "--help" for examples)' );
+	}
+
+	const files = fs.readdirSync( resolvedPath );
+	if ( files.indexOf( uploadPathString ) > -1 ) {
+		const confirm = await prompt( {
+			type: 'confirm',
+			name: 'continue',
+			message: 'The provided path contains an uploads folder inside. Do you want to continue?',
```
pschoffer, line 323: What is meant by this? Is it that we suspect user probably pointed to the wrong folder? Like `/path/to/uploads` to `/path/to` instead of `/path/to/uploads`?

pauarge: Exactly. We want to prevent data ending up in a path like `wp-content/uploads/uploads`. It's a warning, however, so user can decide to go ahead.

pschoffer: Got it.. makes sense!

pschoffer (approved):
Great stuff

PR #857, opened 2021-09-02
Remove select DB checks

dchymko:
## Description
Removes the check for `CREATE DATABASE` and `USE DATABASE`, as we now trim these lines out in the import process

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Create/locate an sql import file with lines containing the above statements
1. Run `./dist/bin/vip-sql-import @site.env filename.sql`
1. Verify the import passes validation


djalmaaraujo (approved):
LGTM. Tested locally, only got the CREATE TABLE, DROP TABLE errors, looking good.

PR #862, opened 2021-09-08
Fix lint warnings

chriszarate:
## Description

Fixes all remaining lint warnings. They were pretty noisy when running `npm run lint` locally. Most of the warnings were related to variable length. Also removing unused `eslines` config.

If we're amenable, I would recommend upgrading warnings to errors to keep lint output noise-free.

## Steps to Test

1. Check out PR.
1. Run `npm run lint`
1. Verify delicious lack of output.


pauarge (approved):
💯

PR #866, opened 2021-09-10
Flow type checking is broken 😢 

chriszarate:
## Description

While working in this repo, I encountered some really strange behavior where Flow errors would sometimes surface locally but disappear on subsequent runs—or appear in WebStorm but not in Vim. I shrugged them off because Flow type checks always seemed to pass in CI. However, this was a mirage: 🏝️ 

```sh
$ npm run flow

> @automattic/vip@2.0.14 flow /Users/czarate/Documents/Code/vip
> flow

Launching Flow server for /Users/czarate/Documents/Code/vip
Spawned flow server (pid=24473)
Logs will go to /private/tmp/flow/zSUserszSczZaratezSDocumentszSCodezSvip.log
Monitor logs will go to /private/tmp/flow/zSUserszSczZaratezSDocumentszSCodezSvip.monitor_log
No errors!
```

The output says no errors, but it notes that logs are sent to a temp file. If we inspect that file, we can see that all is not well:

```
[snipped]
[2021-09-10 13:08:07.053] Persistent response: liveErrorsResponse OK (3 errors, 0 warnings, 0 other) file:///Users/czarate/Documents/Code/vip/src/lib/validations/site-type.js
[2021-09-10 13:08:07.077] Persistent response: liveErrorsResponse OK (2 errors, 0 warnings, 0 other) file:///Users/czarate/Documents/Code/vip/src/lib/validations/sql.js
```

It appears the culprit is our use of `resolve_dirname` in `.flowconfig`:

https://github.com/facebook/flow/issues/5180#issuecomment-341887514

We are essentially telling Flow to treat those files as `node_modules`, which means the type checking is skipped in many contexts. This is not a bug, but intended behavior.

The fix is to move to `module.name_mapper` instead, which shakes loose an avalanche of type errors. I'm happy to work on them in separate PRs but wanted to check to make sure effort would not be better spent migrating to TypeScript. (Good news: Our other projects that use Flow do not use `resolve_dirname`.)

## Steps to Test

1. Check out PR.
1. Run `npm run flow`
1. 😭


chriszarate:
For fun I followed [these steps](https://dev.to/danielhauser/convert-a-react-app-from-flow-to-typescript-without-losing-git-history-32i1) to convert Flow to TypeScript, and the result was ... impressive! Only 41 TypeScript errors vs. 123 Flow errors. I haven't examined this closely, and there would also need to be work to fix the build/publish chain (TypeScript could probably remove the need for Babel, I think).

https://github.com/Automattic/vip/tree/improve/switch-to-typescript

brunobasto:
> For fun I followed [these steps](https://dev.to/danielhauser/convert-a-react-app-from-flow-to-typescript-without-losing-git-history-32i1) to convert Flow to TypeScript, and the result was ... impressive! Only 41 TypeScript errors vs. 123 Flow errors. I haven't examined this closely, and there would also need to be work to fix the build/publish chain (TypeScript could probably remove the need for Babel, I think).
> 
> https://github.com/Automattic/vip/tree/improve/switch-to-typescript

+1 for typescript ❤️

brunobasto:
We should probably include the fixes here before merging, or look into switching to typescript. Thanks for fixing the validation @chriszarate!  

cc @dchymko @pandah3 @djalmaaraujo

pandah3:
Hmm yeah, a myriad of problems 🤔 Thank you for investigating this Chris!
Have you tried out any of the commands locally during your look into switching to Typescript? Definitely don't want to break existing flows 🙈

pandah3 (commented):
Hmm yeah, a myriad of problems 🤔 Thank you for investigating this Chris!
Have you tried out any of the commands locally during your look into switching to Typescript? Definitely don't want to break existing flows 🙈

sjinks:
Related: #977

PR #868, opened 2021-09-13
persist database data in between container restarts

pschoffer:
## Description

We added a volume to DB so that the changes in DB get persisted between stop and start as the volume stays around.

## Steps to Test

1) Create new environment- `vip dev-env create --slug test`
2) Start it: `vip dev-env start --slug test`
3) Create a new post
4) Stop it: `vip dev-env stop --slug test`
5) Start it again: `vip dev-env start --slug test`
6) See your new post


pschoffer:
A downside of this solution is that we are losing health checks as those seem to not be fully supported on custom services. 

On the other hand, the lando service is not playing nice with the volume on official image. 

More info here:
https://github.com/lando/lando/issues/3117

chriszarate (approved):
Nice! Are there instructions about how to remove the volume when a full rebuild is desired? That can be tricky when working with Docker directly (maybe Lando makes this easy).

pschoffer:
@chriszarate both `vip dev-env destroy --soft` and  `vip dev-env destroy` should remove the volumes and therefore clear the DB.

PR #869, opened 2021-09-14
[dev-env] Media redirect to production site. 

pschoffer:
## Description

We would like to redirect the requests for media to the prod sites if not found locally. This is one way of doing it. We are overwriting the nginx config for images to try first locally otherwise redirect to the prod site:

`try_files $uri @prod_site;`

**Note:** For this to work we need a change in container repo: https://github.com/Automattic/vip-container-images/pull/18 - this is in now.

## Steps to Test

Example:

1. Create a site: `vip dev-env create --image-redirect-domain pschoffer-develop.go-vip.net`
2. Go to: `http://test.vipdev.lndo.site/wp/picture.jpg`
3. You will be redirected to `https://pschoffer-develop.go-vip.net/wp/picture.jpg`


rinatkhaziev (changes requested):
Thanks, the code looks good but I think we should rename the parameter to signify it's a domain, because to me `--media-redirect` sounds like a boolean, something like 
`--media-domain` or `--media-redirect-domain`.

On a separate note, let's auto-populate that value for `@ENV_ID` notation when creating the new environment.

pschoffer:
`--media-redirect-domain` sounds good to me. I Will change to that.

> On a separate note, let's auto-populate that value for @ENV_ID notation when creating the new environment.

Yeah, I was thinking about that. I think we should still prompt in the wizard for it if we have @env_id. Sort of "Do you want to redirect missing media request to pschoffer-develop.go-vip.net? (y/n)".

Just to avoid some confusion with magic happening on its own. What do you think?

pschoffer:
@rinatkhaziev I pushed the rename + tested with the new nginx container. Works great!

I plan to leave out the domain from env for a separate PR, we will probably need to add extra logic to get a primary domain for the env from parker first.

assets/dev-env.nginx.template.conf.ejs
```diff
@@ -0,0 +1,13 @@
+<% if ( mediaRedirectDomain ) { %>
+
+location ~* \.(png|jpg|jpeg|gif|ico)$ {
```
rinatkhaziev: I think we should try to find out what we have in the location block for static assets in production and try to match that. This will likely be enough for the majority of the cases but it's not complete.

rinatkhaziev:
> Just to avoid some confusion with magic happening on its own. What do you think?

Yeah, that sounds perfect!

assets/dev-env.nginx.template.conf.ejs
pschoffer: That is what we have in the default config - https://github.com/Automattic/vip-container-images/blob/9194f47bbf3d5832501456dad19d4e68ee8ce032/nginx/default.conf#L59

Or what do you mean?

pschoffer: I tested with  `/wp-content/uploads` and it works great.

rinatkhaziev (dismissed):
Yay! Thanks

rinatkhaziev (approved):

PR #872, opened 2021-09-17
[dev-env] Auto flush cache after import and add vipgo user

pschoffer:
## Description

This change adds to convenience things to `vip dev-env import sql` :
* Automatically flush cache after import
* If parameter `--add-vipgo-user` is provided adds admin user

The admin user is quite often needed to avoid 2fa restriction on users from import, which might not work that well locally.

## Steps to Test

Try `vip dev-env import sql <file> --add-vipgo-user`


src/bin/vip-dev-env-import-sql.js
```diff
@@ -43,6 +43,7 @@ command( {
 	.option( 'slug', 'Custom name of the dev environment' )
 	.option( 'search-replace', 'Perform Search and Replace on the specified SQL file' )
 	.option( 'in-place', 'Search and Replace explicitly on the given input file' )
+	.option( 'add-vipgo-user', 'Add vipgo:password user to the site to enable easy access' )
```
pschoffer: I am not sure if we should do this conditionally as an option or just always add this user. What do you think?

rinatkhaziev: I think always, TBH. I actually stumbled upon it last night and thought "it'd be nice to keep the user". How are you reading my mind?

pschoffer: Done.

rinatkhaziev (approved):

PR #876, opened 2021-09-20
Update envvar list command to only show names

chriszarate:
## Description

Product feedback: Masking values and caveating the command made it confusing for users. Instead, the `list` command should only show the names and not act like a command that could potentially show values.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip config envvar list`

### DIFF
--- a/src/bin/vip-config-envvar-list.js
+++ b/src/bin/vip-config-envvar-list.js
@@ -42,7 +42,7 @@ export async function listEnvVarsCommand( arg: string[], opt ): void {
 	debug( `Request: list environment variables for ${ getEnvContext( opt.app, opt.env ) }` );
 	await trackEvent( 'envvar_list_command_execute', trackingParams );
 
-	const envvars = await listEnvVars( opt.app.id, opt.env.id, opt.format )
+	const envvars = await listEnvVars( opt.app.id, opt.env.id )
 		.catch( async err => {
 			rollbar.error( err );
 			await trackEvent( 'envvar_list_query_error', { ...trackingParams, error: err.message } );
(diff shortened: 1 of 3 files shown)


brunobasto (approved):
LGTM! Tested with `./dist/bin/vip-config-envvar-list.js --format=ids` and `./dist/bin/vip-config-envvar-list.js`. Works great!

src/bin/vip-config-envvar-list.js
```diff
@@ -57,13 +57,18 @@ export async function listEnvVarsCommand( arg: string[], opt ): void {
 		process.exit();
 	}
 
-	// Display context for non-machine-readable formats.
-	if ( [ 'keyValue', 'table' ].includes( opt.format ) ) {
-		console.log( 'For security, the values of environment variables cannot be retrieved.' );
-		console.log();
+	// Vary data by expected format.
+	let key = 'name';
+	if ( 'keyValue' === opt.format ) {
```
brunobasto, line 62: `keyValue` and `ids` doesn't seem to make as much sense. I liked `table`, and `csv` sounds useful, but when I try `--format=csv` I get basically a table without the borders. Is that expected?

chriszarate: Yeah, I agree, with a single column of data, some of these formats aren't very useful. If we're thinking about the use case of piping this data into another script (like Node or Bash or Python), then I think `ids` and `csv` are probably the most useful.

`keyValue` really doesn't make much sense. But the `format` arg is shared among lots of commands and isn't configurable. Maybe I'll look into a way to customize it in a separate PR.

PR #882, opened 2021-09-23
Update message to keep in line with docs

saroshaga:
## Description

Changes the wording on the Media Import prompt to keep in line with public docs: https://docs.wpvip.com/how-tos/migrate-content-media/#h-directory-structure

> The directory to be imported must be named uploads.
 In most cases, media uploaded to a WordPress install is stored in a directory uploads/, which is a subdirectory of wp-content/. To prepare media files for import to VIP, an archived copy of the uploads/ directory is required. If you do not place the uploads/ folder at the root of the archive, the import tool will import the archive to the uploads/ directory by default, using whatever folder structures is present


chrean (approved):
👍

PR #888, opened 2021-10-05
[dev-env] mount wordpress code

pschoffer:
Local WP source code is not working on dev-env. The reason is that we do some patches and add extra files so that wp works similar to our prod site.

Therefore we will drop the option to link to the local codebase, but will instead mount wp source files automatically to the env location so that folks can change those to debug/test.

# Test
* Create new environemnt.
* Notice the wizzard is only prompting for a version of WP not local vs image anymore.
* In the environment location there is a `wordpress` folder containing the mount of wp source files.

**Note:** The user permissions are not happy with this as we have a rather complex system of images now, and they mount into each others file system. Some files ended up owned by root which makes `destroy` that removes those to fail. The solution is to run install not as root, but that breaks copying to mu-plugins. That's why there is that `run_as_root` change. - https://github.com/Automattic/vip-container-images/pull/27

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -106,17 +106,13 @@ services:
       command: node stats.js /config/statsd-config.js
 <% } %>
 
-<% if ( wordpress.mode == 'image' ) { %>
   wordpress:
     type: compose
     services:
       image: <%= wordpress.image %>:<%= wordpress.tag %>
       command: sh -c "rsync -a /wp/ /shared/; sleep infinity"
       volumes:
-        - wordpress:/shared
-    volumes:
-      wordpress: {}
-<% } %>
+        - ./wordpress:/shared
 
 <% if ( muPlugins.mode == 'image' ) { %>
   mu-plugins:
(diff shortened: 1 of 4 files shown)


assets/dev-env.lando.template.yml.ejs
```diff
@@ -106,17 +108,13 @@ services:
       command: node stats.js /config/statsd-config.js
 <% } %>
 
-<% if ( wordpress.mode == 'image' ) { %>
```
pauarge, line 109: 👍

package-lock.json
```diff
@@ -10576,7 +10576,7 @@
         },
         "onetime": {
           "version": "1.1.0",
-          "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
+          "resolved": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
```
pauarge, line 10579: We probably don't want to commit the changes on this file

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -239,32 +240,42 @@ const componentDisplayNames = {
...
 
-	const select = new Select( {
-		message: `How would you like to source ${ componentDisplayName }`,
-		choices,
-		initial,
-	} );
+	let modeResult = initial;
+	const selectMode = choices.length > 1;
+	if ( selectMode ) {
+		const initialIx = choices.findIndex( choice => choice.value === initial );
```
pauarge: Could we change the variable name to something more verbose?

pauarge (changes requested):
Some small nit-picks

src/lib/dev-environment/dev-environment-cli.js
pschoffer: How's this?

pauarge (approved):

PR #889, opened 2021-10-12
[dev-env] Proxy config change

pschoffer:
After the change in images, we can now support non-standard ports. 

This PR now only changes the proxy name so we switched to `vip-dev-env-proxy` (instead of arguably way cooler `landoproxyhyperion5000gandalfedition`)

## Steps to Test

1) Stop your dev-env
2) Check your proxy name in docker

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -41,11 +41,6 @@ function getLandoConfig() {
 			},
 		],
 		proxyName: 'vip-dev-env-proxy',
-		// This is a way to make lando config merging actually not include any of its fallback
-		// Each item in the array will overwrite one fallback port. And there are 4 default fallbacks
-		// for both http an https each.
-		proxyHttpFallbacks: [ '80', '80', '80', '80' ],
-		proxyHttpsFallbacks: [ '443', '443', '443', '443' ],
 	};
 }
 
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -40,6 +40,12 @@ function getLandoConfig() {
 				subdir: '.',
 			},
 		],
+		proxyName: 'vip-dev-env-proxy',
+		// This is a way to make lando config merging actually not include any of its fallback
+		// Each item in the array will overwrite one fallback port. And there are 4 default fallbacks
+		// for both http an https each.
+		proxyHttpFallbacks: [ '80', '80', '80', '80' ],
+		proxyHttpsFallbacks: [ '443', '443', '443', '443' ],
 	};
 }
 


jesse-greathouse (dismissed):
Nicely done, Pavel.

rinatkhaziev:
@pschoffer are there any other changes in behavior that could result from this PR?

Is it possible to look into fixing the underlying issue? 
As a user, I would love dev env start up on a random port in case 80 or 443 are already bound.

On the WordPress site that may require a filter for the home_url and site_url to include the port but I think that's pretty doable.

pschoffer:
@rinatkhaziev it is a bit tricky... the current problem is this:
* Say your proxy will run on `http://vip-local.vipdev.lndo.site:8080` 
* `vip-local.vipdev.lndo.site:8080` gets DNS resolved to  #`127.0.0.1:8080`
* `127.0.0.1:8080` is bound to the proxy so that all works
* **proxy** will forward the request internally within docker network to `nginx`.
* **WordPress** will take over, and `redirect_canonical` will try to redirect to `http://vip-local.vipdev.lndo.site:80`, because that is the port WP is expecting to run
*  `vip-local.vipdev.lndo.site:80` gets DNS resolved to  `127.0.0.1:80`
* `127.0.0.1:80` is NOT bound to the proxy so that will fail

* however I managed to suppress that redirect with a snippet bellow.
* the Second problem that I couldn't figure out how to solve is that the WordPress will still request all the static files and all the links will point to  `http://vip-local.vipdev.lndo.site:80` ( e.g. - `http://vip-local.vipdev.lndo.site/wp-includes/css/dist/block-library/style.min.css?ver=5.7.2`) 


* Would one of the filters you suggested change the URL to both suppress the `redirect_canonical` and point to the correct port? If yes the sub-issue there is that we don't get to know what port will proxy be bound to before we do `lando start`. There might be a way to add some sort of lando hook, after the proxy has started. In the hook we might be able to kinda dynamically generate/update the dev-env mu plugin to implement some filter(s) so that the different port is considered correct.


What do you think?





```
add_filter( 'redirect_canonical', function( $redirect_url, $requested_url ) {
    $redirect_port = parse_url( $redirect_url, PHP_URL_PORT );
    $request_port = parse_url( $requested_url, PHP_URL_PORT );
    if( $redirect_port !== $request_port ) {
        return false;
    }

    return $redirect_url;
}, 10, 2);
```

rinatkhaziev:
@pschoffer This should help: https://github.com/Automattic/vip-container-images/pull/28/files

pschoffer:
@rinatkhaziev cool that does help for frontend!! The `wp-admin` still gets redirected incorrectly.

```
$ curl -v http://test.vipdev.lndo.site:8000/wp-admin
*   Trying 127.0.0.1:8000...
* TCP_NODELAY set
* Connected to test.vipdev.lndo.site (127.0.0.1) port 8000 (#0)
> GET /wp-admin HTTP/1.1
> Host: test.vipdev.lndo.site:8000
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Content-Length: 169
< Content-Type: text/html
< Date: Fri, 15 Oct 2021 07:35:54 GMT
< Location: http://test.vipdev.lndo.site/wp-admin/
< Server: nginx/1.21.3
< 
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.21.3</center>
</body>
</html>
* Connection #0 to host test.vipdev.lndo.site left intact
```

However, that looks like nginx redirect (no WP headers). I'll check some more.

rinatkhaziev (approved):
So long Hyperion, it was nice knowing you.

PR #894, opened 2021-10-21
[dev-env] support windows db import

pschoffer:
## Description

When trying to import on Windows machine - `vip dev-env imposrt sql \Users\<user>\dump.sql` we encounter an error resolving the path. Due to the way this works we need to construct in container path to the import file which doesn't work if `\` instead of `/` are systems path separators.

This change does the manual conversion of `\` to `/` to address that.

One note there is a limitation due to the way the file is processed that the file needs to be under home directory. I will try to document that more in public docs.

## Steps to Test

Example:

1. Downgrade to Windows
1. Run `vip dev-env import sql <path to file>`
2. It works

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -322,9 +322,8 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
 	/**
 	 * Docker container does not have acces to the host filesystem.
 	 * However lando maps os.homedir() to /user in the container. So if we replace the path in the same way
-	 * docker container will get the file from within the mapped volume under /user.
+	 * in the Docker container will get the file from within the mapped volume under /user.
 	 */
-
 	let inContainerPath = resolvedPath.replace( os.homedir(), homeDirPathInsideContainers );
 	if ( path.sep === '\\' ) {
 		// Because the file path generated for windows will have \ instead of / we need to replace that as well so that the path inside the container (unix) still works.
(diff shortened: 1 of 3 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -317,10 +319,22 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
 		fs.renameSync( outputFileName, resolvedPath );
 	}
 
-	const dockerPath = resolvedPath.replace( os.homedir(), '/user' );
+	/**
+	 * Docker container does not have acces to the host filesystem.
+	 * However lando maps os.homedir() to /user in the container. So if we replace the path in the same way
+	 * docker container will get the file from within the mapped volume under /user.
+	 */
+
```
pauarge:
```suggestion
	/**
	 * Docker container does not have acces to the host filesystem.
	 * However lando maps os.homedir() to /user in the container. So if we replace the path in the same way
	 * in the Docker container will get the file from within the mapped volume under /user.
	 */
```

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -317,10 +319,22 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
...
+	let inContainerPath = resolvedPath.replace( os.homedir(), homeDirPathInsideContainers );
+	if ( path.sep === '\\' ) {
+		// Because the file path generated for windows will have \ instead of / we need to replace that as well so that the path inside the container (unix) still works.
```
pauarge, line 329:
```suggestion
		// Because the file path generated for Windows will have \ instead of / we need to replace that so that the path inside the container (Linux) still works.
```

pauarge (changes requested):
Changes look great, left a couple of nit-picky suggestions.

Also, I think it would be worth adding an additional unit test that covers the Windows case.

pschoffer:
How's this @pauarge ?

pauarge (approved):
Looks great, thanks for adding the tests!

PR #895, opened 2021-10-22
Fix rmdir deprecation warning

pauarge:
## Description

With Node 14, the `rmdir` method is deprecated and it will likely be removed in subsequent Node releases. We are anticipating that by replacing it by an equivalent non-deprectaed method, `rm`. This is the deprecation warning you'd get:

```
(node:23769) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

Notice that we are adding a callback function to tell the user if the deletion process went well or something failed.

## Steps to Test

1. Check out PR and run it in Node 16 (prior versions only had the method deprecated in the docs).
2. Run `npm run build` and `npm link`.
3. Destroy any environment.
4. See that you get a success/error message but you don't get the deprecation warning.

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -10,7 +10,6 @@ import debugLib from 'debug';
 import xdgBasedir from 'xdg-basedir';
 import os from 'os';
 import fs from 'fs';
-import fsPromises from 'fs/promises';
 import ejs from 'ejs';
 import path from 'path';
 import chalk from 'chalk';
@@ -129,7 +128,7 @@ export async function destroyEnvironment( slug: string, removeFiles: boolean ) {
 	await landoDestroy( instancePath );
 
 	if ( removeFiles ) {
-		await fsPromises.rm( instancePath, { recursive: true } );
+		await fs.promises.rm( instancePath, { recursive: true } );
 		console.log( `${ chalk.green( '✓' ) } Environment files deleted successfully.` );
 	}
 }
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -128,8 +128,14 @@ export async function destroyEnvironment( slug: string, removeFiles: boolean ) {
 	await landoDestroy( instancePath );
 
 	if ( removeFiles ) {
-		// $FlowFixMe: Seems like a Flow issue, recursive is a valid option and it won't work without it.
-		fs.rmdirSync( instancePath, { recursive: true } );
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -128,8 +128,14 @@ export async function destroyEnvironment( slug: string, removeFiles: boolean ) {
 	await landoDestroy( instancePath );
 
 	if ( removeFiles ) {
-		// $FlowFixMe: Seems like a Flow issue, recursive is a valid option and it won't work without it.
-		fs.rmdirSync( instancePath, { recursive: true } );
+		fs.rm( instancePath, { recursive: true }, err => {
```
pschoffer: What do you think about converting it to promise? 

Sort of:
```
try {
   await new Primise( (resolve, reject) => fs.rm( instancePath, { recursive: true }, err => { if (err) { reject(err) } else {resolve}}););
   
  console.log( `${ chalk.green( '✓' ) } Environment files deleted successfully.` );
} catch (e) {
   console.log( `${ chalk.red( '✕' ) } Could not delete the environment files.` );
   console.log( err );
}
```

I **think** something like that might work, but I haven't tested it. I am happy to do that if you think it is a good idea.

pschoffer (changes requested):

src/lib/dev-environment/dev-environment-core.js
pauarge: That's an interesting approach, however it looks cleaner to me in the current form because we don't need a `try/catch` block and we don't need to use `await/Promise`. What advantages do you see to it?

pschoffer: I think the biggest benefit is that reads sequentially. so that if we want to do something after files are deleted you do it after. Oh there is actually this version - https://nodejs.org/api/fs.html#fspromisesrmpath-options

Also I think we could drop the error print and `try/catch` too. And leave the try/catch one level up to handle error printing in unified way - https://github.com/Automattic/vip/blob/3df1328f579319982d71db6a040d9331336bac74/src/bin/vip-dev-env-destroy.js#L51

pauarge: Good point. I've committed the changes, let me know what you think. It looks super clean and straight forward.

pschoffer: Awesomesouce

pschoffer (approved):

PR #897, opened 2021-10-22
[dev-env] Primary domain prompt for primary domain redirect

pschoffer:
## Description

When using the `@` notation we should prompt for domain. Eg:

```
$ vip dev-env @2891 create 

✔ WordPress site title · pschoffer-develop.go-vip.net
✔ Multisite (pschoffer-develop.go-vip.net is NOT multisite) (y/N) · false
✔ Would you like to redirect to pschoffer-develop.go-vip.net for missing media files? (Y/n) · true

```

### DIFF
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -84,6 +84,7 @@ command()
 			debug( `WARNING: ${ message }`, error.message );
 			console.log( chalk.yellow( 'Warning:' ), message );
 		}
+		debug( 'App Info', appInfo );
 
 		const instanceData = await promptForArguments( opt, appInfo );
 		const instanceDataWithSlug = {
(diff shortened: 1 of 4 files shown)


src/bin/vip-dev-env-create.js
```diff
@@ -84,6 +84,7 @@ command()
 			debug( `WARNING: ${ message }`, error.message );
 			console.log( chalk.yellow( 'Warning:' ), message );
 		}
+		debug( 'App Info', appInfo );
```
pauarge: Is this a development leftover?

pauarge (changes requested):
Small question, but looking good

src/bin/vip-dev-env-create.js
pschoffer: Yes:)

pschoffer: Removed

pauarge (approved):
Looking good

PR #899, opened 2021-10-26
[dev-env] Proxy fix + healthchecks

pschoffer:
## Description

When http_proxy is configured for docker `vip_search` would try to route its healthcheck to that addres as it is doing `curl localhot:9200`.

That doesn't work as we need to reach out to the container's own localhost.

This change switches from using lando's own healthcheck system to the docker's + added hook that checks those states before proceeding with the setup. 

The change also brings DB health check back.

## Steps to Test

1) `npm run build && ./dist/bin/vip-dev-env-create.js --slug test`
2) `npm run build && ./dist/bin/vip-dev-env-start.js --slug test`
3) See ... 
```
Waiting for service vip-search ...
Waiting for service database ...
Waiting for service vip-search ...
Waiting for service database ...
```

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -60,6 +60,8 @@ services:
       command: docker-entrypoint.sh mysqld
       ports:
         - ":3306"
+      healthcheck:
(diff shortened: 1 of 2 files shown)


pauarge (approved):
This is awesome!

PR #902, opened 2021-10-27
[dev-env] Delete file permissions

pschoffer:
## Description

This change:
* simplifies the lando file creation
* Uses new images for mu-plugins and dev-tools
* Combined with that above some other tweaks to make sure files will be own by the current user

## Steps to Test

1) Create env
2) Start env
3) Do some stuff
4) Restart env
5) Delete env

__no errors anywhere__

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -12,7 +12,7 @@ services:
   devtools:
     type: compose
     services:
-      image: ghcr.io/automattic/vip-container-images/dev-tools:0.7
+      image: ghcr.io/automattic/vip-container-images/dev-tools:0.8
       command: sleep infinity
       volumes:
         - devtools:/dev-tools
@@ -49,9 +49,6 @@ services:
 <% wpVolumes() %>
     run:
       - sh /dev-tools/setup.sh database root "http://<%= siteSlug %>.vipdev.lndo.site/" "<%= wpTitle %>" <% if ( multisite ) { %> <%= siteSlug %>.vipdev.lndo.site <% } %>
-    run_as_root:
-      - echo "Copying dev-env-plugin.php to mu-plugins"
-      - cp /dev-tools/dev-env-plugin.php /wp/wp-content/mu-plugins/
 
   database:
     type: compose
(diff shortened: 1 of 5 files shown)


assets/dev-env.lando.template.yml.ejs
```diff
@@ -116,17 +113,17 @@ services:
...
       volumes:
         - ./wordpress:/shared
 
 <% if ( muPlugins.mode == 'image' ) { %>
   mu-plugins:
     type: compose
     services:
-      image: <%= muPlugins.image %>:<%= muPlugins.tag %>
-      command: sh -c 'rsync -a /mu-plugins/ /shared/; sh /autoupdate.sh /shared'
+      image: ghcr.io/automattic/vip-container-images/mu-plugins:0.1
```
pauarge, line 125: mu-plugins should always be `latest`, given it's always-updating nature

src/lib/constants/dev-environment.js
```diff
@@ -1,38 +1,22 @@
...
 	muPlugins: {
-		image: 'ghcr.io/automattic/vip-container-images/mu-plugins',
-		tag: 'latest',
+		mode: 'image',
 	},
 	clientCode: {
-		image: 'ghcr.io/automattic/vip-container-images/skeleton',
-		tag: 'latest',
+		mode: 'image',
 	},
```
pauarge: If the three are 'image', we might simplify this and specify nothing?

pauarge (changes requested):
Looking good, left a couple of questions.

assets/dev-env.lando.template.yml.ejs
pschoffer: Well not really. The mu-plugins code will be self updating but the containe itself has some other scripts and structure that I would like to start version. Here - https://github.com/Automattic/vip-container-images/pull/35 - we changed both how the docker entrypoint `run.sh` (that was unused up till now) and autoupdating works (it didn't before).

If we in the future would like to introduce some breaking change say change the folder that contains the autoupdated code from `/shared` to `/mu-plugins` we would like to NOT break all the dev-envs in the wild already expecting the folder to be `/shared`.

src/lib/constants/dev-environment.js
pschoffer: Oh yeah! We dont use those anymore.

pauarge (approved):

PR #908, opened 2021-11-01
[dev-env] Custom user permissions setup

pschoffer:
In some cases lando's user permission setup fails to map `www-data` user to the current running user.

This causes `dev-env` unusable and undeletable.

One scenario that the user mapping can fail is that user ID is higher than what `alpine` supports by default.

In this change we map `/scritps` from dev-tools to containers that need `www-data` user. Our script is able to create the user even with the high ID as we are able to adjust the script to the specific containers we are using.

## Steps to Test

1) Create and start new environments.
2) [optional] Create and start new environment while your user's OS ID is over `256000`
   - Depends on your OS how, but something like this:
    - `sudo useradd high-id-user -M -N -l -r -u 42000002`
    - `su high-id-user`

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -49,10 +49,10 @@ services:
           volume:
             nocopy: true
         - type: volume
-            source: scripts
-            target: /scripts
-            volume:
-              nocopy: true
+          source: scripts
+          target: /scripts
+          volume:
+            nocopy: true
 <% wpVolumes() %>
     run:
       - sh /dev-tools/setup.sh database root "http://<%= siteSlug %>.vipdev.lndo.site/" "<%= wpTitle %>" <% if ( multisite ) { %> <%= siteSlug %>.vipdev.lndo.site <% } %>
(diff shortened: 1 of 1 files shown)


assets/dev-env.lando.template.yml.ejs
```diff
@@ -39,13 +41,19 @@ services:
       environment:
         XDEBUG: <%= xdebug ? 'enable' : 'disable' %>
         STATSD: <%= statsd ? 'enable' : 'disable' %>
+        LANDO_NO_USER_PERMS: "We have our own user mapping script under /scripts"
```
rinatkhaziev: Let's change that to the same format for consistency ("enable"/"disable")

pschoffer: I am not sure how to do the double negative :) How's this?

```suggestion
        LANDO_NO_USER_PERMS: 'enable'
```

rinatkhaziev: > I am not sure how to do the double negative :) How's this?
> 
> 
> 
> ```suggestion
> 
>         LANDO_NO_USER_PERMS: 'enable'
> 
> ```

Yeah, I was typing fast, sorry for the nonsense :)

rinatkhaziev (approved):
🚢

rinatkhaziev:
A test is failing but it seems unrelated?

PR #913, opened 2021-11-04
Update/no login dev env

pschoffer:
## Description

`dev-env` subcommands don't communicate with parker most of the time. The only exception is when `@app.env` arguments are provided we prefetch some info about env.

This change stops requiring login to use the other parts of `dev-env` subcommands. This was recently requested to support customers-to-be that dont yet have an account to log in to.

## Steps to Test

1) Logout `vip logout`
2) Try some dev-env commands without app `vip dev-env create` / `vip dev-env info` .... should work without login
3) Try some dev-env commands with app `vip @2891 create` / `vip dev-env @2891 info` ... should ask you to login

### DIFF
--- a/__tests__/lib/cli/command.js
+++ b/__tests__/lib/cli/command.js
@@ -1,0 +1,35 @@
+
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import { containsAppEnvArgument } from 'lib/cli/command';
+
+describe( 'utils/cli/command', () => {
+	describe( 'containsAppEnvArgument', () => {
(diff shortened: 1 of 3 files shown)


brunobasto (approved):
Tested and LGTM:
```
./dist/bin/vip.js logout
You are successfully logged out.

./dist/bin/vip-dev-env-create.js

✖ WordPress site title · VIP Dev

./dist/bin/vip-dev-env-create.js @2981
Unauthorized: You are unauthorized to perform this request, please logout with `vip logout` then try again.
```

Thanks @pschoffer

PR #914, opened 2021-11-04
[dev-env] update help wording for dev env

pschoffer:
## Description

Updates some help text:
* Wordpress can no longer be provided as a local path
* Default values don't make sense as the wizzard is triggered when not provided.

## Steps to Test
`vip dev-env create --help`


josejaureguiberry (approved):
🚀

PR #915, opened 2021-11-05
Add warning message when an envvar is set/deleted

vaurdan:
## Description

Add a warning message with a reminder that changes to the environment variables will only be applied upon an application deployment.

If `--skip-confirmation` is passed, the message will not be displayed.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `npm link`
1. Run `vip @1234.production config envvar set TESTING_VAR` (replace `@1234.production` with your environment).
2. You should see `Important: Updates to environment variables will not be available until the application’s next deploy.` after the success message.
3. Run `vip @1234.production config envvar delete TESTING_VAR` and check for the same message.
4. Try the previous commands with the `--skip-confirmation` flag and you shouldn't see the `Important` message.


chriszarate (approved):
Very nice!

PR #916, opened 2021-11-05
[dev-env] Save instance data state

pschoffer:
## Description

Save instance data string to a file on creation.

This will enable us to re-trigger the setup wizard in order to reconfigure the existing environment.

## Steps to Test

1) create new environment.
1) In the location you can find - `instance_data.json`

### DIFF
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -92,7 +92,6 @@ command()
 			statsd: opt.statsd || false,
 			phpmyadmin: opt.phpmyadmin || false,
 			xdebug: opt.xdebug || false,
-			version: 1.0,
 		};
 
 		try {
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -92,6 +92,7 @@ command()
 			statsd: opt.statsd || false,
 			phpmyadmin: opt.phpmyadmin || false,
 			xdebug: opt.xdebug || false,
+			version: 1.0,
 		};
 
 		try {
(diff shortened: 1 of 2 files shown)


pauarge:
Why are we adding a hardcoded version number? What does it mean?

pschoffer:
Good question, the idea there is that if in future releases we would switch some format or structure of data stored we can still kinda detect that. 

Say as an example right now we store `{ phpmyAdmin: true, xdebug:false}` in the future we would store `{extraServices: [ 'phpmyadmin']}` with the version we will be able to know how to parse the file correctly.

I am however open to ideas ... maybe it is unnecessary.

pauarge:
@pschoffer That sounds reasonable and it makes sense. However, we don't have that versioning problem yet, so I think it's a bit overkill. Whenever we introduce a breaking change, we can start at version 2 and assume that if a config file doesn't have an explicit version number, it's version 1.

src/bin/vip-dev-env-create.js
```diff
@@ -92,6 +92,7 @@ command()
 			statsd: opt.statsd || false,
 			phpmyadmin: opt.phpmyadmin || false,
 			xdebug: opt.xdebug || false,
+			version: 1.0,
```
pschoffer:
```suggestion
```

pschoffer:
Sounds good I dropped it. Thanks @pauarge

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -91,6 +92,7 @@ type NewInstanceData = {
 	muPlugins: Object,
 	clientCode: Object,
 	mediaRedirectDomain: string,
+	version: number,
```
pauarge: There's a version leftover here

pauarge (approved):

PR #921, opened 2021-11-10
[dev-env] Introuces update to change existing environment

pschoffer:
## Description

This command is usefull when want to tweak existing environment. Common scenarios are:
* Enable/disable phpmyadmin/statsd/xdebug
* Switch wp version

## Steps to Test

1) Create environment
2) Run `vip dev-env update`
3) Make sure it does what you would expect it to do and correct default values are fetched from existing env

### DIFF
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -18,6 +18,7 @@ import command from 'lib/cli/command';
 import { getEnvironmentName } from 'lib/dev-environment/dev-environment-cli';
 import { DEV_ENVIRONMENT_FULL_COMMAND } from 'lib/constants/dev-environment';
 import { addDevEnvConfigurationOptions, handleCLIException, promptForArguments } from '../lib/dev-environment/dev-environment-cli';
+import type { InstanceOptions } from '../lib/dev-environment/dev-environment-cli';
 import { doesEnvironmentExist, readEnvironmentData } from '../lib/dev-environment/dev-environment-core';
 import { DEV_ENVIRONMENT_NOT_FOUND } from '../lib/constants/dev-environment';
 
@@ -56,7 +57,16 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 
 		debug( 'Combined preselected data are', preselectedOptions );
 
-		const instanceData = await promptForArguments( preselectedOptions, {} );
+		const defaultOptions: InstanceOptions = {
+			clientCode: currentInstanceData.clientCode.dir || currentInstanceData.clientCode.tag,
+			muPlugins: currentInstanceData.muPlugins.dir || currentInstanceData.muPlugins.tag,
+			wordpress: currentInstanceData.wordpress.tag,
+			elasticsearch: currentInstanceData.elasticsearch,
+			mariadb: currentInstanceData.mariadb,
+
+		}
+
+		const instanceData = await promptForArguments( preselectedOptions, currentInstanceData );
 		// const instanceDataWithSlug = {
 		// 	...currentInstanceData,
 		// 	siteSlug: slug,
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -63,7 +63,7 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			mariadb: currentInstanceData.mariadb,
 			statsd: currentInstanceData.statsd,
 			phpmyadmin: currentInstanceData.phpmyadmin,
-			xdebug: currentInstanceData.phpmyadmin,
+			xdebug: currentInstanceData.xdebug,
 			mediaRedirectDomain: currentInstanceData.mediaRedirectDomain,
 		};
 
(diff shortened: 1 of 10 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -308,7 +341,7 @@ export async function getApplicationInformation( appId: number, envType: string
 	return appData;
 }
 
-export async function resolveImportPath( slug: string, fileName: string, searchReplace: string, inPlace: boolean ): Promise<SQLImportPaths> {
+export async function resolveImportPath( slug: string, fileName: string, searchReplace: String | String[], inPlace: boolean ): Promise<SQLImportPaths> {
```
pauarge: `string` should be lowercase

pschoffer: lowercase causes flow issue here: https://github.com/Automattic/vip/blob/fbb6502fea2304bdd69dfc97623cb41fd2d064be/src/lib/search-and-replace.js#L128

But maybe we should change type there instead of here.

pschoffer: Fixed it at the source ...

src/bin/vip-import-sql.js
```diff
@@ -506,6 +506,7 @@ Processing the SQL import for your environment...
 				environmentId: env.id,
 				basename: basename,
 				md5: md5,
+				searchReplace: [],
```
pauarge, line 509: Are this changes related?

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -270,52 +265,76 @@ export async function promptForComponent( component: string, allowLocal: boolean
...
-			choices,
+			choices: modChoices,
 			initial: initialModeIndex,
 		} );
 
 		modeResult = await select.run();
 	}
 
 	const messagePrefix = selectMode ? '\t' : `${ componentDisplayName } - `;
+	// path
```
pauarge: Leftover?

src/lib/dev-environment/types.js
```diff
@@ -0,0 +1,54 @@
+
+export interface InstanceOptions {
+	title?: string,
+	multisite?: boolean,
```
pauarge: Aren't title and multisite always defined?

src/lib/dev-environment/types.js
```diff
@@ -0,0 +1,54 @@
+
```
pauarge, line 1: General comment on this type definitions, which cascade all over the PR -- could we try to have less nullable types? It would save us a lot of `x?.something || ' '`.

src/bin/vip-import-sql.js
pschoffer: Not really, they just break flow validations.

src/lib/dev-environment/dev-environment-cli.js
pschoffer:
```suggestion
```

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -108,12 +100,41 @@ export async function createEnvironment( instanceData: NewInstanceData ) {
...
+		throw new Error( 'Environment doesn\'t exist.' );
+	}
+
+	const preProcessedInstanceData = preProcessInstanceData( instanceData );
+
+	await prepareLandoEnv( preProcessedInstanceData, instancePath );
+}
+
+function preProcessInstanceData( instanceData: InstanceData ): InstanceData {
+	const newInstanceData = {
```
pauarge, line 128: Does it have any side effects to modify the original object instead of having to create a new one?

pschoffer: If you do:
```
const a = {};
const b = preprocess(a)
```
You don't expect a to be changed as well. [Pure functions ](https://en.wikipedia.org/wiki/Pure_function) are kinda easy to think about I feel like.

src/bin/vip-dev-env-update.js
```diff
@@ -0,0 +1,85 @@
...
+
+		const defaultOptions: InstanceOptions = {
+			clientCode: currentInstanceData.clientCode.dir || currentInstanceData.clientCode.tag || 'latest',
+			muPlugins: currentInstanceData.muPlugins.dir || currentInstanceData.muPlugins.tag || 'latest',
+			wordpress: currentInstanceData.wordpress.tag,
+			elasticsearch: currentInstanceData.elasticsearch,
+			mariadb: currentInstanceData.mariadb,
+			statsd: currentInstanceData.statsd,
+			phpmyadmin: currentInstanceData.phpmyadmin,
+			xdebug: currentInstanceData.phpmyadmin,
```
pauarge: Looks like this should be `.xdebug`

src/lib/dev-environment/dev-environment-core.js
pauarge: Agreed, but copying an object over and over has an effect on performance. My question comes from the fact that if what we're doing here is taking an object and putting it through a series of functions that modify it (and those functions are only used here), it's fine to change the object that is passed by reference.

This is a more philosophical comment, so not worth blocking the PR because of it.

src/bin/vip-dev-env-update.js
```diff
@@ -0,0 +1,85 @@
...
+		debug( 'Read instance data', currentInstanceData );
+
+		const preselectedOptions = {
+			// Title and multisite can't be changed during update
+			title: currentInstanceData.wpTitle,
+			multisite: currentInstanceData.multisite,
+			...opt,
+		};
+
+		const defaultOptions: InstanceOptions = {
```
pauarge, line 58: Maybe we could extract the default options (plus some validation that's happening here) outside the command itself.

pauarge (approved):

PR #926, opened 2021-11-18
Add the `vip logs` command to get the most recent logs from a site

robersongomes:
## Description

Add the `vip logs` command to get the most recent logs from a site.

#### Before the API is available in production, the easiest way to test this PR is by pointing to the local API and returning fake logs in GOOP API:
1. Check out the API branch `add/get-runtime-logs` and execute `docker-compose up`
1. On GOOP, replace this line `const logs = await getRecentLogs( site, tier, limitNumber );` in the endpoint implementation with something like this:
```javascript
const logs = Array( limitNumber ).fill( 0 ).map( ( a, index ) => {
	return {
		message: `My ${ tier } log message ${ index + 1 }`,
		timestamp: new Date().toISOString().replace( 'Z', '000000Z' ),
		pod: 'my-pod-123456',
	};
} );
```
then...
1. Check out this PR.
1. Run `npm run build`
1. Run `npm link`
1. Run `VIP_PROXY="" API_HOST=http://localhost:4000 vip @5000.production logs --type app --limit 10`
1. Verify the output is in the format: `2021-11-09T20:47:07.747765934Z My log message`.
1. Verify the logs are limited to the specified `--limit`

#### After the API is available in production:

1. Check out PR.
1. Run `npm run build`
1. Run `npm link`
1. Run `vip @[app-id].[env] logs --type app --limit 10`
1. Verify the logs for the specified type are displayed
1. Verify the output is in the format: `2021-11-09T20:47:07.747765934Z My log message`.
1. Verify the logs are limited to the specified `--limit`

### DIFF
--- a/src/bin/vip-logs.js
+++ b/src/bin/vip-logs.js
@@ -84,15 +84,15 @@ command( {
 	.option( 'limit', 'The maximum number of log lines', 500 )
 	.examples( [
 		{
-			usage: 'vip logs --app 9999 --env develop',
+			usage: 'vip @mysite.production logs',
 			description: 'Get the most recent app logs',
 		},
 		{
-			usage: 'vip logs --app 9999 --env develop --type batch',
+			usage: 'vip @mysite.production logs --type batch',
 			description: 'Get the most recent batch logs',
 		},
 		{
-			usage: 'vip logs --app 9999 --env develop --limit 100',
+			usage: 'vip @mysite.production logs --limit 100',
 			description: 'Get the most recent 100 logs',
 		},
 	] )
--- a/src/bin/vip-logs.js
+++ b/src/bin/vip-logs.js
@@ -14,7 +14,7 @@ const LIMIT_MIN = 1;
 const ALLOWED_TYPES = [ 'app', 'batch' ];
 
 export async function getLogs( arg: string[], opt ): Promise<void> {
-	validateInputs( opt.type, opt.limit );
+	validateInputs( opt.type, opt.limit, opt.env );
 
 	const trackingParams = {
 		command: 'vip logs',
@@ -44,7 +44,11 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
 	await trackEvent( 'logs_command_success', trackingParams );
 }
 
-export function validateInputs( type: string, limit: number ): void {
+export function validateInputs( type: string, limit: number, env: Object ): void {
+	if ( ! env.isK8sResident ) {
+		exit.withError( '`vip logs` is not supported for the specified environment.' );
+	}
+
 	if ( ! ALLOWED_TYPES.includes( type ) ) {
 		exit.withError( `Invalid type: ${ type }. The supported types are: ${ ALLOWED_TYPES.join( ', ' ) }.` );
 	}
@@ -62,6 +66,7 @@ export const appQuery = `
 		appId
 		name
 		type
+		isK8sResident
 	}
 	organization {
 		id
--- a/src/bin/vip-logs.js
+++ b/src/bin/vip-logs.js
@@ -1,0 +1,94 @@
+#!/usr/bin/env node
+// @flow
+/**
+ * Internal dependencies
+ */
+import command from 'lib/cli/command';
+import { rollbar } from 'lib/rollbar';
+import { trackEvent } from 'lib/tracker';
+import * as logsLib from 'lib/logs/logs';
+import * as exit from 'lib/cli/exit';
+
+const LIMIT_MAX = 5000;
+const LIMIT_MIN = 1;
+const ALLOWED_TYPES = [ 'app', 'batch' ];
+
+export async function getLogs( arg: string[], opt ): Promise<void> {
+	validateInputs( opt.type, opt.limit );
+
+	const trackingParams = {
+		command: 'vip logs',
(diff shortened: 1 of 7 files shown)


src/lib/logs/logs.js
```diff
@@ -0,0 +1,52 @@
...
+ * Internal dependencies
+ */
+import API from 'lib/api';
+
+const QUERY_ENVIRONMENT_LOGS = gql`
+	query App( $appId: Int, $envId: Int ) {
+		app( id: $appId ) {
+			environments( id: $envId ) {
+				id
+				logs( type: String, limit: Int ) {
```
netsuso: For it to work, I had to change this to:

```
logs( type: $type, limit: $limit ) {
```

And also the query definition:

```
query App( $appId: Int, $envId: Int, $type: String, $limit: Int ) {
```

(By the way, maybe we could call the query "GetAppLogs", or something like that, so if it gets logged anywhere it's more informative)

robersongomes: Thank you for looking into that! I was using the current Parker implementation for testing, which doesn't have the filters yet. I will fix that!

robersongomes:
The tests are still failing for `current-stretch` and `lts-stretch`, but I think it might not be related to the code, so I'm moving this to "needs review".

src/bin/vip-logs.js
```diff
@@ -0,0 +1,99 @@
...
+		{
+			usage: 'vip logs --app 9999 --env develop',
+			description: 'Get the most recent app logs',
+		},
+		{
+			usage: 'vip logs --app 9999 --env develop --type batch',
+			description: 'Get the most recent batch logs',
+		},
+		{
+			usage: 'vip logs --app 9999 --env develop --limit 100',
```
nickdaugherty: In the examples, we should be using the `@<app>.<env>` style like other commands (we seem to have really inconsistent command examples already 😢 )

https://github.com/Automattic/vip/blob/6927676cae70de83fba2deebc1ad5e6119fbdff7/src/bin/vip-import-sql.js#L176-L209
https://github.com/Automattic/vip/blob/3e962140bf3e6f045912965debda56a7d11a3dde/src/bin/vip-import-media.js#L57-L75

nickdaugherty: Even these are not "right" - it should be `vip @mysite.production logs .......`

nickdaugherty (commented):
This looks good @robersongomes, I left a note on the command examples.

The testing steps don't work b/c the production API PR isn't merged. How can I test this?

```
$ ./dist/bin/vip-logs.js @297.develop
Error: Unknown argument "type" on field "logs" of type "AppEnvironment".
Error: Unknown argument "limit" on field "logs" of type "AppEnvironment".
```

src/bin/vip-logs.js
robersongomes: Thank you for catching that. Fixed on a947ccd9b00630d487cd384700b2baa579a507ac

robersongomes:
Thanks for the feedback @nickdaugherty! I updated the testing steps to cover both cases, before and after the API is available on production.

netsuso:
> The testing steps don't work b/c the production API PR isn't merged. How can I test this?

I'll merge the Parker branch today, so we'll be able to test this much more easily :)

src/bin/vip-logs.js
nickdaugherty: Perfect!

robersongomes:
The builds are finally successful after doing this https://github.com/Automattic/vip/pull/936 :tada: :sweat_smile:

src/bin/vip-logs.js
```diff
@@ -0,0 +1,99 @@
...
+		await trackEvent( 'logs_command_error', { ...trackingParams, error: error.message } );
+
+		return exit.withError( error.message );
+	}
+
+	await trackEvent( 'logs_command_success', trackingParams );
+}
+
+export function validateInputs( type: string, limit: number, env: Object ): void {
+	if ( ! env.isK8sResident ) {
```
chriszarate, line 52: This isn't an input and seems like it should just be validated on the Parker side. This would eliminate the need for to query twice

chriszarate:
This is great / life-changing! :)

I'm wondering if we should target CSV or at least provide that option (via `format: 'csv'` passed to the command options). Messages containing newlines really disrupt the output format and would be fairly impossible to parse / consume programmatically.

src/bin/vip-logs.js
mjangda: That makes sense 👍 

We'll need to update the error message on the GOOP side since it references k8s.

robersongomes: @chriszarate I considered the provided `@app.environment` as an input for the logs command (maybe it causes some confusion... any suggestions?) because I used the common `command` function:
https://github.com/Automattic/vip/blob/a947ccd9b00630d487cd384700b2baa579a507ac/src/bin/vip-logs.js#L77-L82
This function queries the app and environment before starting the `vip logs` logic.

I totally agree that we should have this check on the API side. This was just a check to avoid another unnecessary query to the API.

mjangda:
> I'm wondering if we should target CSV or at least provide that option

Good idea 👍  -- tracking in CAFE-312.

mjangda (approved):
Tested and working well 🚢

PR #927, opened 2021-11-19
[dev-env] Handles user already exists during sql import

pschoffer:
## Description

After SQL import we try to provision the `vipgo` user.

However, that fails if the vipgo user is already present.

```
$ vip dev-env import sql --slug test ~/test.sql 
Success: Imported from '/user/test.sql'.
Error: The 'vipgo' username is already registered.
Error: Error: The 'vipgo' username is already registered.
```

I couldn't find a way to fully suppress the error as that is printed by `wp` cli . I tried to capture output of `wp user list` and to check if the user exists before attempting to create it, but no luck there either. Because the `wp-cli` runs inside a container and passing those `stdio` doesn't work as expected.

We could alternatively create a different CLI, but seems like overdoing it.

So I just removed one error and print extra success line:
```
Success: Imported from '/user/test.sql'.
Error: The 'vipgo' username is already registered.
Success: Skipping user vipgo provisioning
```
 

## Steps to Test

1) `vip dev-env --slug test exec  -- wp db export - > ~/test.sql`
1) `vip dev-env --slug test import sql ~/test.sql`

### DIFF
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -9,6 +9,7 @@
  * External dependencies
  */
 import fs from 'fs';
+import chalk from 'chalk';
 
 /**
  * Internal dependencies
@@ -17,7 +18,6 @@ import command from '../lib/cli/command';
 import { getEnvironmentName, handleCLIException } from '../lib/dev-environment/dev-environment-cli';
 import { exec, resolveImportPath } from '../lib/dev-environment/dev-environment-core';
 import { DEV_ENVIRONMENT_FULL_COMMAND } from '../lib/constants/dev-environment';
-import chalk from 'chalk';
 
 const examples = [
 	{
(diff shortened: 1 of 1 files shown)


src/bin/vip-dev-env-import-sql.js
```diff
@@ -17,6 +17,7 @@ import command from '../lib/cli/command';
 import { getEnvironmentName, handleCLIException } from '../lib/dev-environment/dev-environment-cli';
 import { exec, resolveImportPath } from '../lib/dev-environment/dev-environment-core';
 import { DEV_ENVIRONMENT_FULL_COMMAND } from '../lib/constants/dev-environment';
+import chalk from 'chalk';
```
pauarge: This should be under external dependencies.

pauarge (changes requested):
Small change, but looks good otherwise.

src/bin/vip-dev-env-import-sql.js
pschoffer: Fixed. Thank you!

pauarge (approved):
LGTM

PR #939, opened 2021-12-03
Rename 'logs' library to 'app-logs' to avoid publish-please to fail

netsuso:
## Description

We have recently added a new module `lib/logs/logs.js`. But when we try to publish it using `publish-please`, the path is checked against a hardcoded list of sensitive file names in this tool (it tries to avoid actual logs being published). While we could create our own copy of the `.sensitivedata` file and adapt it to our needs, it seems renaming that module means less maintentance work for the future.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-logs.js`
1. Ensure it keeps working as before
1. Run `npm run publish-please`
1. Ensure the sensitive data step is working (even if the publish process fails because it's not the expected branch)


elazzabi (approved):
Looks good! :shipit:

PR #944, opened 2021-12-13
Add `vip whoami` command

mjangda:
To help identify the currently logged in user. This can help with debugging login issues and during development.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-whoami.js`

```
❯ ./dist/bin/vip-whoami.js
- Howdy VIP User!
- Your user ID is 12345
```

For VIP users, there is an additional note: `- Your account has VIP Staff permissions`.

### DIFF
--- a/src/lib/api/user.js
+++ b/src/lib/api/user.js
@@ -1,0 +1,43 @@
+/**
+ * @flow
+ * @format
+ */
+
+/**
+ * External dependencies
+ */
+import gql from 'graphql-tag';
+
+/**
+ * Internal dependencies
+ */
+import API from 'lib/api';
+
+const QUERY_CURRENT_USER = gql`
+	query Me {
+		me {
+			id
+			displayName
+			isVIP
+			organizationRoles {
+				nodes {
+					organizationId
+					roleId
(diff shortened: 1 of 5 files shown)


src/lib/api/user.js
```diff
@@ -0,0 +1,48 @@
...
+
+export async function getCurrentUserInfo() {
+	const api = await API();
+
+	let response = {};
+	try {
+		response = await api.query( { query: QUERY_CURRENT_USER } );
+	} catch ( err ) {
+		throw err;
+	}
```
nickdaugherty: Since we're not doing anything with the error (except re-throwing), we could omit the try/catch.

nickdaugherty (approved):
This is awesome! I left a non-blocking comment about a try/catch, but 🚢

src/lib/api/user.js
mjangda: Good idea; ff0a619

simonwheatley:
Thank you for building a feature just for me, @mjangda 🥰

PR #945, opened 2021-12-17
Fallback to the first element in the array of environments when fetching the app info for dev env wizard and no environment type is passed

rinatkhaziev:
## Description

Currently, if a `@ID `notation is used (without the environment type is used) we don't fetch the needed information.  We should match the behavior where omitting the env fetched the production info

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Try `vip @1337 dev-env create`
1. Verify the correct properties are picked for the environment


pschoffer:
Hmm, the size one check was there to be consistent with other subcommands. The vip cli used to prompt for environment if there is more than one in that app. I wonder if that changed or it doesn't prompt always. 

On the other hand for our purposes, I guess it is fine to fall back to the first one anyways if not specified. What do you think @rinatkhaziev ?

rinatkhaziev:
> Hmm, the size one check was there to be consistent with other subcommands. The vip cli used to prompt for environment if there is more than one in that app. I wonder if that changed or it doesn't prompt always.

It does prompt, but for not dev-env, for dev-env it just doesn't populate the data (hence this PR)

Ideally, we want to match the behavior of the rest of vip subcommands, but given that production is always the first element in the list I think it's a safe fallback.

rinatkhaziev:
@pschoffer ^^ forgot to `@` you in the above reply.

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -323,7 +323,7 @@ export async function getApplicationInformation( appId: number, envType: string
 		let envData;
 		if ( envType ) {
 			envData = environments.find( candidateEnv => candidateEnv.type === envType );
-		} else if ( 1 === environments.length ) {
+		} else if ( 1 === environments.length || ! envType ) {
 			envData = environments[ 0 ];
```
jesse-greathouse, line 336: Im not sure how this will ever be truthy since the first condition in this else/if tests envType for truthy. Based on the code this seems like an irrelevant check. Maybe Im just not understanding, but it seems like it would be better to flip the order of how it checks conditions so it's checking for a 1 member list first.

e.g....

```php
if ( 1 === environments.length ) {
  envData = environments[ 0 ]
} else if ( envType ) {
    envData = environments.find( candidateEnv => candidateEnv.type === envType );
} else {
```

jesse-greathouse (approved):
Had a small  suggestion/question about the logic. Overall I think this is a good change.

rinatkhaziev:
So I think this was fixed and now it presents with a choice to pick the env, closing it.

PR #946, opened 2021-12-23
Add support for the site logs tailing feature

netsuso:
## Description

This adds a new `--follow` parameter to the `logs` command that uses the new Graphql parameter to fetch logs only from a specific timestamp. This way, `vip-cli` will poll the API to fetch only the new logs since the previous request.

The API functionality is still not integrated, so this needs a full local dev environment

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-logs.js @<site>.<env> --follow`
1. This will run forever, fetching new logs every 30 seconds

### DIFF
--- a/__tests__/bin/vip-logs.js
+++ b/__tests__/bin/vip-logs.js
@@ -28,6 +28,7 @@ jest.mock( 'lib/tracker', () => ( {
 } ) );
 
 jest.mock( 'lib/app-logs/app-logs', () => ( {
+	// Only mock what is really needed, otherwise exported constants like LIMIT_MAX would be `undefined` during the tests
 	...jest.requireActual( 'lib/app-logs/app-logs' ),
 	getRecentLogs: jest.fn(),
 } ) );
--- a/__tests__/bin/vip-logs.js
+++ b/__tests__/bin/vip-logs.js
@@ -28,6 +28,7 @@ jest.mock( 'lib/tracker', () => ( {
 } ) );
 
 jest.mock( 'lib/app-logs/app-logs', () => ( {
+	...jest.requireActual( 'lib/app-logs/app-logs' ),
 	getRecentLogs: jest.fn(),
 } ) );
 
@@ -56,10 +57,13 @@ describe( 'getLogs', () => {
 	beforeEach( jest.clearAllMocks );
 
 	it( 'should display the logs in the output', async () => {
-		logsLib.getRecentLogs.mockImplementation( async () => [
-			{ timestamp: '2021-11-05T20:18:36.234041811Z', message: 'My container message 1' },
-			{ timestamp: '2021-11-09T20:47:07.301221112Z', message: 'My container message 2' },
-		] );
+		logsLib.getRecentLogs.mockImplementation( async () => ( {
+			nextCursor: null,
+			nodes: [
+				{ timestamp: '2021-11-05T20:18:36.234041811Z', message: 'My container message 1' },
+				{ timestamp: '2021-11-09T20:47:07.301221112Z', message: 'My container message 2' },
+			],
+		} ) );
 
 		await getLogs( [], opts );
 
@@ -78,6 +82,7 @@ describe( 'getLogs', () => {
 			env_id: 3,
 			type: 'app',
 			limit: 500,
+			follow: false,
 			format: 'text',
 		};
 
@@ -85,7 +90,7 @@ describe( 'getLogs', () => {
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'logs_command_execute', trackingParams );
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'logs_command_success', {
 			...trackingParams,
-			logs_output: 2,
+			total: 2,
 		} );
 
 		expect( rollbar.error ).not.toHaveBeenCalled();
@@ -94,10 +99,13 @@ describe( 'getLogs', () => {
 	it( 'should display the logs in the output with JSON format', async () => {
 		opts.format = 'json';
 
-		logsLib.getRecentLogs.mockImplementation( async () => [
-			{ timestamp: '2021-11-05T20:18:36.234041811Z', message: 'My container message 1' },
-			{ timestamp: '2021-11-09T20:47:07.301221112Z', message: 'My container message 2' },
-		] );
+		logsLib.getRecentLogs.mockImplementation( async () => ( {
+			nextCursor: null,
+			nodes: [
+				{ timestamp: '2021-11-05T20:18:36.234041811Z', message: 'My container message 1' },
+				{ timestamp: '2021-11-09T20:47:07.301221112Z', message: 'My container message 2' },
+			],
+		} ) );
 
 		await getLogs( [], opts );
 
@@ -124,6 +132,7 @@ describe( 'getLogs', () => {
 			env_id: 3,
 			type: 'app',
 			limit: 500,
+			follow: false,
 			format: 'json',
 		};
 
@@ -131,7 +140,7 @@ describe( 'getLogs', () => {
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'logs_command_execute', trackingParams );
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'logs_command_success', {
 			...trackingParams,
-			logs_output: 2,
+			total: 2,
 		} );
 
 		expect( rollbar.error ).not.toHaveBeenCalled();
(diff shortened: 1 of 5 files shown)


src/bin/vip-logs.js
```diff
@@ -51,6 +56,33 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+
+export async function followLogs( opt ): Promise<void> {
+	let interval = 30;
+
+	let since;
+
+	while ( true ) {
+		let logs = [];
+		try {
+			logs = await logsLib.getRecentLogs( opt.app.id, opt.env.id, opt.type, opt.limit, since );
```
robersongomes: It's probably better to pass `5000` as the limit for the subsequent requests.

robersongomes: done in 8177be3

__tests__/bin/vip-logs.js
```diff
@@ -28,6 +28,7 @@ jest.mock( 'lib/tracker', () => ( {
 } ) );
 
 jest.mock( 'lib/app-logs/app-logs', () => ( {
+	...jest.requireActual( 'lib/app-logs/app-logs' ),
```
mjangda, line 32: Can we add in inline comment noting why we're doing this? I'm guessing this is because we just want to mock `getRecentLogs` but it's not super obvious.

src/bin/vip-logs.js
```diff
@@ -43,14 +48,47 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+	let isFirstRequest = true;
+
+	while ( true ) {
+		const limit = isFirstRequest ? opt.limit : 5000;
+
+		let logs;
+		try {
+			logs = await logsLib.getRecentLogs( opt.app.id, opt.env.id, opt.type, limit, after );
+		} catch ( error ) {
+			console.error( 'Failed to fetch logs. Trying again after the polling interval' );
```
mjangda: Do we want to have some sort of exponential backoff as well as a failure limit? If the API is down for an extended period (or if the computer has network issues), it would be better to have any following clients ease off and automatically stop rather than polling forever.

mjangda: Can we mention the polling interval here?

```
Failed to fetch logs. Trying again in ${ delay } second(s)
```

mjangda (commented):
Looks pretty good. Left some suggestions.

__tests__/bin/vip-logs.js
robersongomes: I added this because the constant `LIMIT_MAX` defined in the module would end up being `undefined` during the test execution. 
Comment added in 6ae2a73ce4f93b8111537a2344adfcaaecc8b57b

src/bin/vip-logs.js
netsuso: Done! I've refactored the delay handling, so it informs the user when it's going to retry, we have a backoff (not exponential, but we don't want to wait too much in case of a sequence of errors, I think). I've also changed the API call method so it doesn't immediately exit the process for the logs command, as it was preventing retries on certain errors (such a network error)

mjangda (commented):
Looking good! I pushed two commits to clean up some constants and also added Tracks events for the follow stuff. If there's no issues with those two changes, I think this is good to go.

src/bin/vip-logs.js
```diff
@@ -43,14 +42,89 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+	const trackingParams = getBaseTrackingParams( opt );
+
+	// Set an initial default delay
+	let delay = DEFAULT_POLLING_DELAY_IN_SECONDS;
+
+	while ( true ) {
+		const limit = isFirstRequest ? opt.limit : LIMIT_MAX;
+
+		requestNumber++;
+		trackingParams.request_number = requestNumber;
```
robersongomes, line 71: [MINOR] I'm wondering if it's worth adding a unique identifier (uuid maybe) in case we want to easily filter a sequence of calls.

src/lib/api.js
```diff
@@ -31,7 +31,7 @@ export function disableGlobalGraphQLErrorHandling() {
 	globalGraphQLErrorHandlingEnabled = false;
 }
 
-export default async function API(): Promise<ApolloClient> {
+export default async function API( exitOnError = true ): Promise<ApolloClient> {
```
robersongomes: This would be more readable (in the usages) if we transform it into an options object.
```javascript
const api = await API( { exitOnError: false } );
```
it's much easier to understand than:
```javascript
const api = await API( false );
```

```suggestion
export default async function API( { exitOnError = true } = {} ): Promise<ApolloClient> {
```

src/lib/app-logs/app-logs.js
```diff
@@ -13,24 +13,34 @@ import gql from 'graphql-tag';
...
-export async function getRecentLogs( appId: number, envId: number, type: string, limit: number ): Promise<Array<{ timestamp: string, message: string }>> {
-	const api = await API();
+type GetRecentLogsResponse = {
+	nodes: Array<{ timestamp: string, message: string }>,
+	nextCursor: string,
+	pollingDelaySeconds: number,
+};
+
+export async function getRecentLogs( appId: number, envId: number, type: string, limit: number, after?: string ): Promise<GetRecentLogsResponse> {
+	const api = await API( false );
```
robersongomes: In case we change the param to an options object:
```suggestion
	const api = await API( { exitOnError: false } );
```

src/bin/vip-logs.js
```diff
@@ -43,14 +42,89 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+			await trackEvent( 'logs_command_follow_error', { ...trackingParams, error: error.message } );
+
+			// If the first request fails we don't want to retry (it's probably not recoverable)
+			if ( isFirstRequest ) {
+				console.error( 'Error fetching initial logs' );
+				break;
+			}
+			// Increase the delay on errors to avoid overloading the server, up to a max of 5 minutes
+			delay += DEFAULT_POLLING_DELAY_IN_SECONDS;
+			delay = Math.min( delay, MAX_POLLING_DELAY_IN_SECONDS );
```
robersongomes, line 93: Awesome! :clap:

src/bin/vip-logs.js
```diff
@@ -43,14 +42,89 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+			await trackEvent( 'logs_command_follow_success', {
+				...trackingParams,
+				total: logs?.nodes.length,
+			} );
+		} catch ( error ) {
+			await trackEvent( 'logs_command_follow_error', { ...trackingParams, error: error.message } );
+
+			// If the first request fails we don't want to retry (it's probably not recoverable)
+			if ( isFirstRequest ) {
+				console.error( 'Error fetching initial logs' );
```
robersongomes, line 88: [1] Shouldn't we print the error message? If the user is hitting a rate limit for example, they will never know.
```suggestion
				console.error( `Error fetching initial logs: ${ error.message }` );
```

src/bin/vip-logs.js
```diff
@@ -43,14 +42,89 @@ export async function getLogs( arg: string[], opt ): Promise<void> {
...
+				break;
+			}
+			// Increase the delay on errors to avoid overloading the server, up to a max of 5 minutes
+			delay += DEFAULT_POLLING_DELAY_IN_SECONDS;
+			delay = Math.min( delay, MAX_POLLING_DELAY_IN_SECONDS );
+			console.error( `Failed to fetch logs. Trying again in ${ delay } seconds` );
```
robersongomes, line 94: [2] Shouldn't we print the error message? If the user is hitting a rate limit for example, they will never know.
```suggestion
			console.error( `Failed to fetch logs: ${ error.message }` );
			console.error( `Trying again in ${ delay } seconds` );
```

robersongomes (approved):
This is awesome!
I only left some minor suggestions.

src/lib/api.js
netsuso: Sounds good, I'll change both! Thanks!

src/bin/vip-logs.js
netsuso: I'm not sure about this... In the end this will always be an error coming from the Apollo client, and it manages errors with the `onError` handler. And we are already printing there all the errors of the GraphQL request. So this `console.error()` (and the one below) should always be printed after the actual errors. I'll try it anyway, forcing some errors, to see what we would obtain here.

robersongomes: oh, if the errors are already being printed, please just ignore this suggestion :upside_down_face:

netsuso: Yep, I confirm we are printing at least the same as `error.message` (and more if we have multiple errors), so this is not needed 😊  But thanks for the suggestion anyway! It made sense without knowing the context

src/bin/vip-logs.js
netsuso: Skipping this as per the explanation in the other comment

src/bin/vip-logs.js
netsuso: Sounds interesting... Not sure if this kind of filtering can be easily done in the tracking functionality, maybe @mjangda  knows better.

In any case, I think I'm merging this now so I can release the testing version, we may include this change in another PR

PR #952, opened 2022-01-14
FORNO-1047: Fix SQL Import for compressed files

nejclovrencic:
## Description

Due to a bug on the backend, compressed files currently cannot be imported as the md5 hash calculation fails. This PR moves md5 calculation to take place just before the file is uploaded.

In addition, this PR skips validation and search replace when customers try to import compressed SQL file.

## Steps to Test

1. Check out PR.
2. Import compressed file to site 4361 (contains fixed jobs image): `vip import sql @4361.production data.sql.gz`
3. Import uncompressed file to site 4361 (contains fixed jobs image): `vip import sql @4361.production data.sql`
4. Verify that both imports work as expected

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -10595,7 +10595,7 @@
         },
         "onetime": {
           "version": "1.1.0",
-          "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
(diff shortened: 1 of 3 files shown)


pschoffer (approved):
Looks good to me

brunobasto (approved):
LGTM! Thank you!

PR #961, opened 2022-01-26
Set fileMeta.fileName to latest fileNameToUpload

nejclovrencic:
## Description

When we execute search-replace as part of an import, we can write the file to another location. Due to recent changes, the wrong fileName was passed to the upload function so the md5 calculation failed.

## Steps to Test

1. Check out PR.
2. Verify that the following imports work:

- compressed file
- uncompressed file
- uncompressed file with search-replace
- uncompressed file larger than 16MB
- uncompressed file larger than 16MB with search-replace


brunobasto (approved):
LGTM! Thank you

PR #964, opened 2022-01-28
[dev-env] Validate sql on import

pschoffer:
## Description

Mostly it is just lifting the same validation rules we have for production import.

One exception is the addition of validation for "USE <database_name>;". We saw this as one of the often issue with import files when the import is done to an incorrect database thanks to this.

## Steps to Test
Download example SQL:
```

USE something;

/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!40101 SET character_set_client = utf8 */;
  `ID` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `timestamp` bigint(20) unsigned NOT NULL,
  `args` longtext NOT NULL,
  `interval` int(10) unsigned DEFAULT 0,
  `created` datetime NOT NULL,
  `last_modified` datetime NOT NULL,
  PRIMARY KEY (`ID`),
  UNIQUE KEY `ts_action_instance_status` (`timestamp`,`action`(191),`instance`,`status`),
  KEY `status` (`status`)
) ENGINE=InnoDB AUTO_INCREMENT=104 DEFAULT CHARSET=latin1;
/*!40101 SET character_set_client = @saved_cs_client */;

```

Example:

1. Check out PR.
1. Run `npm run build && ./dist/bin/vip-dev-env-import-sql.js ~/Downloads/test.sql`

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -170,6 +170,15 @@ const checks: Checks = {
 		excerpt: "'DROP DATABASE' should not be present (case-insensitive)",
 		recommendation: 'Remove these lines',
 	},
+	useStatement: {
+		matcher: /^USE /i,
+		matchHandler: lineNumber => lineNumber,
+		outputFormatter: errorCheckFormatter,
+		results: [],
+		message: 'USE <DATABASE_NAME> statement',
+		excerpt: "'USE <DATABASE_NAME>' should not be present (case-insensitive)",
+		recommendation: 'Remove these lines',
+	},
 	alterUser: {
 		matcher: /^(ALTER USER|SET PASSWORD)/i,
 		matchHandler: lineNumber => lineNumber,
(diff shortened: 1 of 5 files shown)


src/lib/validations/sql.js
```diff
@@ -353,15 +363,15 @@ export const staticSqlValidations = {
 };
 
 // For standalone SQL validations
-export const validate = async ( filename: string, isImport: boolean = false ) => {
+export const validate = async ( filename: string, skipChecks: string[] = [ 'useStatement' ] ) => {
```
pschoffer, line 366: `isImport` was never used so I removed it.

I added new optional parameter to skip checks... and default to skip this new useStatement validation. From dev-env we send an empty array so that all validation goes.

jesse-greathouse (approved):
Everything looks good.

I tested the code and it worked as expected.
Screen Shot 2022-02-03 at 6 13 32 AM

PR #965, opened 2022-01-28
CI: Node version updates

mjangda:
Drop 10 since it's long-gone. Add explict testing for Node 14/16/17.

## Steps to Test

CI


mjangda:
Not needed anymore.

PR #966, opened 2022-01-31
[dev-env] Dynamic WordPress Image List

jesse-greathouse:
## Description
* On `vip dev-env create` The list of available WordPress Images will be produced dynamically based on a real time call to the images manifest on github.
* Will show a "Pre-Release" tag for images that are pre-release
* The default selection will always be the most recent stable release unless another version was selected.

Screen Shot 2022-02-03 at 1 17 14 PM

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `vip dev-env create`
1. Verify  list of WordPress Images are up to date on the 3rd question of the form.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -362,7 +362,6 @@ async function getVersionList() {
 	let res;
 	const local = xdgBasedir.data || os.tmpdir();
 	const cacheTtl = 86400; // number of seconds that the cache can be considered active.
-	console.log( DEV_ENVIRONMENT_WORDPRESS_CACHE_KEY );
 	const cacheFile = path.join( local, 'vip', DEV_ENVIRONMENT_WORDPRESS_CACHE_KEY );
 
 	// Try to retrieve the file from cache or cache it if invalid
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -346,7 +346,7 @@ async function populateWordPressVersionList( versionList ) {
 		const https = require( 'https' );
 		const req = https.request( apiOptions, res => {
 			let data = '';
-			let tag, tagFormatted, locked, prerelease, mapping;
+			let tag, tagFormatted, prerelease, mapping;
 
 			res.on( 'data', chunk => {
 				data += chunk;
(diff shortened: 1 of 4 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -334,14 +335,53 @@ export function addDevEnvConfigurationOptions( command ) {
...
-	];
+async function populateWordPressVersionList( versionList ) {
+	const apiOptions = {
+		hostname: 'raw.githubusercontent.com',
+		path: '/Automattic/vip-container-images/master/wordpress/versions.json',
+		method: 'GET',
+	};
+
+	return new Promise( resolve => {
+		const https = require( 'https' );
```
pauarge: What about importing the `https` module at the top of the file, with the rest of the imports?

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -334,14 +344,51 @@ export function addDevEnvConfigurationOptions( command ) {
...
-	return [
-		'5.9',
-		'5.8.3',
-		'5.8.2',
-		'5.8.1',
-		'5.8',
-		'5.7.3',
-		'5.7.2',
-	];
+async function populateWordPressVersionList( versionList ) {
```
rinatkhaziev: 1. Instead of using `https` we should use `node-fetch` (as other parts of VIP-CLI do), this will allow for uniformity and better error handling, and make the code more concise.
2. The error handling is a bit rudimentary:
  - What happens if GH  is down?
  - What happens if someone accidentally removed the list from the repo?

rinatkhaziev (dismissed):
Thanks! I left a comment inline regarding the `populateWordPressVersionList`.

Other things to consider:
1) It'd be nice to add some caching (especially in the context of handling potential failures) 
2) Tests seem only to test the happy path

src/lib/dev-environment/dev-environment-cli.js
jesse-greathouse: Thank you for your feedback. I took some time to address the concerns you had, and refactored the code a bit.

Here are the highlights:

- I changed the https api code to use `node-fetch`. This definitely made the net code more concise.
- I improved the error handling around this so it's not going to be hectic for the end user, but uses debug statements to give debugging some help.
- Because of caching, I was careful to design logic around retaining the cache file when it exists. Now, in circumstances where github is down or if someone removes the manifest from the repo, the program will continue to run using the cache; and with a small yellow message about it.
Screen Shot 2022-02-09 at 7 19 16 AM

jesse-greathouse:
> It'd be nice to add some caching (especially in the context of handling potential failures)
Great suggestion. I added caching around the web call to github. 
- Upon successful call to the github file, it will retain that result with an expiration of 24 hours.
- No matter what happens with the remote request, the operation should continue normally if it has the cache from a previous operation (even if the cache is expired).

> Tests seem only to test the happy path
- Also great suggestion. Let me get back to you on that...

sjinks (approved):

rinatkhaziev:
@jesse-greathouse great work, one final thing, let's fix these linting issues:

```
/home/circleci/repo/src/lib/dev-environment/dev-environment-core.js
  57:1  error  Missing JSDoc for parameter 'instancePath'  valid-jsdoc
  57:1  error  Missing JSDoc for parameter 'options'       valid-jsdoc
```

jesse-greathouse:
@rinatkhaziev Please see the most recent commit with regards to the new function: updateWordPressImage

This new function enters the execution path in landoStart and is a bugfix for the scenario where a user has an existing environment with an image that has been removed from publication.

This is also an enhancement because it adds a prompt, asking users if they would like  to update WordPress when it finds that they are not using the most recent release.

PR #967, opened 2022-02-02
Do not use /tmp as a userConfRoot

sjinks:
## Description

`vip dev-env create` uses Lando in such a way that `/tmp` (`os.tmpdir()`) is chosen as a Lando user root configuration directory. This means that `/tmp` (or its equivalent) is mounted as `/lando` into containers.

The issue here is that Lando tries to change the owner of `/lando` and its files and subdirectories to match the UID/GID of the user running Lando. In Linux, this unfortunately succeeds, and as a result, `/tmp` and all of its files and directories change their ownership. This becomes catastrophic for, say, Firefox (at some point, all network requests freeze, the browser crashes), Chromium (it fails to start), and for some software installed from Snap.

For containers created by `vip dev-env create`, `docker inspect` will show something like this:

```
            {
                "Type": "bind",
                "Source": "/tmp",
                "Destination": "/lando",
                "Mode": "cached",
                "RW": true,
                "Propagation": "rprivate"
            },
```

Lando's `scripts/user-perm-helpers.sh` has function, `perm_sweep()`, which becomes destructive when used on host's `/tmp`:

```bash
nohup find /lando -not -user $USER -execdir chown $USER:$GROUP {} \+ > /tmp/perms.out 2> /tmp/perms.err &
```

This PR offers a dedicated temporary directory to Lando so that the `chown` operation does not break the host system.

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Create and start a test environment.
1. Observe that `/tmp`'s permissions/ownership were not changed.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -37,7 +37,7 @@ let landoConfRoot;
 async function getLandoUserConfigurationRoot() {
 	if ( ! landoConfRoot ) {
 		const tmpDir = os.tmpdir();
-		landoConfRoot = await mkdtemp( `${ tmpDir }${ path.sep }lando` );
+		landoConfRoot = await mkdtemp( path.join( tmpDir, 'lando' ) );
 	}
 
 	return landoConfRoot;
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -7,12 +7,16 @@
  * External dependencies
  */
 import debugLib from 'debug';
+import fs from 'fs';
+import os from 'os';
 import path from 'path';
+import { promisify } from 'util';
 import Lando from 'lando/lib/lando';
 import landoUtils from 'lando/plugins/lando-core/lib/utils';
 import landoBuildTask from 'lando/plugins/lando-tooling/lib/build';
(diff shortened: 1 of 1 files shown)


sjinks:
r2259-stacks

rinatkhaziev (approved):
LGTM but cc @pschoffer @pauarge for awareness.

pschoffer:
Great thanks for this!

pschoffer (approved):

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -23,8 +27,26 @@ import App from 'lando/lib/app';
...
-function getLandoConfig() {
+let landoConfRoot;
+
+/**
+ * @returns {Promise<string>} User configuration root directory (aka userConfRoot in Lando)
+ */
+async function getLandoUserConfigurationRoot() {
+	if ( ! landoConfRoot ) {
+		const tmpDir = os.tmpdir();
+		landoConfRoot = await mkdtemp( `${ tmpDir }${ path.sep }lando` );
```
pauarge: Could we use `path.join` instead of concatenating strings here?

sjinks: Fixed in b8b447e

pauarge (approved):

PR #968, opened 2022-02-03
Update minimum Node version

mjangda:
Node 10 was EOL some time ago so drop support for it.


djalmaaraujo (approved):
Thanks @mjangda! LGTM.

PR #971, opened 2022-02-04
[dev-env] Use custom add user command

pschoffer:
## Description

In https://github.com/Automattic/vip-container-images/pull/71 we added a custom command to add user after sql import. This command covers more edge cases and is better suited for after import on dev-env.

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Run `.npm run build && ./dist/bin/vip-dev-env-import-sql.js ~/Downloads/test.sql`
1. Verify the new user gets created and is admin


jesse-greathouse (approved):
very nice.

PR #975, opened 2022-02-17
[dev-env] prompt On Unselected Env

pschoffer:
## Description

If on dev-env users provides only app id `@123` but not the env we used to ignore env settings.

This change prompts user to choose from existing envs

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build && ./dist/bin/vip-dev-env-create.js @123`


jesse-greathouse (approved):
nice work

sjinks (approved):

PR #976, opened 2022-02-17
Fix/duplicate shortcut parameter

jesse-greathouse:
## Description

`vip dev-env create -h` has two occurrences of suggesting the parameters `-M`

```
Usage: vip dev-env-create [options]

  Options:
    -c, --client-code            Use the client code from a local directory or VIP skeleton
    -e, --elasticsearch          Explicitly choose Elasticsearch version to use
    -h, --help                   Output the help for the (sub)command
    -M, --mariadb                Explicitly choose MariaDB version to use
    -M, --media-redirect-domain  Domain to redirect for missing media files. This can be used to still have images without the need to import them locally.
```
This commit will change the second `-M` to be `-r` instead:

Screen Shot 2022-02-17 at 11 42 37 AM

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `vip dev-env create -h`
1. Look at the usage options to confirm that there are not 2x`-M`.


sjinks (approved):
Thank you!

PR #977, opened 2022-02-19
Try to fix Flow errors

sjinks:
## Description

An attempt to fix Flow errors in Dev Env-related code.

PS: I have no idea why `npm run flow` did not find these errors. VSCode with Flow extension found lots of errors immediately.

## Steps to Test

The CI should pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ workflows:
       - lint
       - flow
       - test:
-          version: 12-stretch
-          name: 12-stretch
+          version: 14-stretch
+          name: 14-stretch
           requires:
             - lint
             - flow
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=12"
+    "node": ">=14.14.0"
   },
   "bugs": {
     "url": "https://github.com/Automattic/vip/issues"
@@ -97,7 +97,7 @@
     "eslint-plugin-no-async-foreach": "0.1.1",
     "eslint-plugin-react": "7.26.0",
     "eslint-plugin-wpcalypso": "5.0.0",
-    "flow-bin": "0.150.0",
+    "flow-bin": "0.172.0",
     "jest": "27.2.1",
     "nock": "13.0.11",
     "prettier": "npm:wp-prettier@2.0.5",
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -65,6 +65,8 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			phpmyadmin: currentInstanceData.phpmyadmin,
 			xdebug: currentInstanceData.xdebug,
 			mediaRedirectDomain: currentInstanceData.mediaRedirectDomain,
+			multisite: false,
+			title: '',
 		};
 
 		const instanceData = await promptForArguments( preselectedOptions, defaultOptions );
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -157,7 +157,7 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
 	}
 
 	for ( const component of DEV_ENVIRONMENT_COMPONENTS ) {
-		const option = preselectedOptions[ component ].toString();
+		const option = ( preselectedOptions[ component ] ?? '' ).toString();
 		const defaultValue = ( defaultOptions[ component ] ?? '' ).toString();
 
 		const result = await processComponent( component, option, defaultValue );
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -25,7 +25,7 @@ import {
 	DEV_ENVIRONMENT_COMPONENTS,
 	DEV_ENVIRONMENT_NOT_FOUND,
 } from '../constants/dev-environment';
-import { InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
+import type { AppInfo, ComponentConfig, InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
@@ -47,7 +47,7 @@ export function handleCLIException( exception: Error ) {
 	}
 }
 
-export function getEnvironmentName( options: EnvironmentNameOptions ) {
+export function getEnvironmentName( options: EnvironmentNameOptions ): string {
 	if ( options.slug ) {
 		return options.slug;
 	}
@@ -61,7 +61,7 @@ export function getEnvironmentName( options: EnvironmentNameOptions ) {
 	return DEFAULT_SLUG;
 }
 
-export function getEnvironmentStartCommand( options: EnvironmentNameOptions ) {
+export function getEnvironmentStartCommand( options: EnvironmentNameOptions ): string {
 	if ( options.slug ) {
 		return `${ DEV_ENVIRONMENT_FULL_COMMAND } start --slug ${ options.slug }`;
 	}
@@ -101,12 +101,8 @@ export function processComponentOptionInput( passedParam: string, allowLocal: bo
 }
 
 export function getOptionsFromAppInfo( appInfo: AppInfo ): InstanceOptions {
-	if ( ! appInfo ) {
-		return {};
-	}
-
 	return {
-		title: appInfo.environment?.name || appInfo.name,
+		title: appInfo.environment?.name || appInfo.name || '',
 		multisite: !! appInfo?.environment?.isMultisite,
 		mediaRedirectDomain: appInfo.environment?.primaryDomain,
 	};
@@ -118,7 +114,7 @@ export function getOptionsFromAppInfo( appInfo: AppInfo ): InstanceOptions {
  * @param {InstanceOptions} defaultOptions - options to be used as default values for prompt
  * @returns {any} instance data
  */
-export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: InstanceOptions ): InstanceData {
+export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: $Shape<InstanceOptions> ): Promise<InstanceData> {
 	debug( 'Provided preselected', preselectedOptions, 'and default', defaultOptions );
 
 	console.log( DEV_ENVIRONMENT_PROMPT_INTRO );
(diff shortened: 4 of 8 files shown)


package.json
```diff
@@ -70,7 +70,7 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=12"
+    "node": ">=14.14.0"
```
sjinks, line 73: The code uses [`fs.promises.rm`](https://github.com/Automattic/vip/blob/808ad5287ee8a8974df46ff4ccb4e4604e2f80d4/src/lib/dev-environment/dev-environment-core.js#L161), which was added in [Node 14.14](https://nodejs.org/api/fs.html#fspromisesrmpath-options).

src/bin/vip-dev-env-update.js
```diff
@@ -65,6 +65,8 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			phpmyadmin: currentInstanceData.phpmyadmin,
 			xdebug: currentInstanceData.xdebug,
 			mediaRedirectDomain: currentInstanceData.mediaRedirectDomain,
+			multisite: false,
+			title: '',
```
sjinks, line 69: Cannot assign object literal to `defaultOptions` because property `multisite` is missing in  object literal [1] but exists in  `InstanceOptions` [2].

Cannot assign object literal to `defaultOptions` because property `title` is missing in  object literal [1] but exists in  `InstanceOptions` [2].

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -47,7 +47,7 @@ export function handleCLIException( exception: Error ) {
 	}
 }
 
-export function getEnvironmentName( options: EnvironmentNameOptions ) {
+export function getEnvironmentName( options: EnvironmentNameOptions ): string {
```
sjinks, line 50: Here and below:

Cannot build a typed interface for this module. You should annotate the exports of this module with types. Missing type annotation at function return

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -25,7 +25,7 @@ import {
 	DEV_ENVIRONMENT_COMPONENTS,
 	DEV_ENVIRONMENT_NOT_FOUND,
 } from '../constants/dev-environment';
-import { InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
+import type { AppInfo, ComponentConfig, InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
```
sjinks, line 28: Types should be imported with `type`.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -101,12 +101,8 @@ export function processComponentOptionInput( passedParam: string, allowLocal: bo
 }
 
 export function getOptionsFromAppInfo( appInfo: AppInfo ): InstanceOptions {
-	if ( ! appInfo ) {
-		return {};
-	}
```
sjinks, line 106: `{}` does not match `InstanceOptions`. Instead of returning an empty object, we try to make up one.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -118,7 +114,7 @@ export function getOptionsFromAppInfo( appInfo: AppInfo ): InstanceOptions {
  * @param {InstanceOptions} defaultOptions - options to be used as default values for prompt
  * @returns {any} instance data
  */
-export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: InstanceOptions ): InstanceData {
+export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: $Shape<InstanceOptions> ): Promise<InstanceData> {
```
sjinks, line 117: 1. `async` functions always return Promises.
2. When we call `promptForArguments()`, we may pass a subset of `InstanceOptions`; therefore, change the type to `$Shape<InstanceOptions>`.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -137,34 +133,48 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
...
-		clientCode: {},
+		wordpress: {
+			mode: 'image',
+		},
+		muPlugins: {
+			mode: 'image',
+		},
+		clientCode: {
+			mode: 'image',
+		},
```
sjinks, line 144: Cannot assign object literal to `instanceData` because property `mode` is missing in  object literal [1] but exists in  `ComponentConfig` [2] in property `XXX`.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -137,34 +133,48 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
...
 		statsd: false,
 		phpmyadmin: false,
 		xdebug: false,
+		siteSlug: '',
```
sjinks, line 148: Cannot assign object literal to `instanceData` because property `siteSlug` is missing in  object literal [1] but exists in  `InstanceData` [2].

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -137,34 +133,48 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
...
 	};
 
 	if ( ! instanceData.mediaRedirectDomain && defaultOptions.mediaRedirectDomain ) {
 		const mediaRedirectPromptText = `Would you like to redirect to ${ defaultOptions.mediaRedirectDomain } for missing media files?`;
 		const setMediaRedirectDomain = await promptForBoolean( mediaRedirectPromptText, true );
 		if ( setMediaRedirectDomain ) {
-			instanceData.mediaRedirectDomain = defaultOptions.mediaRedirectDomain;
+			instanceData.mediaRedirectDomain = defaultOptions.mediaRedirectDomain ?? '';
```
sjinks, line 155: Cannot assign `defaultOptions.mediaRedirectDomain` to `instanceData.mediaRedirectDomain` because  undefined [1] is incompatible with  string [2].

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -137,34 +133,48 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
...
 		}
 	}
 
 	for ( const component of DEV_ENVIRONMENT_COMPONENTS ) {
-		const option = preselectedOptions[ component ];
-		const defaultValue = defaultOptions[ component ];
+		const option = preselectedOptions[ component ].toString();
+		const defaultValue = ( defaultOptions[ component ] ?? '' ).toString();
```
sjinks: I wish we used TypeScript…

Value types of `InstanceOptions` properties can be strings and booleans. Flow does not like the `boolean` part, and it is hard to tell it that `boolean` won't happen here; `toString()` seems to be the least evil here.

The situation is much worse with `defaultValue`, as it is `$Shape<InstanceOptions>` and, therefore, we can have `undefined` here, which does not have a `toString()` method.

We have to use strings because this is what `processComponent()` expects.

sjinks: The error was

Cannot call `processComponent` with `option` bound to `preselectedValue` because  boolean [1] is incompatible with  string [2].

Cannot call `processComponent` with `defaultValue` bound to `defaultValue` because  boolean [1] is incompatible with  string [2].

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -137,34 +133,48 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
...
 
-		instanceData[ component ] = await processComponent( component, option, defaultValue );
+		const result = await processComponent( component, option, defaultValue );
+		if ( null === result ) {
+			throw new Error( 'processComponent() returned null' );
+		}
```
sjinks, line 166: Now, `processComponent()` is allowed to return `null`, and `null` is not allowed here. I am not sure that to throw an exception is the best way here, but I don't know what else we can do.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -244,7 +252,7 @@ const componentDisplayNames = {
 	clientCode: 'site-code',
 };
 
-export async function promptForComponent( component: string, allowLocal: boolean, defaultObject: ComponentConfig ): Promise<ComponentConfig> {
+export async function promptForComponent( component: string, allowLocal: boolean, defaultObject: ComponentConfig | null ): Promise<ComponentConfig> {
```
sjinks, line 255: We may pass `null` here in https://github.com/Automattic/vip/blob/c6c7ee04be649d965f666cdab7b0e680bdd3d271/src/lib/dev-environment/dev-environment-cli.js#L192, https://github.com/Automattic/vip/blob/c6c7ee04be649d965f666cdab7b0e680bdd3d271/src/lib/dev-environment/dev-environment-cli.js#L207

Cannot call `promptForComponent` with `defaultObject` bound to `defaultObject` because  null [1] is incompatible with  `ComponentConfig` [2].

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -126,7 +126,7 @@ export async function updateEnvironment( instanceData: InstanceData ) {
 
 function preProcessInstanceData( instanceData: InstanceData ): InstanceData {
 	const newInstanceData = {
-		...instanceData,
+		...( instanceData: Object ),
```
sjinks, line 129: Cannot spread object literal because Flow cannot determine a type for  object literal [1].  `InstanceData` [2] cannot be spread because interfaces do not track the own-ness of their properties. Try using an object type instead.

src/lib/dev-environment/dev-environment-cli.js
sjinks: And it looks like `preselectedOptions[ component ]` can also be `undefined` :-(

Ref: https://app.circleci.com/pipelines/github/Automattic/vip/7754/workflows/b00467fb-3d47-4aad-b174-d8a14cb30c73/jobs/41049?invite=true#step-103-632

.circleci/config.yml
```diff
@@ -6,8 +6,8 @@ workflows:
       - lint
       - flow
       - test:
-          version: 12-stretch
-          name: 12-stretch
+          version: 14-stretch
+          name: 14-stretch
```
sjinks, line 10: Because of [this](https://github.com/Automattic/vip/pull/977#discussion_r810415553)

pschoffer:
Looks good to me, but we might need somebody from the User Experience team to sign off on this. Especially on that Node requirement change.

We might need to rewrite the code to not use the promisified `fs` operations. I am not sure what node versions do we suppoert.

pschoffer (approved):

package.json
sjinks: Per https://docs.wpvip.com/technical-references/vip-cli/installing-vip-cli/#h-update-node-js-and-npm, it should be OK:

> VIP-CLI requires a minimum installed version of Node.js v14 and npm v8. It is recommended to use the [latest active long-term support release](https://nodejs.org/en/about/releases/) of Node.js, which is currently Node.js v16.

PR #980, opened 2022-02-28
Fix/tag formatting on stapled images

jesse-greathouse:
### Bugfix 1
When the user selects an option for which WordPress version, to use in a new environment, if they select one of the stapled images, it tries to use that entire text as the option for the image, so the .lando.yml file looks like this:

```yml
  wordpress:
    type: compose
    services:
      image: ghcr.io/automattic/vip-container-images/wordpress:5.7   →  5.7.5
      command: sh -c "rsync -a /wp/ /shared/; chown www-data -R /shared; sleep infinity"
      volumes:
        - ./wordpress:/shared
        - type: volume
          source: scripts
          target: /scripts
          volume:
            nocopy: true
```

And that obviously will not work :-)

This change extracts the actual image tag from that string, and it throws a fatal error if it cannot.

### Bugfix 2
The way that the cli saw the currently used WP version in the dev-env was from the version php file in WordPress.

This created certain situations where the current image availability could not be evaluated and still resulted in error.

To fix this I have changed it so that the CLI does not determine the current WP version from the php file, but rather it scrapes the image tag from the .lando.yml file.

This way it will always be able to evaluate if the currently used image has availability.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-dev-env-create.js --slug=test`
1. Select WordPress Version:5.7   →  5.7.5
1. Finish the create sequence.
1. Run `./dist/bin/vip-dev-env-start.js --slug=test`
1. Verify the WordPress 5.7 image gets pulled successfully.


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -305,26 +305,43 @@ export async function promptForComponent( component: string, allowLocal: boolean
...
+			choices: formatted,
 			initial: initialTagIndex,
 		} );
-		const tag = await selectTag.run();
+		const option = await selectTag.run();
+
+		// Validate the input
+		// Some of the options are like: '5.7   →  5.7.5'
+		// Extract first occurrence of something that looks like a tag
+		const tagRgx = new RegExp( /(\d+\.\d+(?:\.\d+)?)/ );
```
pschoffer, line 338: Should we use the meta file instead of parsing lando files?

https://github.com/Automattic/vip/blob/6866b9275a8b3721162076912611c28bd42a8ec9/src/lib/dev-environment/dev-environment-core.js#L239

We already should have all that information in meta file that is used during `update` in a way easier to parse format.

pschoffer: Oh sorry this is a wrong place I guess... I am reacting to this from description:

> To fix this I have changed it so that the CLI does not determine the current WP version from the php file, but rather it scrapes the image tag from the .lando.yml file.

pschoffer (changes requested):
Looks good just had that one parsing logic to read

src/lib/dev-environment/dev-environment-cli.js
jesse-greathouse: It took me a sec to get what you are saying but I think you are right. I will make this change and see how it works.

jesse-greathouse: In importing `readEnvironmentData` to `dev-environment-cli.js` I realized there is a circular dependency in that `dev-environment-core.js` also imports from `dev-environment-cli.js`

I have looked at the imports on both of these files and also noticed another instance of where I have created a circular dependency with something else.

I need to do some refactoring but before I do that I am going to do this circular dependency anyways because it doesn't actually cause problems it's just grotesque.

This is all just to say that right after this I am going to do a refactor that addresses these circular dependencies.

jesse-greathouse: @pschoffer please see the most recent commit. 

I used `readEnvironmentData` to read the environment config, and `prepareLandoEnv` to save changes to the config.

Please let me know if you see anything else that I can fix.

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -440,20 +439,16 @@ export async function importMediaPath( slug: string, filePath: string ) {
...
-	const landoFile = `${ instancePath }/.lando.yml`;
-	const versionFile = `${ instancePath }/${ DEV_ENVIRONMENT_WORDPRESS_VERSION_FILE }`;
+	const instancePath = getEnvironmentPath( slug );
 
-	// If versionFile doesn't exist it means that the environment has not been initiated.
-	if ( ! fs.existsSync( versionFile ) ) {
-		return;
-	}
+	// Get the current image tag that the WP image is currently using in the .lando.yml file
+	const current = readEnvironmentData( slug );
```
pschoffer: This can throw an exception if the env is build with older `vip` version ... we should probably catch it here

Or maybe change the method signature to catch inside it and return null.

Check `update` logic on how it is done there (and potentially change there too) - https://github.com/Automattic/vip/blob/801ed99d178eaacc5639fef7ac41a1e55df9f7bc/src/bin/vip-dev-env-update.js#L80

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -440,20 +439,16 @@ export async function importMediaPath( slug: string, filePath: string ) {
...
-	const vsnRgx = new RegExp( /\$wp_version \= \'(.*)\'\;/ );
-	const landoFile = `${ instancePath }/.lando.yml`;
-	const versionFile = `${ instancePath }/${ DEV_ENVIRONMENT_WORDPRESS_VERSION_FILE }`;
+	const instancePath = getEnvironmentPath( slug );
 
-	// If versionFile doesn't exist it means that the environment has not been initiated.
-	if ( ! fs.existsSync( versionFile ) ) {
-		return;
-	}
+	// Get the current image tag that the WP image is currently using in the .lando.yml file
```
pschoffer: The comment here needs updating or? We no longer use `.lando.yml` file, correct?

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -504,20 +488,16 @@ async function updateWordPressImage( instancePath, options ) {
...
-
-		// Change the lando file and rebuild.
-		const data = fs.readFileSync( landoFile, { encoding: 'utf8', flag: 'r' } );
-		const edit = data.replace( /ghcr\.io\/.*wordpress\:.*\d+\.\d+(?:\.\d+)?/g, selectedImage );
-		fs.writeFileSync( landoFile, edit );
 
-		// Stage for rebuild
+		// // Write new data and stage for rebuild
+		current.wordpress.tag = version.tag;
 		options.skipRebuild = false;
```
pschoffer: Hmm this mutating of argument can be a bit confusing. What do you think of returning boolean from `updateWordPressImage`. True if it was changed.

Then in the caller you can do:
```
const updated = await updateWordPressImage( slug, options );

	if ( options.skipRebuild && !updated) {
		await landoStart( instancePath );
	} else {
		await landoRebuild( instancePath );
	}
```

pschoffer (dismissed):
Great job. It looks way cleaner now. Not to mention the old code was breaking the  metafile by not updating it.

pschoffer (approved):
Nice work!

PR #982, opened 2022-03-02
Remove unused dependencies:

pschoffer:
## Description

After fixing of lando's conf root https://github.com/Automattic/vip/pull/970  we introduced dependency to `core-js`.

That dependency causes issues on some versions of node. And mainly seems to be not needed. 

This change removes all the unused dependencies.

## Steps to Test

` npm run build && ./dist/bin/vip-dev-env-start.js --slug test` or similar commands

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -7,16 +7,13 @@
  * External dependencies
  */
 import debugLib from 'debug';
-import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import { promisify } from 'util';
 import Lando from 'lando/lib/lando';
 import landoUtils from 'lando/plugins/lando-core/lib/utils';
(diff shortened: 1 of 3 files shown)


pschoffer:
Please review carefully. There might be some basics from javascript module imports.

However, everything seems to work as expected after the change.

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -7,16 +7,13 @@
...
-import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import { promisify } from 'util';
 import Lando from 'lando/lib/lando';
 import landoUtils from 'lando/plugins/lando-core/lib/utils';
 import landoBuildTask from 'lando/plugins/lando-tooling/lib/build';
 import chalk from 'chalk';
 import App from 'lando/lib/app';
-import { _ } from 'core-js';
```
nejclovrencic, line 19: We should probably remove this from package.json as well

nejclovrencic (approved):

PR #986, opened 2022-03-08
Clean the build folder prior to rebuilding it

ingeniumed:
## Description

This is meant to clean out the dist that's been generated prior to building it again. It's been targeted at the build step since that's whats called the most.

## Steps to Test

1. Check out PR.
2. Run `npm run prepare`

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,10 +42,11 @@
   },
   "scripts": {
     "test": "npm run lint && npm run flow && jest --coverage",
+    "clean": "rimraf dist",
     "prepare": "npm run build",
     "prepack": "npm run prepareConfig:publish",
     "postinstall": "node ./helpers/check-version.js",
-    "build": "npm run prepareConfig:local && babel src -d dist",
+    "build": "npm run clean && npm run prepareConfig:local && babel src -d dist",
     "build:watch": "babel src -d dist --watch",
     "flow": "flow",
     "jest": "jest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,10 +42,11 @@
   },
   "scripts": {
     "test": "npm run lint && npm run flow && jest --coverage",
+    "clean": "rimraf dist",
     "prepare": "npm run build",
     "prepack": "npm run prepareConfig:publish",
     "postinstall": "node ./helpers/check-version.js",
-    "build": "npm run prepareConfig:local && babel src -d dist",
+    "build": "npm run clean && npm run prepareConfig:local && babel src -d dist",
```
mjangda: I wonder if we should just do this in `prepare` only to avoid slowing down local development (i.e. `"prepare": "npm run clean && npm run build"`).

ingeniumed: Fair point, I can pop it in prepare instead. That would honestly ensure that it would clean out only during publishing

I had been debating on doing that, or the build step. Ultimately decided on the build step so it would be constantly cleaned + built.

ingeniumed: Moved it over to prepare instead

package.json
```diff
@@ -100,7 +101,8 @@
     "jest": "27.2.1",
     "nock": "13.0.11",
     "prettier": "npm:wp-prettier@2.0.5",
-    "publish-please": "5.5.2"
+    "publish-please": "5.5.2",
+    "rimraf": "^3.0.2"
```
mjangda:
```suggestion
    "rimraf": "3.0.2"
```

mjangda (approved):
Minor note to pin the new `rimraf dependency, but otherwise, this is great! 🚢

package.json
ingeniumed: Good point, just pushed up a fix to pin that version

PR #987, opened 2022-03-08
Add support for M1 Macs in the search-replace tests

andrea-sdl:
## Description

Currently, the CLI can't be fully tested on M1 Macs because the binary for the go-search-replace is missing the darwin-arm version.

This PR adds the missing binary, so everyone with an M1 Mac can also test the app locally with no need to comment code.

## Steps to Test

To effectively test the change, it should be better to test on an M1 Mac. 
The difference will be that after this change on an M1 you can run `npm t` and the tests will pass

1. Check out PR.
2. Run `npm run build`
3. Run `npm t`
4. Tests should pass on both M1 and not M1


brunobasto (approved):
Approving since this was tested by folks with a Mac M1 and confirmed to work on that architecture. Thank you @andrea-sdl 🎉

PR #990, opened 2022-03-10
[dev-env] Add check for wp folder map

pschoffer:
## Description

We have had a long running and hard to reproduce issue that results in the following error:
```
Warning: require(/wp/wp-includes/pomo/mo.php): failed to open stream: No such file or directory in /wp/wp-settings.php on line 116 [$ /usr/local/bin/wp option get siteurl] [wp-settings.php:116 require(''), phar:///usr/local/binvendor/wp-cli/wp-cli/php/WP_CLI/Runner.php:1271 require('wp-settings.php'), phar:///usr/local/binvendor/wp-cli/wp-cli/php/WP_CLI/Runner.php:1192 WP_CLI\Runner-&gt;load_wordpress(), phar:///usr/local/binvendor/wp-cli/wp-cli/php/WP_CLI/Bootstrap/LaunchRunner.php:23 WP_CLI\Runner-&gt;start(), phar:///usr/local/binvendor/wp-cli/wp-cli/php/bootstrap.php:77 WP_CLI\Bootstrap\LaunchRunner-&gt;process(), phar:///usr/local/binvendor/wp-cli/wp-cli/php/wp-cli.php:27 WP_CLI\bootstrap(), phar:///usr/local/binphp/boot-phar.php:11 include('phar:///usr/local/binvendor/wp-cli/wp-cli/php/wp-cli.php'), /usr/local/bin/wp:4 include('phar:///usr/local/binphp/boot-phar.php')]
```

We were not able to pinpoint the root cause at it seems to happen randomly and not very often. My suspicion is that the `rsync` from WordPress image isn't done by the time we run the initial setup - https://github.com/Automattic/vip/blob/e23e04a3d9f2e08b10c9c96b68124aa395838811/assets/dev-env.lando.template.yml.ejs#L129

In this PR we added a healthcheck on `php` container, which is the target of the mount. This healthcheck verifies that there is `/wp/wp-settings.php` file already mounted. If the above is the root cause of the issue this should force initial setup to not start until `wp` files are rsynced and correctly mounted.

## Steps to Test

Impossible...:)

1) create new env
2) start it 
3) See: 
```
Waiting for service database ...
Waiting for service php ...
Waiting for service vip-search ...
```
4) It should start up eventually

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -41,7 +41,7 @@ services:
       command: run.sh
       working_dir: /wp
       healthcheck:
-        test: 'cat /wp/wp-settings.php'
+        test: 'cat /wp/wp-includes/pomo/mo.php'
       environment:
         XDEBUG: <%= xdebug ? 'enable' : 'disable' %>
         STATSD: <%= statsd ? 'enable' : 'disable' %>
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -40,6 +40,8 @@ services:
       image: ghcr.io/automattic/vip-container-images/php-fpm:7.4
       command: run.sh
       working_dir: /wp
+      healthcheck:
+        test: 'cat /wp/wp-settings.php'
       environment:
         XDEBUG: <%= xdebug ? 'enable' : 'disable' %>
         STATSD: <%= statsd ? 'enable' : 'disable' %>


jesse-greathouse (approved):

rinatkhaziev:
> whats the worst that could happen? 🤷

@pschoffer can this potentially result in dev-env not starting at all?

rinatkhaziev:
Marking this as blocked until we figure out the potential implications. I.e. my worry is that this can prevent the environment from starting without meaningful ways to debug.

pschoffer:
> Marking this as blocked until we figure out the potential implications. I.e. my worry is that this can prevent the environment from starting without meaningful ways to debug.

I am not sure what you mean? Local dev-env needs WordPress source files in order to work in any sort of meaningful way. We do provide those files always in an image and they are mapped to the `php-fpm` image. They are in no way optional is my point.

User could go ahead and delete them which would result in dev-env waiting on this healthcheck till reaching the upper limit of [20 retries](https://github.com/Automattic/vip/blob/335db457284efc257204e1a0f966790be2ba5ad5/src/lib/dev-environment/dev-environment-lando.js#L116). But I guess that is ok or?

rinatkhaziev:
@pschoffer I tried to reproduce a bazillion times again, was able to only once. I think the healthcheck is the correct approach, anecdotally I did see vip-search and database get ready earlier than PHP. If your theory is correct about the rsync is correct this should fix, but if it's chown that's failing for some reason this may result in hitting max retries. We'll see I guess.

PR #992, opened 2022-03-17
Add --debug flag to all commands

nejclovrencic:
## Description

This PR adds a `--debug` flag to all commands, in order to simplify turning on debug output. There are other options to turn on debug, but it's done differently on Mac/Linux and Windows machine, thus we want to introduce this flag.

NOTE: When we use `DEBUG=* vip app list` debug output is enabled since the very start of code execution, while the newly introduced flag enables it in the command module. For this reason, in some edge cases we might still have to ask customers to run commands using `DEBUG=*`.

## Steps to Test

1. Check out PR.
2. Build the new version
3. Run any VIP command with debug option and test different combinations:
- `vip app list -d`: should show all debug output 
- `vip app list --debug`: same as above
- `vip app list -d="@automattic/*"`: should show only @automattic/* debug output

### DIFF
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -25,6 +25,7 @@ import pager from 'lib/cli/pager';
 import { parseEnvAliasFromArgv } from './envAlias';
 import { rollbar } from '../rollbar';
 import * as exit from './exit';
+import debugLib from 'debug';
 
 function uncaughtError( err ) {
 	// Error raised when trying to write to an already closed stream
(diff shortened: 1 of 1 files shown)


saroshaga (approved):
Makes sense! thanks for adding this in! 
Deploy Notes would be update the docs??

yolih:
I have my eye on this and will update the Docs once it's merged. :)

nejclovrencic:
@yolih I just merged but we still need to wait for the official release. I guess it should happen next week

mjangda:
> For this reason, in some edge cases we might still have to ask customers to run commands using DEBUG=*.

One concern I'll flag with this is that we're trying to solve a documentation problem with extra code that will make it harder to debug stuff given the edge case noted.

PR #993, opened 2022-03-18
Fix in caching version list.

jesse-greathouse:
## Description
  * TTL was undefined because errantly imported from dev-environment constants
  * fixed typo in cache file

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-dev-env-create.js --slug="newnew"`
1. Verify cache file is being refreshed every 60 seconds: `/Users/username/.local/share/vip/wordpress-versions.json`


rinatkhaziev (approved):
thanks!

PR #999, opened 2022-04-05
Add Feature to Support HTTPS/HTTP/NO_PROXY Settings from System

Frank-Cerny-FRBC:
## Description

**Goal:** To allow individuals operating within non-SOCKS corporate proxies (namely HTTP/HTTPS) to use the VIP-CLI since node-fetch does not "respect" system environment variables

- Added "feature" environment variable VIP_PROXY_OTHER_ENABLED to activate this functionality (while keeping the library backwards compatible for those who don't use a proxy)
- Once enabled, all requests that originally used the SocksProxyAgent will use a proxy agent based on their respective environment variables (including NO_PROXY)
- Allows for upper or lowercase environment variables to be specified (upper case takes precedence)
- The VIP_PROXY functionality is unchanged. Regardless of the existence of any other environment variable, VIP_PROXY takes precedence no matter what

## Steps to Test

1. Check out PR.
2. Run `npm install`
3. Run `npm run build`
4. Run `./dist/bin/vip.js` with your favorite commands (too many here to mention)
5. Verify commands run as expected given the environment variables set on your system

**Note: I realize this is an "impactful" change given how often web requests are made. That being said, I am not sure the best way to test this with confidence since everyone works in a slightly different environment proxy wise. So any thoughts on that would be appreciated.**

## Potential Related Work

https://github.com/Automattic/vip/tree/add/proxy-support
https://github.com/Automattic/vip/tree/renovate/socks-proxy-agent-6.x


mjangda:
Thanks for the PR, @Frank-Cerny-FRBC! Really appreciate you taking the time to propose this. We'll take a look and follow-up with any feedback + next steps.

andrea-sdl:
Hi @Frank-Cerny-FRBC and thanks again for the contribution!

This is a nice feature, you already found out some branches that were partially working on extending proxy support, but this is the most complete implementation feature-wise.

I'm adding some comments into the PR, let us know if you're open to iterate on the PR or if you don't have the bandwith / time since we'd love to include these features too.

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
...
+// Parameters:
+//	- url (string): absolute desintation URL (including the protocol)
+//	- httpsProxy (string | null): string location of https proxy (or null)
+//	- httpProxy (string | null): string location of http proxy (or null)
+// Requires:
+//	- a no_proxy variable is not applicable to the url
+//	- either the environment variable HTTP_PROXY or HTTPS_PROXY is set
+// Returns:
+//	- Either an instance of httpsProxyAgent or httpProxy agent depending on passed in values
+function GetWebProxyAgentBasedOnProtocol( url, httpsProxy, httpProxy ) {
```
andrea-sdl: Mostly Aestetic stuff here.
You'll notice that functions are usually lowerCamelCase. 
It doesn't seem to be enforced by the linter but to avoid confusion we might want to move them to this format to have coherence with the rest of the code.

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
...
+	const protocol = url.substr( 0, 5 );
+	if ( protocol !== 'https' && protocol !== 'http:' ) {
```
andrea-sdl: I expect the production environment to be in HTTPS only, did you come across the need for HTTP too?

If not we can consider simplifying this part since we would need it mostly for local testing.

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
...
+//	- url (string): absolute desintation URL (including the protocol)
+//	- noProxyString (string | null): string representation of the environments current NO_PROXY or no_proxy variable (NO_PROXY takes precedence)
+// Returns:
+//	- (boolean) true/false depending on result
+// NO_PROXY Rules (based directly on underlying dependency):
+// 1. * (alone): proxy nothing
+// 2. *.site: do not proxy any subdomain of a domain
+// 3. abc.com: do not proxy www.abc.com, abc.com, etc.
+// See getProxyForUrl on npmjs.org for full "ruleset"
+function CoveredInNoProxy( url, noProxyString ) {
```
andrea-sdl: not required but we might want to consider also https://www.npmjs.com/package/uses-proxy as an alternative

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * External dependencies
+ */
+import { SocksProxyAgent } from 'socks-proxy-agent';
+import { HttpsProxyAgent } from 'https-proxy-agent';
+import { HttpProxyAgent } from 'http-proxy-agent';
+import { getProxyForUrl } from 'proxy-from-env';
+
```
andrea-sdl, line 10:
```suggestion
import debugLib from 'debug';
```
Adding some debug logs will help us debug when things don't go as expected.

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
...
+// 5. NO_PROXY is set along with VIP_PROXY_OTHER ENABLED and HTTP_PROXY OR HTTPS_PROXY: An HTTP_PROXY/HTTPS_PROXY is returned assuming NO_PROXY is not applicable to the URL
+function createProxyAgent( url ) {
+	const HTTPS_PROXY = process.env.HTTPS_PROXY || process.env.https_proxy || null;
+	const HTTP_PROXY = process.env.HTTP_PROXY || process.env.http_proxy || null;
+	const VIP_PROXY = process.env.VIP_PROXY || process.env.vip_proxy || null;
+	const NO_PROXY = process.env.NO_PROXY || process.env.no_proxy || null;
+
+	// VIP Socks Proxy should take precedence, should be fully backward compatible
+	if ( VIP_PROXY ) {
+		return new SocksProxyAgent( VIP_PROXY );
```
andrea-sdl, line 37:
```suggestion
		debug( `Enabling VIP_PROXY proxy support using config: ${ VIP_PROXY }` );
		return new SocksProxyAgent( VIP_PROXY );
```
Same as before

__tests__/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,194 @@
...
+import { SocksProxyAgent } from 'socks-proxy-agent';
+import { HttpProxyAgent } from 'http-proxy-agent';
+import { HttpsProxyAgent } from 'https-proxy-agent';
+
+/**
+ * Internal dependencies
+ */
+// Reference for testing with env variables within a test: https://github.com/vuejs/vue-test-utils/issues/193
+
+describe( 'validate CreateProxyAgent', () => {
```
andrea-sdl, line 14: Nice adding the tests for backwards compatibility and so on. 
Have you thought about refactoring into table driven tests?

It helps to see what we expect from a test at a glance and also reduce code verbosity a little bit.

Here's a similar example in case you're interested 
https://github.com/Automattic/vip/blob/add/proxy-support/__tests__/lib/http/proxy-agent.js

And here's an overall overview if you don't already know them 
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,85 @@
...
+	} else if ( process.env.VIP_PROXY_OTHER_ENABLED && ! CoveredInNoProxy( url, NO_PROXY ) && ( HTTPS_PROXY || HTTP_PROXY ) ) {
+		return GetWebProxyAgentBasedOnProtocol( url, HTTPS_PROXY, HTTP_PROXY );
```
andrea-sdl: Here too we might benefit from extra debug logs (once we define if we need both HTTP/HTTPS_PROXY or not)

andrea-sdl (changes requested):
Added some comments. 
The code looks solid, as I said previously let us know if you have the time to iterate or not ;)

frank-cerny:
Thanks for all the great feedback @andrea-sdl! Greatly appreciated. This is my other account, but I was able to make all the suggested changes. I'll reply to each of your comments.

More feedback on the new changes is also appreciated, as I haven't used the table driven approach before (so thanks for showing me!)

src/lib/http/proxy-agent.js
frank-cerny: Done, thanks for catching this.

src/lib/http/proxy-agent.js
frank-cerny: I think I misunderstood on some level the differences between HTTPS and HTTP proxies. For now, I removed all references to using an HTTP proxy (which simplified the implementation). If you think using HTTP for development is going to happen, we can figure out how to enable in development only (or something along those lines).

But as you said, since production will always be HTTPS, there is little need for HTTP proxy support.

src/lib/http/proxy-agent.js
frank-cerny: Added

src/lib/http/proxy-agent.js
frank-cerny: Added

src/lib/http/proxy-agent.js
frank-cerny: Added

__tests__/lib/http/proxy-agent.js
frank-cerny: I refactored the tests into a table driven approach. I think they are definitely less verbose, although you do lose some control (which isn't a big deal). Hard to say if its more complex to read or now (subjective I suppose) but I think the file is significantly cleaner than before (so thank you for the suggestion!)

src/lib/http/proxy-agent.js
frank-cerny: I think either package can work. While proxy-from-env is a bit more complex to use, it seems more popular and supported. Given no_proxy related features don't change much, so again, either can work. I'll still with proxy-from-env for now, but we think uses-proxy will be easier to understand, then I can make that change

__tests__/lib/http/proxy-agent.js
andrea-sdl: Thanks for considering the advice ;) I agree the complexity while reading might vary. 
Personally, I've found them to be beneficial also to test more things (It's often a matter of copy/paste instead of reworking code, which makes adding conditions much easier).

But I'm digressing ;) Closing this since now it seems fine.

src/lib/http/proxy-agent.js
```diff
@@ -0,0 +1,68 @@
...
+function createProxyAgent( url ) {
+	const HTTPS_PROXY = process.env.HTTPS_PROXY || process.env.https_proxy || null;
+	const VIP_PROXY = process.env.VIP_PROXY || process.env.vip_proxy || null;
+	const NO_PROXY = process.env.NO_PROXY || process.env.no_proxy || null;
+
+	// VIP Socks Proxy should take precedence, should be fully backward compatible
+	if ( VIP_PROXY ) {
+		debug( `Enabling VIP_PROXY proxy support using config: ${ VIP_PROXY }` );
+		return new SocksProxyAgent( VIP_PROXY );
+	} else if ( process.env.VIP_PROXY_OTHER_ENABLED && ! coveredInNoProxy( url, NO_PROXY ) && HTTPS_PROXY ) {
```
andrea-sdl: @frank-cerny I've spent some time thinking about this, here's my take, let me know your thoughts.

The whole goal of this PR is to enable people using the CLI to use the Systems proxy they have configured.
So, if we're not using the `VIP_PROXY` we might find ourselves using the `HTTPS_PROXY` environment variable, or even the `HTTP_PROXY` one .

It also seems that some apps allow `HTTP_PROXY` with https sites. The reason why this is possible is that in those cases we are using a proxy that works in `HTTP`. So the communication with the proxy will be unencrypted.

Therefore, after some thought, I was thinking this:
Instead of having a more generic `VIP_PROXY_OTHER_ENABLED` what about changing it to `VIP_USE_SYSTEM_PROXY` and then return the first we find between
* `SOCKS_PROXY`
* `HTTPS_PROXY`
* `HTTP_PROXY`

Always keeping in mind the `NO_PROXY` config.

This would allow us to remove the check on the protocol and also to make full use of whatever proxy configuration we might have on the final system, while, at the same time, allowing us to disable the proxy if we need it. 

What do you think? 
It could look something like
```js
if (process.env.VIP_PROXY_OTHER_ENABLED && ! coveredInNoProxy( url, NO_PROXY )) {
  if ( SOCKS_PROXY ) {
	  debug( `Enabling SOCKS proxy support using config: ${ HTTPS_PROXY }` );
	  return new SocksProxyAgent( SOCKS_PROXY );
  }
  if ( HTTPS_PROXY ) {
	  debug( `Enabling HTTPS proxy support using config: ${ HTTPS_PROXY }` );
	  return new HttpsProxyAgent( HTTPS_PROXY );
  }
  if ( HTTP_PROXY ) {
	  debug( `Enabling HTTP proxy support using config: ${ HTTP_PROXY }` );
	  return new HttpProxyAgent( HTTP_PROXY );
  }
}

**Review by andrea-sdl — CHANGES_REQUESTED** [review]
@frank-cerny Great job! I've left only another minor comment which should simplify a little bit the code.

File: src/lib/http/proxy-agent.js (hunk 8 above)
Comment (Frank-Cerny-FRBC) [reply src/lib/http/proxy-agent.js]: @andrea-sdl I think this is a solid idea. I've used https -> http proxies so your scenario makes sense to me. This should simplify the implementation even more too (as your snippet shows).

I also like how there is still a "feature flag" (VIP_USE_SYSTEM_PROXY) to ensure its all backwards compatible. VIP_USE_SYSTEM_PROXY sounds WAY better than VIP_PROXY_OTHER_ENABLED too haha.

I think as long as the "order of precedence" of proxies is documented (socks -> https -> http) then this approach will work great!'

I can make these updates (i.e. "borrow" your snippet and update the tests within the next couple of days. Thank you for the feedback! (Keep it coming :) )

Comment (frank-cerny) [reply src/lib/http/proxy-agent.js]: @andrea-sdl Changes have been made. Now I fully see the allure of table driven tests. Please let me know if I missed anything or you have any additional feedback!

**Comment by andrea-sdl** [comment]
@frank-cerny ! That's great. I am doing some final testing with your PR and will probably approve in the next days. 
One thing though: I noticed the npm-shrinkwrap.json differs quite a lot from our main one.

I noticed it because I was having issues with a sha512 from the `lando` cli. 
I believe the underlying issue here is that you probably used a different node version.

Would you mind updating it with this? https://github.com/Automattic/vip/blob/add/url_proxy_configuration/npm-shrinkwrap.json? I think this should be enough.

**Comment by frank-cerny** [comment]
@andrea-sdl It has been updated. Thanks for the fix! The diff looks much cleaner in the PR, and actually looks correct now.

I did notice it kept updating to shrinkwrap version 2 when I used node 16, so I think you are right, that was the issue. For future reference (and so I don't make this mistake again), what version of node should I be using? 14?

**Comment by andrea-sdl** [comment]
> For future reference (and so I don't make this mistake again), what version of node should I be using? 14?

Yes, at least for now that's the [minimum supported version we have](https://github.com/Automattic/vip/blob/develop/package.json#L74) so testing/developing with it is a good idea to avoid hiccups.

In any case I'm making a personal TODO to update the documentation so it's clearer on this ;) Thanks again for your help!

File: src/lib/http/proxy-agent.js [hunk 9]
```diff
@@ -0,0 +1,80 @@
+/**
+ * External dependencies
+ */
+import { SocksProxyAgent } from 'socks-proxy-agent';
+import { HttpsProxyAgent } from 'https-proxy-agent';
+import { HttpProxyAgent } from 'http-proxy-agent';
+import { getProxyForUrl } from 'proxy-from-env';
+import debug from 'debug';
```
andrea-sdl: @frank-cerny while testing I noticed that the debug line doesn't work because the first argument is the environment part.
This change should fix it. 

```suggestion
import debugLib from 'debug';
const debug = debugLib( 'vip:proxy-agent' );
```

andrea-sdl (approved):

andrea-sdl:
@frank-cerny merged! Thanks again for your help and collaboration! 
I'll ping you again once this is released to the public and list you+your change in the CHANGELOG.md

nickdaugherty:
Noting that http proxy support has been removed in https://github.com/Automattic/vip/pull/1102

PR #1001, opened 2022-04-06
Fix reconnect event listeners

luismulinari:
## Description

Since version 3, `socket.io-client` no longer forwards the events of its manager. So, the reconnect events need to be listened directly from the manager.

https://github.com/socketio/socket.io-client/blob/master/CHANGELOG.md#300-2020-11-05

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-wp long-running-command`
4. Deploy the public API while the above command is still running.
6. The command should resume after the public API is deployed.


jblz (approved):
Thanks for getting to the bottom of this!

PR #1005, opened 2022-04-14
dev-env: updateWordPress image improvements

rinatkhaziev:
## Description

This PR addresses a few issues:

The most critical is that the selection didn't work properly when a `ref` was pointing to the commit. It's fixed by removing extra filtering that relied on regex. 

Also introduces `don't ask anymore` option to avoid being repeatedly asked whether an upgrade is desired if not on the latest version.

## Steps to Test

Edit `~/.local/share/vip/dev-environment/wordpress-versions.json` to include the following:

```
  {
    "ref": "2be90dc589a1c2e2bde5efa691eafe5407d0f753",
    "tag": "6.0",
    "cacheable": true,
    "locked": true,
    "prerelease": true
  },
```

Either create a new environment or start an existing one and follow the prompts, you should be able to create an environment using 6.0 prerelease.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -47,9 +47,9 @@ export function handleCLIException( exception: Error ) {
 		console.log( errorPrefix, message );
 
 		if ( ! process.env.DEBUG ) {
-			console.log( `Please re-run the command with "${ chalk.bold( 'DEBUG=@automattic/vip:bin:dev-environment' ) }" prepended to it and provide the stack trace on the support ticket.` );
+			console.log( `Please re-run the command with "--debug ${ chalk.bold( '@automattic/vip:bin:dev-environment' ) }" appended to it and provide the stack trace on the support ticket.` );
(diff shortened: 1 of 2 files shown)


WPprodigy:
The new choice should probably be more clear like “No. (Don’t ask anymore)”. Perhaps there is more pleasant wording, but I can’t think of anything 😆

is there a process to manually upgrade even without the prompt? If so we should mention it when they choose this option. If not, should there be?

rinatkhaziev:
> " “No. (Don’t ask anymore)”." 

Yeah, good suggestion.

> is there a process to manually upgrade even without the prompt? If so we should mention it when they choose this option. If not, should there be?

Also a good suggestion. `vip dev-env update` would work.

jesse-greathouse:
I tested all of these paths and found this branch to be successful in the 3 scenarios I tried.

1. No upgrade required
1. With Upgrade
2. No upgrade with "don't ask again"

All of these testing paths get a 👍  from me. Well done.

sjinks (approved):

PR #1008, opened 2022-04-18
WP-CLI: Reattach Reconnect-Related Events After a Successful Reconnection

jblz:
## Description

Currently, when a connection to the back end is broken and the client successfully reconnects a first time, the event listeners are not reset to allow subsequent reconnections to succeed.  This change re-binds the listeners, so additional reconnections per run are possible.

Also, in case the `cli` receives an `exit` signal, it will exit accordingly to the code and message received.

This change also adds some logging so we can monitor these situations and further improve the experience.

### Testing reattach reconnect-related events
1. Check out PR.
2. Run npm run build
3. Run ./dist/bin/vip-wp long-running-command
4. Simulate a network issue
5. Wait for the until to reconnect
6. Simulate another network issue
7. Wait for the until to reconnect
8. The command should resume/reconnect multiple times.

### Testing exit code

1. Check out PR.
2. Run npm run build
3. Run ./dist/bin/vip-wp long-running-command
4. Deploy a new version of the site
5. An error message should appear
6. Run `echo $?`
7. The output should be `101`

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -10,6 +10,7 @@ import SocketIO from 'socket.io-client';
 import IOStream from 'socket.io-stream';
 import readline from 'readline';
 import { Writable } from 'stream';
+import debugLib from 'debug';
 
 /**
  * Internal dependencies
@@ -23,6 +24,8 @@ import { trackEvent } from 'lib/tracker';
 import Token from '../lib/token';
 import { rollbar } from 'lib/rollbar';
 import createSocksProxyAgent from 'lib/http/socks-proxy-agent';
+
+const debug = debugLib( '@automattic/vip:wp' );
 
 const appQuery = `id, name,
 	organization {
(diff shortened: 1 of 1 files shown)


src/bin/vip-wp.js
```diff
@@ -199,6 +202,55 @@ const launchCommandAndGetStreams = async ( { guid, inputToken, offset = 0 } ) =>
 	return { stdinStream, stdoutStream, socket };
 };
 
+const bindReconnectEvents = ( { cliCommand, inputToken, subShellRl, commonTrackingParams, isSubShell } ) => {
+	currentJob.socket.io.on( 'reconnect', async () => {
+		debug( '-------- reconnect' );
+		rollbar.info( 'WP-CLI socket.io.on( \'reconnect\' )', { custom: { code: 'wp-cli-on-reconnect', commandGuid: cliCommand.guid } } );
```
jblz, line 200: I thought some rollbar events might be handy here to make sure we're seeing these come through to the client.  What do you think, @luismulinari?

luismulinari: It is perfect!

jblz:
@luismulinari -- Thanks for taking this over.  It's working well for me, but I can't approve the PR as the author.

Consider this a :+1:  from me!

luismulinari (approved):
LGTM! 🚢 

Thanks, Jeff!

PR #1009, opened 2022-04-20
Add ability to choose a PHP image to use

sjinks:
## Description

This PR adds an option to choose a PHP image to use; by default, we use `ghcr.io/automattic/vip-container-images/php-fpm:7.4`.

With this option, users can try newer versions of PHP in their development environment.

## Steps to Test

After checking this PR out and building the files, `node ./dist/bin/vip-dev-env-create.js -P ghcr.io/automattic/vip-container-images/php-fpm-alt:8.0` should pull the new image, and `node ./dist/bin/vip-dev-env-start.js` should start it. There is a bug in the php-fpm-alt:8.0 which prevents WP from being installed, I am working on it (Automattic/vip-container-images#195).

### DIFF
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -20,7 +20,7 @@ import { DEV_ENVIRONMENT_FULL_COMMAND } from 'lib/constants/dev-environment';
 import { addDevEnvConfigurationOptions, handleCLIException, promptForArguments } from '../lib/dev-environment/dev-environment-cli';
 import type { InstanceOptions } from '../lib/dev-environment/types';
 import { doesEnvironmentExist, readEnvironmentData, updateEnvironment } from '../lib/dev-environment/dev-environment-core';
-import { DEV_ENVIRONMENT_NOT_FOUND } from '../lib/constants/dev-environment';
+import { DEV_ENVIRONMENT_NOT_FOUND, DEV_ENVIRONMENT_PHP_VERSIONS } from '../lib/constants/dev-environment';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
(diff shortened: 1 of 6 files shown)


jesse-greathouse (approved):
I tested this PR. Here are my findings:

I found that the create sequence worked as expected, and the following start sequence worked as expected.
Screen Shot 2022-04-20 at 8 38 50 AM

When I checked the environment profile, the .lando.yml had the correct url for the PHP 8 image:

Screen Shot 2022-04-20 at 8 42 31 AM

I verified that docker was running the correct image:
Screen Shot 2022-04-20 at 8 46 29 AM

I verified the dev-env test site is working under PHP 8
Screen Shot 2022-04-20 at 8 47 42 AM

I am happy to approve this PR.

Thank you for the excellent work, @sjinks !

rinatkhaziev (changes requested):
THANK YOU!

As discussed today we should present the user with a choice of available versions rather than the full image name to make it friendlier. The default choice should be 7.4 (the simpler option) or if a `@ID` notation is provided it should match that value. - but that could be done later.

sjinks:
@rinatkhaziev please take a look

src/bin/vip-dev-env-update.js
```diff
@@ -60,6 +60,7 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			muPlugins: currentInstanceData.muPlugins.dir || currentInstanceData.muPlugins.tag || 'latest',
 			wordpress: currentInstanceData.wordpress.tag,
 			elasticsearch: currentInstanceData.elasticsearch,
+			php: currentInstanceData.php,
```
rinatkhaziev: this would be undefined on older instances so needs to `|| default`

rinatkhaziev (dismissed):
Thanks, left one inline comment to catch the `php` for already existing environments that wouldn't have that key in the `instanceData`.
I opted to do the following: https://github.com/Automattic/vip/commit/95b738114cbc59789d457b5bb13b3ff0449a4bca

jesse-greathouse:
I tested the latest changes. This is good to go.
Screen Shot 2022-04-26 at 9 21 22 AM

PR #1010, opened 2022-04-22
[dev-env] Add Enterprise Search, XDebug, phpMyAdmin options to the config wizard.

rinatkhaziev:
## Description

This PR adds Enterprise Search, XDebug, and phpMyAdmin as selectable options in the `create` and `update` commands, with defaults to false.

Additionally, this PR removes a PHP healthcheck which was not helpful, instead of the healthcheck we now wait for the core files to be present instead of in https://github.com/Automattic/vip-container-images/pull/197.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `npm run link` (this is the easiest)
1. `vip dev-env create --slug=wizard`
1. Verify that the prompts are being displayed
1. Verify that `vip-search` is either present or absent in the generated .lando.yml
1. Verify that older versions of the environment start up as expected (the template file now relies on new `enterpriseSearchEnabled` instance var.

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-cli.js
+++ b/__tests__/lib/dev-environment/dev-environment-cli.js
@@ -7,7 +7,6 @@
  */
 import { prompt, selectRunMock, confirmRunMock } from 'enquirer';
 import nock from 'nock';
-
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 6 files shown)


rinatkhaziev:
I had to remove a test cause I have no idea how to fix it. 😢

jesse-greathouse (approved):
This is looking really great @rinatkhaziev . I've tested this and I approve.
Screen Shot 2022-04-22

sjinks (approved):

PR #1020, opened 2022-05-02
More tracking

pschoffer:
## Description

Adds tracking to the rest of the dev-env subcommands.

## Steps to Test
`npm run build && ./dist/bin/vip-dev-env-update.js --debug` to see track events sent or check 

or check

https://mc.a8c.com/tracks/live/?eventname=vip_cli_dev%25&user=&useragent=

### DIFF
--- a/src/bin/vip-dev-env-exec.js
+++ b/src/bin/vip-dev-env-exec.js
@@ -12,10 +12,12 @@
 /**
  * Internal dependencies
  */
+import { trackEvent } from 'lib/tracker';
 import command from 'lib/cli/command';
 import { getEnvironmentName, handleCLIException } from 'lib/dev-environment/dev-environment-cli';
 import { exec } from 'lib/dev-environment/dev-environment-core';
 import { DEV_ENVIRONMENT_FULL_COMMAND } from 'lib/constants/dev-environment';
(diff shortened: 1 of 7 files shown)


src/bin/vip-dev-env-exec.js
```diff
@@ -38,6 +40,9 @@ command( { wildcardCommand: true } )
 	.argv( process.argv, async ( unmatchedArgs, opt ) => {
 		const slug = getEnvironmentName( opt );
 
+		const trackingInfo = getEnvTrackingInfo( slug );
+		await trackEvent( 'dev_env_exec_command_execute', trackingInfo );
```
sjinks, line 44: My concern here is that if `trackEvent()` fails for some reason, this will abort the whole thing.

// There is a bug in `trackEvent()`: because [this line](https://github.com/Automattic/vip/blob/develop/src/lib/tracker.js#L47) returns a `Promise` instead of `await`'ing it, the `catch` block won't catch any exceptions, and we will end up with an unhandled rejection

pschoffer: I fixed the bug, now the catching should happen in trackEvent correctly.

sjinks (approved):

PR #1022, opened 2022-05-03
[dev-env] Validate docker installed

pschoffer:
## Description

This change mimics [lando native](https://github.com/Automattic/lando-cli/blob/57d2cd8a23e73b3e5165c3e2862b42aa0881ef2f/lib/cli.js#L327) check for docker and docker-compose. We can give user a sensible error rather than failing in the middle with a stack trace.

## Steps to Test

Uninstall docker/docker-compose e.g. `sudo mv /usr/share/lando/bin/docker-compose /usr/share/lando/bin/docker-compose-back` 

Run start/create command.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -30,7 +30,7 @@ import {
 } from '../constants/dev-environment';
 import { getVersionList, readEnvironmentData } from './dev-environment-core';
 import type { AppInfo, ComponentConfig, InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
-import { ensureDockerInstalled } from './dev-environment-lando';
+import { validateDockerInstalled } from './dev-environment-lando';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
@@ -71,7 +71,7 @@ export async function handleCLIException( exception: Error, trackKey?: string, t
 
 export const validateDependencies = async () => {
 	try {
-		await ensureDockerInstalled();
+		await validateDockerInstalled();
 	} catch ( exception ) {
 		exit.withError( exception.message );
 	}
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -17,6 +17,7 @@ import os from 'os';
 /**
  * Internal dependencies
  */
+import * as exit from 'lib/cli/exit';
 import { trackEvent } from '../tracker';
 import {
 	DEV_ENVIRONMENT_FULL_COMMAND,
@@ -29,6 +30,7 @@ import {
 } from '../constants/dev-environment';
 import { getVersionList, readEnvironmentData } from './dev-environment-core';
 import type { AppInfo, ComponentConfig, InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
+import { ensureDockerInstalled } from './dev-environment-lando';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
@@ -66,6 +68,14 @@ export async function handleCLIException( exception: Error, trackKey?: string, t
 		debug( exception );
 	}
 }
+
+export const validateDependencies = async () => {
+	try {
+		await ensureDockerInstalled();
+	} catch ( exception ) {
+		exit.withError( exception.message );
+	}
+};
 
 export function getEnvironmentName( options: EnvironmentNameOptions ): string {
 	if ( options.slug ) {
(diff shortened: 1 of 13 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -17,6 +17,7 @@ import os from 'os';
 /**
  * Internal dependencies
  */
+import * as exit from 'lib/cli/exit';
```
sjinks, line 20: Out of curiosity, will this work?

```js
import exit from 'lib/cli/exit';
```

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -294,3 +294,17 @@ async function ensureNoOrphantProxyContainer( lando ) {
...
 	await proxyContainer.remove();
 }
+
+export async function ensureDockerInstalled() {
+	const lando = new Lando( getLandoConfig() );
+	await lando.bootstrap();
+
+	lando.log.verbose( 'docker-engine exists: %s', lando.engine.dockerInstalled );
+	if ( lando.engine.dockerInstalled === false ) {
+		throw Error( 'docker could not be located!' );
```
sjinks: TIL that `throw Error()` and `throw new Error()` [are equivalent](https://tc39.es/ecma262/#sec-error-constructor) 😀

sjinks (approved):

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -294,3 +294,17 @@ async function ensureNoOrphantProxyContainer( lando ) {
 
 	await proxyContainer.remove();
 }
+
+export async function ensureDockerInstalled() {
```
jesse-greathouse: I see this function name as an anti-pattern. I will explain.

The way I see it is that this is a distributed validation pattern, in `validateDependencies`, which is going to fire any number of validation methods respective to each dependency. Please correct me if I'm wrong.

This being the case, I think all individual validation methods, should be named similarly to indicate they are aligned with this validation purpose; e.g. instead of `ensureDockerInstalled` it should be more like `validateDocker` or `validateDockerDependency` something along those lines.

jesse-greathouse (changes requested):

src/lib/dev-environment/dev-environment-lando.js
rinatkhaziev: What do you think about checking for both first and then throwing a combined error instead? 

Also providing links to Docker site may be a nice touch and spare extra click.

As for the error message if on mac or windows "Please install Docker Desktop", if on Linux: Please install Docker Engine and Docker Compose.

src/lib/dev-environment/dev-environment-cli.js
pschoffer: Nope didn't work (it is still kinda mystery how these different JS module types work to me :)).

src/lib/dev-environment/dev-environment-lando.js
pschoffer: Good point. renamed it to `validateDockerInstalled`

src/lib/dev-environment/dev-environment-lando.js
pschoffer: Good idea with the docs. I added those.

I was trying to avoid doing platform checks there. So I opted to not do different text per platform if you don't feel strongly about that?

Same for having the message composed as the scenario when you have docker but not docker-compose is kinda rare (linux only), and it seems not for long as they are rolling out Docker Desktop to linux :(

rinatkhaziev: > I was trying to avoid doing platform checks there. So I opted to not do different text per platform if you don't feel strongly about that?
Makes sense!

jesse-greathouse (changes requested):
The code looks great.

I did test it, and I came up with some mixed results.

Here are my findings:

The new code succeeds in several instances:

- When docker is installed, everything is perfect. ✅ 
- When docker is not installed the following dev-env directives work as expected:
  - `vip dev-env create` ✅ 
  - `vip dev-env start` ✅ 

Screen Shot 2022-05-10 at 11 54 44 AM

The following directives did not get the desired result:
  - `vip dev-env info` 🚫 
  - `vip dev-env stop` 🚫 
  - `vip dev-env destroy` 🚫 
  
When I ran these directives, my output looked like this:

Screen Shot 2022-05-10 at 11 53 41 AM

I think this is pretty close, but I would like to see the docker installation prompt happen on every dev-env directive.

pschoffer:
Good point. Initially, I added it only to `create` and `start`. Now I added it to all of them.

jesse-greathouse (approved):
If this was any more Nice, it would be a city in France. Well done.

PR #1023, opened 2022-05-03
[dev-env] Skip the trunk from the prompt about the latest available WordPress versions

rinatkhaziev:
## Description

`trunk`  shouldn't show up as the latest version. Technically it's always the latest code, however, we only want to present actual versions (E.g. 6.0, 5.9.3) when displaying the upgrade notice.

1. Check out PR.
1. Run `npm run build && npm link`
1. Run `vip dev-env start ` on an older environment
1. Verify that trunk is not showing up as the most recent version and is the last

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -467,10 +467,10 @@ async function updateWordPressImage( slug ) {
 	}
 
 	// sort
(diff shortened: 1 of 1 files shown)


sjinks (approved):

pschoffer:
Merging so that it gets included in the release.

PR #1025, opened 2022-05-05
[dev-env] Optimize tracking

pschoffer:
## Description

Two track related fixes:
* Renames `processingTime` => `processing_time` to avoid event being dropped for camelCase
* JSON stringify complex object to record those too

## Steps to Test

Run with `--debug @automattic/vip:analytics:clients:tracks`

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -434,7 +434,7 @@ export function getEnvTrackingInfo( slug: string ): any {
 		for ( const key of Object.keys( envData ) ) {
 			// track doesnt like camelCase
 			const snakeCasedKey = key.replace( /[A-Z]/g, letter => `_${ letter.toLowerCase() }` );
-			const value = typeof envData[ key ] === 'object' ? JSON.stringify( envData[ key ] ) : envData[ key ];
+			const value = DEV_ENVIRONMENT_COMPONENTS.includes( key ) ? JSON.stringify( envData[ key ] ) : envData[ key ];
 
 			result[ snakeCasedKey ] = value;
 		}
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -434,7 +434,9 @@ export function getEnvTrackingInfo( slug: string ): any {
 		for ( const key of Object.keys( envData ) ) {
 			// track doesnt like camelCase
 			const snakeCasedKey = key.replace( /[A-Z]/g, letter => `_${ letter.toLowerCase() }` );
-			result[ snakeCasedKey ] = envData[ key ];
+			const value = typeof envData[ key ] === 'object' ? JSON.stringify( envData[ key ] ) : envData[ key ];
```
jesse-greathouse: I think it's very weird that at this point the `envData[ key ]` could be a string or an object. I guess it represents a data type, in our environment data graph, which could be a string or an object. The polymorphic nature of that gives me pause, because JavaScript doesn't have super great polymorphic tools, and JSON doesn't have great ways to indicate that a node could be polymorphic, like XML does. I guess it's not a problem but it's something that we should think about.

jesse-greathouse (approved):

src/lib/dev-environment/dev-environment-cli.js
sjinks: @pschoffer `typeof null === 'object'`, `JSON.stringify(null) === 'null'`.

Is this the intended behavior?

rinatkhaziev: I think we should enumerate the properties we care about rather than trying to handle it abstractly in this fashion

pschoffer: @jesse-greathouse yeah. It is typed though, so you get all that nice doc completion (e.g. data.wordpress.type = 'image'). 

@sjinks I think that's fine. But I am also fine with adding a null check there

pschoffer: Oh right we already have a list of those components so I did that. Looks better. Thanks!

PR #1028, opened 2022-05-10
PIE-3135 - Re-calculate the fileMeta if file gets changed by the searchAndReplace

djalmaaraujo:
## Description

This PR fixes a BUG when importing a SQL file using the search and replace method. When using `searchAndReplace`, the file size differs after the fact, and we were using the same fileSize as the original file. This works for most cases, but when the search and replace makes a more significant difference in the file size, the upload part fails because AWS S3 always expects the original file size.

We now re-calculate the complete fileMeta, if the user uses the `searchAndReplace` method.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-import-sql.js @SITEID.develop test-file.sql  --search-replace="https://old-url,htts://new-url --search-replace="https://another-search-and-replace-url,htts://new-search-and-replace-new-url`
4. The import should work always


brunobasto (approved):
Tested and solves the issue. Thank you

PR #1035, opened 2022-05-13
New config module

mjangda:
Instead of requiring a generated config file, just read the raw local and production files instead. For dev, we'll default to the local one. For published packages, we'll npmignore the local version so it uses the publish config instead.

This reduces the risk of things going wrong with our hacky config generation during npm lifecycle scripts (which has caused problems before; see v2.11.0 + v2.11.1).

Props @andrea-sdl for the idea

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `npm t` and check that the tests are running fine.
1. Run `DEBUG=* ./dist/bin/vip-whoami.js` (should see analytics being tracked with `_dev_` prefix).
1. Rename the `config/config.local.json` file to something else.
1. Run `DEBUG=* ./dist/bin/vip-whoami.js` (should see analytics being tracked without the `_dev_` prefix).

There should also be a debug message noting which config file is loaded.

After this test the creation of the pack
1. Restore the `config/config.local.json` and run `npm pack` with npm 6 (you can run `nvm use 14`)
1. check that the content of the tar includes the `config/config.publish.json` but not the `config/config.local.json` with the command `tar -tf *.tgz | grep  'config.*json'`
1. Delete the `tgz` file 
1. run `npm pack` with npm 8 (`nvm use 16`)
1. check again that the content of the tar includes the `config/config.publish.json` but not the `config/config.local.json` with the command `tar -tf *.tgz | grep  'config.*json'`


andrea-sdl:
@mjangda should be ready, added static tests and changed the description

mjangda:
Nice! LGTM (can't approve since it's my PR :D)

PR #1038, opened 2022-05-18
[dev-env] Adds domain validation during sql import

pschoffer:
## Description

This change adds a validation that detects if siteURL is pointing to lando subdomain during DB import. 

If the siteUrl will be kept pointing to a live environment, WordPress will redirect there resulting to local dev-env not actually being used (which is quite confusing). 

## Steps to Test
Should fail without search and replace
```
npm run build && ./dist/bin/vip-dev-env-import-sql.js ~/Downloads/data 
```

Should work after search and replace
```
npm run build && ./dist/bin/vip-dev-env-import-sql.js ~/Downloads/data  --search-replace="pschoffer-develop.go-vip.net,vip-local.vipdev.lndo.site"
```

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -251,7 +251,6 @@ const checks: Checks = {
 		matcher: "'(siteurl|home)',\\s?'(.*?)'",
 		matchHandler: ( lineNumber, results, expectedDomain ) => {
 			const foundDomain = results[ 2 ].replace( /https?:\/\//, '' );
-			console.log('Check', foundDomain, expectedDomain);
 			if ( ! foundDomain.trim() ) {
 				return { falsePositive: true };
 			}
@@ -403,7 +402,7 @@ const perLineValidations = ( line: string, options: ValidationOptions = DEFAULT_
 	checkForTableName( line );
 
 	const checkKeys = Object.keys( checks ).filter( checkItem => ! options.skipChecks.includes( checkItem ) );
-	for (const checkKey of checkKeys) {
+	for ( const checkKey of checkKeys ) {
 		const check: CheckType = checks[ checkKey ];
 		const results = line.match( check.matcher );
 		const extraCheckParams = options.extraCheckParams[ checkKey ];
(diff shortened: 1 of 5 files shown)


src/lib/validations/sql.js
```diff
@@ -215,6 +215,15 @@ const checks: Checks = {
 		excerpt: 'Siteurl/home options',
 		recommendation: '',
 	},
+	siteHomeUrlLando: {
+		matcher: "'(siteurl|home)',\\s?'([^']+vipdev.lndo.site)'",
```
sjinks:
```suggestion
		matcher: "'(siteurl|home)',\\s*'([^']+vipdev\\.lndo\\.site)'",
```

rinatkhaziev: @sjinks why the double escaping in `\\.`?

sjinks: @rinatkhaziev because this regexp is a string. We need smth like

```js
/'(siteurl|home)',\s*'([^']+vipdev\.lndo\.site)'/
```

But `"\."` is `"."`, and to get `\.`, we need to escape that slash as well.

sjinks: Screenshot 2022-05-19 at 01 23 04

rinatkhaziev: ah gotcha.

src/lib/validations/sql.js
```diff
@@ -215,6 +215,15 @@ const checks: Checks = {
...
+		matchHandler: ( lineNumber, results ) => results[ 2 ],
+		outputFormatter: requiredCheckFormatter,
+		results: [],
+		message: 'Siteurl/home options pointing to *.vipdev.lndo.site domain',
+		excerpt: 'Siteurl/home options pointing to lando domain',
+		recommendation: 'Use search-replace to change environment\'s domain',
```
rinatkhaziev, line 275: @pschoffer how is the recommendation exposed? Should we include the example here like `--search-replace=prod-domain,local-dev-env.vipdev.lndo.site`? Since we likely know both the prod domain (assuming users use `@ID` notation and the the dev env target domain

pschoffer: Yeah I tried to get something like that, but the issue is that this validation mechanism is shared with regular SQL import and I would have to build a bit of custom outputFormatter for  just that one. 

I'll think about it some more.

pschoffer: How's this? :)

sjinks (approved):

PR #1040, opened 2022-05-20
[dev-env] track dev-env start time in seconds

pschoffer:
## Description

We used to track processing_time in milliseconds, which is a bit harder to read for a human (I am told :) ). This change switches to seconds. 

## Steps to Test

```
 npm run build && ./dist/bin/vip-dev-env-start.js --debug @automattic/vip:analytics:clients:tracks
```

### DIFF
--- a/src/bin/vip-dev-env-start.js
+++ b/src/bin/vip-dev-env-start.js
@@ -68,7 +68,7 @@ command()
 
 			await startEnvironment( slug, options );
 
-			const processingTime = new Date() - startProcessing;
+			const processingTime = Math.ceil( ( new Date() - startProcessing ) / 1000 ); // in secods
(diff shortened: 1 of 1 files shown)


src/bin/vip-dev-env-start.js
```diff
@@ -68,7 +68,7 @@ command()
 
 			await startEnvironment( slug, options );
 
-			const processingTime = new Date() - startProcessing;
+			const processingTime = Math.ceil( ( new Date() - startProcessing ) / 1000 ); // in secods
```
sjinks:
```suggestion
			const processingTime = Math.ceil( ( new Date() - startProcessing ) / 1000 ); // in seconds
```

PR #1041, opened 2022-05-23
[dev-env] Fix healthchecks

pschoffer:
## Description

The DB container usually needs a few seconds to startup. However, dev-env seems to wait much longer (30s).

The reason is that although we check the state with an interval of 1s, docker-compose only checks every 30s (by default). This change changes the docker-compose interval to 2s. The default docker-compose retries are just 3 so I bumped that too with the much more frequent checks. 

## Steps to Test

1) Create (or update) dev-env
2) `npm run build && ./dist/bin/vip-dev-env-start.js --slug test --debug @automattic/vip:bin:dev-environment`

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -70,6 +70,8 @@ services:
         - ":3306"
       healthcheck:
         test: 'mysql -uroot --silent --execute "SHOW DATABASES;"'
+        interval: 2s
+        retries: 60
       environment:
         MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: 'true'
       volumes:
@@ -108,6 +110,8 @@ services:
         - ":9200"
       healthcheck:
         test: "curl --noproxy '*' -XGET localhost:9200"
+        interval: 2s
+        retries: 60
       volumes:
         - search_data:/usr/share/elasticsearch/data
     volumes:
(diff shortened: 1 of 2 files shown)


sjinks:
I am afraid that this might make things slower on a Mac/Windows.

What if we try the [`start_period`](https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck) ([description](https://docs.docker.com/engine/reference/builder/#healthcheck)) first? The less the check interval is, the higher the CPU utilization goes :-(

rinatkhaziev:
I was gonna say what @sjinks said. As far as my understanding goes it will be run regardless (i.e. not just once to set the healthy state but periodically (in this PR every two seconds) so that's a lot of wasted CPU cycles. 

One alternative thing that comes to mind is to move healthcheck inside of a bash loop, so that we can have healthcheck interval higher but a single run of the healthcheck.

@sjinks it seems that start_period is only available in compose file version 3.4 and above, so not sure it'll work.

sjinks:
@rinatkhaziev what is the minimum version of docker-compose which we support?

3.4 is Docker Engine 17.09.0, that was December 2017...

rinatkhaziev:
@sjinks well, given this https://github.com/Automattic/lando-cli/pull/4/files, the **minimum** is v1, but I don't believe we even tested on 3

sjinks:
@rinatkhaziev I think we misunderstood each other 
You are probably talking about the [docker-compose release version](https://github.com/docker/compose/releases/), I am talking about the [compose file format version](https://docs.docker.com/compose/compose-file/compose-versioning/)

rinatkhaziev:
oh yeah, my bad!

pschoffer:
Hmm yeah, that I haven't thought of (I thought it would stop after healthy).

> One alternative thing that comes to mind is to move healthcheck inside of a bash loop, so that we can have healthcheck interval higher but a single run of the healthcheck

This I wanted to avoid as using docker is giving you this integration with docker tools + we don't need to know in setup.sh script which containers we have and should check for.

But then again it is only a DB and and search at this point and we know if search is enabled in that script already.

pschoffer:
Closing in favour of  https://github.com/Automattic/vip/pull/1044

PR #1042, opened 2022-05-25
[dev-env] Fix search popup during wizard

pschoffer:
## Description

We were not parsing if the current config has Search enabled or not. This resulted in default being always false even if the current environment has search enabled during update.

```
$ vip dev-env  update

✔ PHP version to use · default
✔ WordPress - Which version would you like · 6.0-RC2
✔ How would you like to source vip-go-mu-plugins · local
✔       What is a path to your local vip-go-mu-plugins · /home/pavel/git/automattic/vip-go-mu-plugins
✔ How would you like to source site-code · local
✔       What is a path to your local site-code · /home/pavel/git/automattic/vip-go-skeleton
? Enable Enterprise Search? (y/N) ‣ false
```

After this change the data is reflected and true is preselected if the dev env already has search running.

` npm run build && ./dist/bin/vip-dev-env-update.js`

### DIFF
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -70,6 +70,7 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			statsd: currentInstanceData.statsd,
 			phpmyadmin: currentInstanceData.phpmyadmin,
 			xdebug: currentInstanceData.xdebug,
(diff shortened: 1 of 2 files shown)


sjinks (approved):
LGTM

rinatkhaziev (approved):
I knew I was forgetting something :)

PR #1044, opened 2022-05-30
[dev-env] Fix healthchecks No. 2

pschoffer:
## Description

The DB container usually needs a few seconds to startup. However, dev-env seems to wait much longer (30s).

The reason is that although we check the state with an interval of 1s, docker-compose only checks every 30s (by default). This change switches from using docker-compose checks to us calling it directly on the container.

## Steps to Test

1) Create (or update) dev-env
2) `npm run build && ./dist/bin/vip-dev-env-start.js --slug test --debug @automattic/vip:bin:dev-environment`

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -101,13 +101,40 @@ function addHooks( app: App, lando: Lando ) {
 	app.events.on( 'post-start', 1, () => healthcheckHook( app, lando ) );
 }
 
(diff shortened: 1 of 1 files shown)


pschoffer:
@rinatkhaziev / @sjinks thanks for pushing me for a better solution. I think this is it.

(we can't do the checks in setup script easily btw... because that script runs on `php` container, so it is more difficult to test other containers)

sjinks (approved):
LGTM

PR #1046, opened 2022-05-31
Add Cache Purge Command

luiztiago:
## Description

Customers could use the CLI to purge cache as a similar command used on WP sites.

## Steps to Test

1. Check out PR.
1. Run `npm run build` and `npm link`
1. Run `node ./dist/bin/vip @4716 cache purge-url https://luiztiago-test-node.go-vip.net/`
1. Verify `Purged URLs` success message

### DIFF
--- a/__tests__/bin/vip-cache-purge.js
+++ b/__tests__/bin/vip-cache-purge.js
@@ -56,11 +56,7 @@ describe( 'cachePurgeCommand()', () => {
 		await cachePurgeCommand( args, opts );
 
 		expect( console.log ).toHaveBeenCalledTimes( 1 );
-		expect( console.log ).toHaveBeenCalledWith(
-			/* eslint-disable indent */
-	'- Purged URLs: url'
-				/* eslint-enable indent */
-		);
+		expect( console.log ).toHaveBeenCalledWith( '- Purged URLs: url' );
 
 		const trackingParams = {
 			app_id: 1,
@@ -91,7 +87,7 @@ describe( 'cachePurgeCommand()', () => {
 		};
 
 		expect( exit.withError ).toHaveBeenCalledTimes( 1 );
-		expect( exit.withError ).toHaveBeenCalledWith( 'Failed to purge cache object error: Something went wrong :(' );
+		expect( exit.withError ).toHaveBeenCalledWith( 'Failed to purge cache object: Something went wrong :(' );
 
 		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'cache_purge_command_execute', trackingParams );
--- a/__tests__/bin/vip-cache-purge.js
+++ b/__tests__/bin/vip-cache-purge.js
@@ -29,7 +29,7 @@ jest.mock( 'lib/tracker', () => ( {
 } ) );
 
 describe( 'cachePurgeCommand()', () => {
-	const args = { urls: [ 'url' ] };
+	const args = [ 'url' ];
 	const opts = {
 		app: {
 			id: 1,
@@ -56,12 +56,13 @@ describe( 'cachePurgeCommand()', () => {
 		await cachePurgeCommand( args, opts );
 
 		expect( console.log ).toHaveBeenCalledTimes( 1 );
-		expect( console.log ).toHaveBeenCalledWith( '- Purged URLs: url' );
+		expect( console.log ).toHaveBeenCalledWith( '- Purged URL: url' );
 
 		const trackingParams = {
 			app_id: 1,
 			command: 'vip cache purge',
 			env_id: 3,
+			from_file: false,
 		};
 
 		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
@@ -84,6 +85,7 @@ describe( 'cachePurgeCommand()', () => {
 			app_id: 1,
 			command: 'vip cache purge',
 			env_id: 3,
+			from_file: false,
 		};
 
 		expect( exit.withError ).toHaveBeenCalledTimes( 1 );
@@ -92,5 +94,27 @@ describe( 'cachePurgeCommand()', () => {
 		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'cache_purge_command_execute', trackingParams );
 		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'cache_purge_command_error', { ...trackingParams, error: 'Something went wrong :(' } );
+	} );
+
+	it( 'should handle error if urls is empty', async () => {
+		const promise = cachePurgeCommand( [], opts );
+
+		await expect( promise ).rejects.toBe( 'EXIT CACHE PURGE WITH ERROR' );
+
+		expect( console.log ).toHaveBeenCalledTimes( 0 );
+
+		const trackingParams = {
+			app_id: 1,
+			command: 'vip cache purge',
+			env_id: 3,
+			from_file: false,
+		};
+
+		expect( exit.withError ).toHaveBeenCalledTimes( 1 );
+		expect( exit.withError ).toHaveBeenCalledWith( 'You need at least an URL to purge cache' );
+
+		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'cache_purge_command_execute', trackingParams );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'cache_purge_command_error', trackingParams );
 	} );
 } );
--- a/__tests__/bin/vip-cache-purge.js
+++ b/__tests__/bin/vip-cache-purge.js
@@ -1,0 +1,100 @@
+/**
+ * Internal dependencies
+ */
+import * as tracker from 'lib/tracker';
+import * as exit from 'lib/cli/exit';
+import purgeCacheLib from 'lib/api/cache-purge';
+import { cachePurgeCommand } from '../../src/bin/vip-cache-purge';
+
+jest.spyOn( console, 'log' ).mockImplementation( () => {} );
+jest.spyOn( exit, 'withError' ).mockImplementation( () => {
+	throw 'EXIT CACHE PURGE WITH ERROR'; // Prevent actually exiting like the lib does
+} );
+
+jest.mock( 'lib/cli/command', () => {
+	const commandMock = {
+		argv: () => commandMock,
+		examples: () => commandMock,
+		option: () => commandMock,
+	};
+	return jest.fn( () => commandMock );
+} );
+
+jest.mock( 'lib/api/cache-purge', () => ( {
+	purgeCache: jest.fn(),
+} ) );
+
+jest.mock( 'lib/tracker', () => ( {
+	trackEvent: jest.fn(),
+} ) );
+
+describe( 'cachePurgeCommand()', () => {
+	const args = { urls: [ 'url' ] };
+	const opts = {
+		app: {
+			id: 1,
+			organization: {
+				id: 2,
+			},
+		},
+		env: {
+			id: 3,
+			type: 'develop',
+		},
+	};
+
+	beforeEach( jest.clearAllMocks );
+
+	it( 'should return success when object is purged', async () => {
+		purgeCacheLib.purgeCache.mockImplementation( () => {
+			return {
+				success: true,
+				urls: [ 'url' ],
+			};
+		} );
+
+		await cachePurgeCommand( args, opts );
+
+		expect( console.log ).toHaveBeenCalledTimes( 1 );
+		expect( console.log ).toHaveBeenCalledWith(
+			/* eslint-disable indent */
+	'- Purged URLs: url'
+				/* eslint-enable indent */
+		);
+
+		const trackingParams = {
+			app_id: 1,
+			command: 'vip cache purge',
+			env_id: 3,
+		};
+
+		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'cache_purge_command_execute', trackingParams );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'cache_purge_command_success', trackingParams );
+	} );
+
+	it( 'should handle error when request fails', async () => {
+		purgeCacheLib.purgeCache.mockImplementation( () => {
+			throw new Error( 'Something went wrong :(' );
+		} );
+
+		const promise = cachePurgeCommand( args, opts );
+
+		await expect( promise ).rejects.toBe( 'EXIT CACHE PURGE WITH ERROR' );
+
+		expect( console.log ).toHaveBeenCalledTimes( 0 );
+
+		const trackingParams = {
+			app_id: 1,
+			command: 'vip cache purge',
+			env_id: 3,
+		};
+
+		expect( exit.withError ).toHaveBeenCalledTimes( 1 );
+		expect( exit.withError ).toHaveBeenCalledWith( 'Failed to purge cache object error: Something went wrong :(' );
+
+		expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 1, 'cache_purge_command_execute', trackingParams );
+		expect( tracker.trackEvent ).toHaveBeenNthCalledWith( 2, 'cache_purge_command_error', { ...trackingParams, error: 'Something went wrong :(' } );
+	} );
+} );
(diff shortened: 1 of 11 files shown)


__tests__/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,100 @@
...
+		} );
+
+		await cachePurgeCommand( args, opts );
+
+		expect( console.log ).toHaveBeenCalledTimes( 1 );
+		expect( console.log ).toHaveBeenCalledWith(
+			/* eslint-disable indent */
+	'- Purged URLs: url'
+				/* eslint-enable indent */
+		);
```
chriszarate:
```suggestion
		expect( console.log ).toHaveBeenCalledWith( '- Purged URLs: url' );
```

npm-shrinkwrap.json
```diff
@@ -10146,6 +10146,14 @@
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.1.tgz",
           "integrity": "sha512-+O9Jct8wf++lXxxFc4hc8LsjaSq0HFzzL7cVsw8pRDIPdjKD2mT4ytDZlLuSBZ4cLKZFXIrMGO7DbQCtMJJMKw=="
         },
+        "cli-table": {
```
chriszarate: hmm why did this file change?

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+ * Internal dependencies
+ */
+import { appQuery, purgeCache } from 'lib/api/cache-purge';
+import command from 'lib/cli/command';
+import { trackEvent } from 'lib/tracker';
+import * as exit from 'lib/cli/exit';
+
+const examples = [
+	{
+		usage: 'vip @123.production cache purge <URL1> <URL2>',
```
chriszarate: Is this how we want to support purging multiple URLs? Or would it be better to allow passing in multiple URLs via `stdin` or a text file?

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+		description: 'Clear cache for a Node.js URL',
```
chriszarate: This works for WP sites too, no?

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+
+	await trackEvent( 'cache_purge_command_execute', trackingParams );
+
+	let purgeCacheObject = {};
+	try {
+		purgeCacheObject = await purgeCache( opt.app.id, opt.env.id, urls );
+	} catch ( err ) {
+		await trackEvent( 'cache_purge_command_error', { ...trackingParams, error: err.message } );
+
+		exit.withError( `Failed to purge cache object error: ${ err.message }` );
```
chriszarate:
```suggestion
		exit.withError( `Failed to purge cache object: ${ err.message }` );
```

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	await trackEvent( 'cache_purge_command_success', trackingParams );
+
+	const output = [
+		`- Purged URLs: ${ purgeCacheObject.urls.join( ',' ) }`,
+	];
+
+	console.log( output.join( '\n' ) );
```
chriszarate: this will make for much better output in the terminal

```suggestion
	purgeCacheObject.urls.forEach( ( url ) => {
		console.log( `- Purged URL: ${ url }` );
	} );
```

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+command( {
+	appContext: true,
+	appQuery,
+	envContext: true,
+	requiredArgs: 1,
+	wildcardCommand: true,
```
chriszarate: i forget what this means. does it allow the argument to take any form?

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,62 @@
...
+		app_id: opt.app.id,
+		command: 'vip cache purge',
+		env_id: opt.env.id,
+	};
+
+	await trackEvent( 'cache_purge_command_execute', trackingParams );
+
+	let purgeCacheObject = {};
+	try {
+		purgeCacheObject = await purgeCache( opt.app.id, opt.env.id, urls );
```
chriszarate: should we do any inspection of the input?

src/lib/api/cache-purge.js
```diff
@@ -0,0 +1,65 @@
...
+		appId,
+		envId,
+		urls,
+	};
+
+	const response = await api.mutate( { mutation, variables } );
+
+	const { data: { purgeCacheObject } } = response;
+	if ( ! purgeCacheObject ) {
+		throw new Error( 'Something went wrong :(' );
```
chriszarate: Can we be more specific here? Will there be an error returned? Is this even necessary or can we handle the error in the calling code?

src/bin/vip-cache-purge.js
luiztiago: @chriszarate  This is used to accept args when using multiple URLs

src/bin/vip-cache-purge.js
luiztiago: @chriszarate Do you have an example with stdin or text file working? I've tried a few things but with no success

chriszarate: here's an example using a file (via an option):
https://github.com/Automattic/vip/blob/5372034e86d3aba96358286d6eef01a5c879432f/src/bin/vip-config-envvar-set.js#L67

this is all that's necessary, since someone can use this option to pass in `stdin`, e.g.:

`echo "my value" | vip @4092 config envvar set MYVAR --from-file=/dev/stdin`

npm-shrinkwrap.json
ingeniumed: For me, this and the colours bit below get removed when I run `npm install` so not sure why this got added here. I repeated that 2-3 times and it got removed for me

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+import { appQuery, purgeCache } from 'lib/api/cache-purge';
+import command from 'lib/cli/command';
+import { trackEvent } from 'lib/tracker';
+import { readFromFile } from 'lib/read-file';
+import * as exit from 'lib/cli/exit';
+
+const examples = [
+	{
+		usage: 'vip @123.production cache purge <URL>',
```
ingeniumed: Should this be URLs, based on the design session and the list below in the command?

src/bin/vip-cache-purge.js
```diff
@@ -0,0 +1,76 @@
...
+		const value = await readFromFile( opt.fromFile );
+		if ( value ) {
+			urls = value.split( '\n' ).map( url => url.trim() );
+		}
+	}
+
+	if ( ! urls.length ) {
+		await trackEvent( 'cache_purge_command_error', trackingParams );
+
+		exit.withError( 'You need at least an URL to purge cache' );
```
ingeniumed: Could I suggest changing this to `You need at least one URL to purge the cache`?

src/lib/envvar/read-file.js
```diff
@@ -6,25 +6,16 @@
...
 /**
  * Internal dependencies
  */
-import * as exit from 'lib/cli/exit';
 import { debug } from 'lib/envvar/logging';
+import { readFromFile } from '../read-file';
 
 export async function readVariableFromFile( path ): Promise<string> {
 	debug( `Loading variable value from file ${ JSON.stringify( path ) }` );
+	const data = await readFromFile( path );
```
ingeniumed: Could return it inline here

src/bin/vip-cache-purge.js
chriszarate: I'm open to discussion, but I think I would prefer a single entrypoint to purging multiple URLs. The `--from-file` option makes the most sense to me. If I'm bringing a list of URLs to this command, it makes sense that I'd have them in a newline-separated list instead of a space-separated list.

So I'd argue for keeping this usage the way it is and only allowing one URL:

```
vip @123.production cache purge <URL>
```

For multiple URLs, users can use `--from-file` or just run the command multiple times:

```
vip @123.production cache purge <URL1>
vip @123.production cache purge <URL2>
```

chriszarate: I think I'd also advocate for renaming the command to `purge-url`. I can envision adding other commands in the future; perhaps purging via a `BAN` directive (pattern) or potentially `purge-all` or purging other types other types of cache. So being specific here will help us avoid ambiguity in the future.

luiztiago: @chriszarate I agree, I'm renaming it and removing the list of URLs using args.

src/lib/api/cache-purge.js
luiztiago: @chriszarate If the command throws an error, it will be printed on the CLI command, as the domain exception. This error message is just a "backup" if the purgeCacheObject is null and the mutation is not throwing an error.

chriszarate: Understood, it's just not particularly useful as an error message. If you removed this check/throw, it would raise an exception whenever you tried to destructure from `null`, which is probably just as useful, if not more so.

If there are situations where we think we might return `null`, we should probably convert them to useful error messages in the API.

luiztiago: @chriszarate I agree 👍

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+ */
+import { appQuery, purgeCache } from 'lib/api/cache-purge';
+import command from 'lib/cli/command';
+import { trackEvent } from 'lib/tracker';
+import { readFromFile } from 'lib/read-file';
+import * as exit from 'lib/cli/exit';
+
+const examples = [
+	{
+		usage: 'vip @123.production cache purge-url <URL>',
```
chriszarate: We generally don't add the `@` shorthand to usage commands since it is provided separately by the echoed options help text

```suggestion
		usage: 'vip cache purge-url <URL>',
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+		description: 'Purge cache for a URL',
+	},
+	{
+		usage: 'vip @123.production cache purge-url --from-file=/dev/vip/urls.txt',
```
chriszarate:
```suggestion
		usage: 'vip cache purge-url --from-file=/dev/vip/urls.txt',
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+		console.log( `- Purged URL: ${ url }` );
+	} );
+}
+
+command( {
+	appContext: true,
+	appQuery,
+	envContext: true,
+	wildcardCommand: true,
+	usage: 'vip @123.production cache purge-url <URL>',
```
chriszarate:
```suggestion
	usage: 'vip cache purge-url <URL>',
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+} )
+	.option( 'from-file', 'Read URLs from file (useful for multiline input)' )
```
chriszarate:
```suggestion
	.option( 'from-file', 'Read URLs from file (useful to purge multiple URLs)' )
```

src/bin/vip-cache.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Internal dependencies
+ */
+import command from 'lib/cli/command';
+
+command( {
+	requiredArgs: 2,
+} )
+	.command( 'purge-url', 'Purge cache' )
```
chriszarate:
```suggestion
	.command( 'purge-url', 'Purge page cache' )
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+import { appQuery, purgeCache } from 'lib/api/cache-purge';
+import command from 'lib/cli/command';
+import { trackEvent } from 'lib/tracker';
+import { readFromFile } from 'lib/read-file';
+import * as exit from 'lib/cli/exit';
+
+const examples = [
+	{
+		usage: 'vip cache purge-url <URL>',
+		description: 'Purge cache for a URL',
```
chriszarate: Today's session taught me that we should be very specific with our terms :)

```suggestion
		description: 'Purge a URL from page cache',
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+	},
+	{
+		usage: 'vip cache purge-url --from-file=/dev/vip/urls.txt',
+		description: 'Purge cache for multiple URLs',
```
chriszarate:
```suggestion
		description: 'Purge multiple URLs from page cache',
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+
+export async function cachePurgeCommand( urls = [], opt = {} ): void {
+	const trackingParams = {
+		app_id: opt.app.id,
+		command: 'vip cache purge-url',
+		env_id: opt.env.id,
+		from_file: !! opt.fromFile,
+	};
+
+	await trackEvent( 'cache_purge_command_execute', trackingParams );
```
chriszarate:
```suggestion
	await trackEvent( 'cache_purge_url_command_execute', trackingParams );
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+
+	if ( opt.fromFile ) {
+		const value = await readFromFile( opt.fromFile );
+		if ( value ) {
+			urls = value.split( '\n' ).map( url => url.trim() );
+		}
+	}
+
+	if ( ! urls.length ) {
+		await trackEvent( 'cache_purge_command_error', trackingParams );
```
chriszarate:
```suggestion
		await trackEvent( 'cache_purge_url_command_error', { ...trackingParams, error: 'Missing URL' } );
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+
+		exit.withError( 'You need at least one URL to purge the cache' );
+	}
+
+	let purgeCacheObject = {};
+	try {
+		purgeCacheObject = await purgeCache( opt.app.id, opt.env.id, urls );
+	} catch ( err ) {
+		await trackEvent( 'cache_purge_command_error', { ...trackingParams, error: err.message } );
```
chriszarate:
```suggestion
		await trackEvent( 'cache_purge_url_command_error', { ...trackingParams, error: err.message } );
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+		const value = await readFromFile( opt.fromFile );
+		if ( value ) {
+			urls = value.split( '\n' ).map( url => url.trim() );
+		}
+	}
+
+	if ( ! urls.length ) {
+		await trackEvent( 'cache_purge_command_error', trackingParams );
+
+		exit.withError( 'You need at least one URL to purge the cache' );
```
chriszarate:
```suggestion
		exit.withError( 'Please supply at least one URL.' );
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	let purgeCacheObject = {};
+	try {
+		purgeCacheObject = await purgeCache( opt.app.id, opt.env.id, urls );
+	} catch ( err ) {
+		await trackEvent( 'cache_purge_command_error', { ...trackingParams, error: err.message } );
+
+		exit.withError( `Failed to purge cache object: ${ err.message }` );
```
chriszarate:
```suggestion
		exit.withError( `Failed to purge URL(s) from page cache: ${ err.message }` );
```

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	await trackEvent( 'cache_purge_command_success', trackingParams );
```
chriszarate:
```suggestion
	await trackEvent( 'cache_purge_url_command_success', trackingParams );
```

src/bin/vip-cache-purge-url.js
luiztiago: @chriszarate All suggestions are much better, thanks! :)

ingeniumed (approved):

npm-shrinkwrap.json
```diff
@@ -10104,8 +10104,8 @@
       "dev": true
     },
     "lando": {
-      "version": "git+https://github.com/Automattic/lando-cli.git#d4d836c1ece783a2eb4aead589cf5d3889efa1bd",
```
chriszarate: can you revert the changes to this file? should be unnecessary

src/bin/vip-cache-purge-url.js
```diff
@@ -0,0 +1,80 @@
...
+
+	if ( opt.fromFile ) {
+		const value = await readFromFile( opt.fromFile );
+		if ( value ) {
+			urls = value.split( '\n' ).map( url => url.trim() );
+		}
+	}
+
+	if ( ! urls.length ) {
+		await trackEvent( 'cache_purge_url_command_error', trackingParams );
```
chriszarate:
```suggestion
		await trackEvent( 'cache_purge_url_command_error', { ...trackingParams, error: 'No URL provided' } );
```

chriszarate (changes requested):

chriszarate (approved):
YES!!1!

chriszarate:
Need to update tests, otherwise g2g

chriszarate (approved):

PR #1055, opened 2022-06-27
[dev-env] Makes exec attempt to run the task even if env seems to be down

pschoffer:
## Description

The environment is considered up if all its public URLs are accessible from the terminal. In most cases, this is what we want. However, there might be some network issue making the site "look" down while it is actually running. This issue should be discovered and solved.

However, as a workaround, we can provide users the ability to skip the check for the env status so that they can still use cli commands. 

## Steps to Test

Run `npm run build && ./dist/bin/vip-dev-env-exec.js --force  -- wp post list` (it will still fail later if the env isn't started, but it is possibly to run it)

### DIFF
--- a/src/bin/vip-dev-env-exec.js
+++ b/src/bin/vip-dev-env-exec.js
@@ -36,6 +36,7 @@ const examples = [
 
 command( { wildcardCommand: true } )
 	.option( 'slug', 'Custom name of the dev environment' )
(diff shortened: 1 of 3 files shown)


sjinks (approved):

jesse-greathouse (approved):
LGTM

PR #1059, opened 2022-07-08
Update engines to show support for npm > 6

mjangda:
`vip` works fine with npm 6 + 8. This will prevent npm from throwing a warning on install / update.

See BB8-9424

Props @GaryJones

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   "license": "MIT",
   "engines": {
     "node": ">=14.14.0",
-    "npm": "^6 "
+    "npm": ">6"
   },
   "bugs": {
     "url": "https://github.com/Automattic/vip/issues"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
   "license": "MIT",
   "engines": {
     "node": ">=14.14.0",
-    "npm": "^6 "
+    "npm": ">6"
```
GaryJones: Should that be >= to include v6?

andrea-sdl (approved):
LGTM thank you for this change ;)

GaryJones (commented):
Apologies - wrote my comment a couple of days ago but it didn't get submitted!

package.json
mjangda: Yep, good catch!

PR #1060, opened 2022-07-19
2.14.0 prep

pschoffer:
### 2.14.0 (19 Jul 2022)

- #1059 Update engines to show support for npm > 6
- #1058 [dev-env] switch dev-env error during import to warning
- #1057 [dev-env] Update debug instruction example
- #1055 [dev-env] Makes exec attempt to run the task even if env seems to be down


brunobasto (approved):
🚢

PR #1062, opened 2022-07-21
[dev-env] Change wizard wording

pschoffer:
## Description

Dev-env wizzard would ask if:
`Enable Enterprise Search?`
This can be a bit confusing as this option would decide if elasticsearch container would be spinned up or not. If the Enterprise search would be enabled is only controlled by the constants as in regular site.

## Steps to Test

```
$ npm run build && ./dist/bin/vip-dev-env-create.js --slug test
...

✔ WordPress site title · VIP Dev
✔ Multisite (y/N) · false
✔ PHP version to use · default
✔ WordPress - Which version would you like · 6.0
✔ How would you like to source vip-go-mu-plugins · image
✔ How would you like to source site-code · image
? Enable Elasticsearch (needed by Enterprise Search)? (y/N) ‣ false
```

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -205,7 +205,7 @@ export async function promptForArguments( preselectedOptions: InstanceOptions, d
 		instanceData[ component ] = result;
 	}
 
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
cc @yolih

yolih:
This looks great! I can update our "Use Enterprise Search on a VIP Local Development Environment" Docs article as soon as this is merged. Thank you! ❤️

PR #1066, opened 2022-07-27
[dev-env] Support custom location

pschoffer:
## Description

Adds support for `VIP_DEV_ENV_LOCATION` environment variable that lets users set custom location for dev-env.

**Note:** It does not play well with already existing dev-env you need to remove all containers and volumes including proxy container.

## Steps-to-test

1) export VIP_DEV_ENV_LOCATION=~/dev-env
2) npm run build && ./dist/bin/vip-dev-env-create.js


mjangda:
I would say `wontfix` for this unless we have a really compelling reason for adding this. Just adds more complexity and variance and makes supporting the feature harder.

pschoffer:
I am closing this for now

PR #1067, opened 2022-07-29
FORNO-1244: SQL Import: Increase max import size for launched sites to 1GB

t-wright:
## Description

Increases the max file size for DB imports into launched sites from 350MB to 1GB. 

## Steps to Test

1. Check out PR.
1. Run `npm run build`
2. Targeting a **launched** test site, run the following command with an SQL file larger than 1GB: `./dist/bin/vip-import-sql.js @<app>.<env> <path_to_file.sql>`
1. Verify [an error ](https://github.com/Automattic/vip/blob/c027a204b6b7d8a4e23ecd489e7b8b993a10afc0/src/bin/vip-import-sql.js#L144-L150) is thrown
2. Try the same command using a file less than 1GB in size and verify that the error is not thrown

### DIFF
--- a/src/lib/site-import/db-file-import.js
+++ b/src/lib/site-import/db-file-import.js
@@ -9,7 +9,7 @@
 import { GB_IN_BYTES, MB_IN_BYTES } from 'lib/constants/file-size';
 
 export const SQL_IMPORT_FILE_SIZE_LIMIT = 100 * GB_IN_BYTES;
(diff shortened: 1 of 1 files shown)


aagam-shah (approved):
🚀

src/lib/site-import/db-file-import.js
```diff
@@ -9,7 +9,7 @@
 import { GB_IN_BYTES, MB_IN_BYTES } from 'lib/constants/file-size';
```
mjangda:
```suggestion
import { GB_IN_BYTES } from 'lib/constants/file-size';
```

PR #1068, opened 2022-07-29
Software Management: Get settings

pschoffer:
## Description

First software settings command. This one lists settings.

## Steps to Test

```
npm run build && ./dist/bin/vip-settings-software-get.js @2891 --format json
```


WPprodigy (commented):
Looks good. A few notes:

1) Preferably the order in the result table would be: WordPress > PHP > MU. This is the UI order of tabs, and was suggested in the design meeting.

2) Only WP can be "pinned" right now. And even though we get it in the api response, I think we should probably avoid the language. Perhaps flip it around and have a column for "Managed Updates Enabled" that is true only for WP if unpinned.

3) Instead of `current` property, maybe use `version`?

4) In the json output, could we maybe add the `slug` property to the software object? Would help with scripting.

5) IDK if this is enforced / normal or not, but does the CLI have to "hang" after running it and require cntrl-c to get out? Feels a bit strange to not just get the response in the terminal output, esp with like `--format json`. Seems like it would make scripting a bit tricky.  Ignore me if this is standard though.

src/bin/vip-settings.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+/**
+ * @flow
+ * @format
+ */
+
+/**
+ * External dependencies
+ */
```
WPprodigy: I hate these doc blocks 😆 . Wish I could kill them all with fire, but we at least blast away the unused ones :)

rinatkhaziev (dismissed):
Thanks! Tested, I testedm functionally looks good for both PHP and Node sites. I second what Caleb says. I think the command should exit once done rather than wait for user input. Especially since the output gets wiped once you terminate.

The other thing - we already have `config` command, having a `settings` command is going to be confusing so I propose we move it under `config` 

`vip config software <subcommand>`

src/bin/vip-settings.js
rinatkhaziev: @WPprodigy time to drop Calypso from lint settings :)

pschoffer:
Thank you both!

> Preferably the order in the result table would be: WordPress > PHP > MU. This is the UI order of tabs, and was suggested in the design meeting.

Fixed

> Only WP can be "pinned" right now. And even though we get it in the api response, I think we should probably avoid the language. Perhaps flip it around and have a column for "Managed Updates Enabled" that is true only for WP if unpinned.

How about this? I recently read that this sort of columns should not exist and be grouped for better UX.
```
┌────────────┬───────────┬───────────────────────┐
│ name       │ slug      │ version               │
├────────────┼───────────┼───────────────────────┤
│ WordPress  │ wordpress │ 6.0 (managed updates) │
├────────────┼───────────┼───────────────────────┤
│ PHP        │ php       │ 7.4                   │
├────────────┼───────────┼───────────────────────┤
│ MU Plugins │ muplugins │ staging               │
└────────────┴───────────┴───────────────────────┘
```

> Instead of current property, maybe use version?

Fixed

> In the json output, could we maybe add the slug property to the software object? Would help with scripting.

I added slug to all. I would rather not change the columns based on the format. I am fine with introducing some other options to add/remove columns though (regardless of the output format). Any suggestions here?

> IDK if this is enforced / normal or not, but does the CLI have to "hang" after running it and require cntrl-c to get out? Feels a bit strange to not just get the response in the terminal output, esp with like --format json. Seems like it would make scripting a bit tricky. Ignore me if this is standard though.

Yeah this is some weird VIP cli thing. It actually seems to work fine when scripting:
```
./dist/bin/vip-config-software-get.js @2891 --format json | jq
```
Also works fine for other formats I tested.

> The other thing - we already have config command, having a settings command is going to be confusing so I propose we move it under config

Good idea. I moved it there.

rinatkhaziev (approved):
Thanks!

PR #1070, opened 2022-08-08
[dev-env] validate client code - look for specific subdirectories

pschoffer:
## Description

To prevent folks from linking to an incorrect site-code folder we added basic validation that there is `themes` subdirectory at least.

## Steps to Test

```
npm run build && ./dist/bin/vip-dev-env-create.js --slug test --client-code="/tmp"
```

Or type something like that during wizard.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -238,19 +238,50 @@ async function processComponent( component: string, preselectedValue: string, de
 		const resolvedPath = resolvePath( result.dir || '' );
 		result.dir = resolvedPath;
 
-		const isDirectory = resolvedPath && fs.existsSync( resolvedPath ) && fs.lstatSync( resolvedPath ).isDirectory();
-		const isEmpty = isDirectory ? fs.readdirSync( resolvedPath ).length === 0 : true;
+		const { result: isPathValid, message } = validateLocalPath( component, resolvedPath );
 
-		if ( isDirectory && ! isEmpty ) {
+		if ( isPathValid ) {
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
I think we should match the validation in production and list all that is missing. 
For example, I pointed to a copy of my repo that only has a themes folder, I mistakenly point to that folder, and VIP-CLI deems this valid. I spin up. the environment only to realize the site is broken. I get annoyed and open a ticket.

rinatkhaziev (changes requested):
It's ok :)

pschoffer:
Done I added the same as the prod check does.

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -253,6 +251,47 @@ async function processComponent( component: string, preselectedValue: string, de
...
+
+		const missingFiles = [];
+		for ( const file of files ) {
+			const filePath = path.resolve( providedPath, file );
+			if ( ! fs.existsSync( filePath ) ) {
+				missingFiles.push( file );
+			}
+		}
+		if ( missingFiles.length > 0 ) {
+			const message = `Provided path "${ providedPath }" is missing following files/folders: ${ missingFiles.join( ', ' ) }`;
```
rinatkhaziev:
```suggestion
			const message = `Provided path "${ providedPath }" is missing following files/folders: ${ missingFiles.join( ', ' ) }. Learn more: https://docs.wpvip.com/technical-references/vip-codebase/#1-wordpress`;
```

rinatkhaziev (approved):

PR #1071, opened 2022-08-08
[dev-env] update wizard prompts to be fully configurable using config switches

pschoffer:
## Description

During previous additions we lost the ability to run `vip dev-env create` without the use of wizard. This change revisits processing on user provided switches mostly around elasticsearch in order to be again able to avoid all the prompts. That way the tool can be used in CI/CD pipelines among other things.

## Steps to Test

```
npm run build && ./dist/bin/vip-dev-env-create.js --slug test --client-code ~/git/automattic/vip-go-skeleton --title Test --multisite false --php 8.0 --wordpress 6.0 --mu-plugins image --elasticsearch 7.17 -p false -x false
```


rinatkhaziev:
@pschoffer Did you test the upgrade path?

pschoffer:
Yeah I did, I thought it was fine:) But you convinced me to do it properly. So I added support for the old name too as we discussed.

sjinks (approved):

PR #1073, opened 2022-08-10
[dev-env] e2e Windows tests

pschoffer:
# Description
It would be great to get an automated test that validates dev-env working on windows machines. 

Investigating this I ran into an issue with running Linux based images on `windows-latest` runner. I was not able to figure out how to enable them. So I opened a [github discussion](https://github.com/community/community/discussions/29255)

In the meantime, though I think it would be worth adding the creation test which works fine. The way I coded it is very rough, so I would be happy to get any feedback on it. 

ping @bsessions85

### DIFF
--- a/__tests__/e2e_test.bat
+++ b/__tests__/e2e_test.bat
@@ -1,57 +1,22 @@
+call npm pack
 
-@REM call DockerCli -SwitchLinuxEngine
-call docker version
-call where docker
-@REM call dir "C:\Program Files\Docker"
-call where docker-compose
+echo "== Installing vip"
 
-@REM call dir \s DockerCli.exe
-C:
-call dir \s DockerCli.exe
-@REM call dir "C:\ProgramData\Chocolatey"
+FOR /R "." %%f IN ( *.tgz) DO  (
+    call npm i -g %%f
+)
 
-@REM call cat "C:\ProgramData\Docker\config\daemon.json"
+echo "== Running e2e tests"
 
-@REM call Install-WindowsFeature -Name Hyper-V -IncludeManagementTools -Restart
-call Install-Module -Name DockerMsftProvider -Repository PSGallery -Force
-call Get-PackageProvider -ListAvailableget-packagesource -ProviderName DockerMsftProvider
+call vip --help
 
-call wsl -l -o
+rem dev-env tests
 
-call wsl --status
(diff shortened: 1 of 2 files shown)


mjangda:
I looked into this before and one blocker was that GitHub didn't support running docker inside docker. Is that not an issue any more? Or do we have a workaround?

pschoffer:
Hmm I am still fighting with other dev-env issues. But guessing from the fact that windows runner comes with docker and docker-compose preinstalled it might work?

https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#tools

@mjangda

pschoffer:
It seems to be able to do docker in docker, but I am having hard time convincing it to run Linux containers as the daemon runs in win architecture. I need to read up and try more stuff in order to figure out how to massage it some more.

mjangda:
This is what I was thinking of:

https://github.com/actions/runner/issues/904

__tests__/e2e_test.bat
```diff
@@ -0,0 +1,22 @@
...
+rem dev-env tests
+
+echo "== Creating dev-env"
+
+call vip dev-env create --client-code image --title Test --multisite false --php 8.0 --wordpress 6.0 --mu-plugins image -e false -p false -x false
+
+if NOT %errorlevel% == 0 (
+    echo "== Failed to create dev-env"
+    exit 1
+)
```
mjangda, line 22: Do we want to check that the config files for the environment were actually created?

mjangda (commented):
One minor note but this is a very nice improvement 👍

__tests__/e2e_test.bat
pschoffer: Good idea. Done.

rinatkhaziev (approved):
Thanks! Admittedly the usefulness is limited because we can't run the containers.
@bsessions85 any ideas on how can we side-step this?

bsessions85:
> Thanks! Admittedly the usefulness is limited because we can't run the containers. @bsessions85 any ideas on how can we side-step this?

Thats a good question @rinatkhaziev. I have made it work with linux environment but haven't tried it on Windows before. I just did some quick searching and it doesn't appear to be possible as things stand today. I'll keep thinking through it, perhaps there is a workaround somehow that I'm not thinking of.

PR #1077, opened 2022-08-15
[dev-env] Remove app.env support from non-create command

pschoffer:
## Description

It turns out that havingt he support of `@app.env` is causing a bit of confusion and also a risk of running command meant for local dev-environment against real production site. This can lead to  negative outcomes (e.g. running `vip @123 import file.sql` instead of `vip @123 dev-env import file.sql`)

## Steps to Test

1) Test that only de-env subcommand that accepts @app.env is create, which fetches some data from prod instance
2) Other subcommand will issue a warning and NOT use `@app.env` in order to target env

Peek 2022-08-15 12-54

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-cli.js
+++ b/__tests__/lib/dev-environment/dev-environment-cli.js
@@ -68,8 +68,26 @@ describe( 'lib/dev-environment/dev-environment-cli', () => {
 				},
 				expected: 'foo',
 			},
+			{ // When app.env is not allowed use default value
+				options: {
+					allowAppEnv: false,
+					app: '123',
+					env: 'bar.car',
+					slug: 'foo',
+				},
+				expected: 'foo',
+			},
+			{ // When app.env is not allowed use default value
+				options: {
+					allowAppEnv: false,
+					app: '123',
+					env: 'bar.car',
+				},
+				expected: 'vip-local',
+			},
 			{ // construct name from app and env
 				options: {
+					allowAppEnv: true,
 					app: '123',
 					env: 'bar.car',
 				},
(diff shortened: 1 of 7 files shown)


yolih:
Beautiful, @pschoffer ! You even updated the examples in the help menu. Thank you so much!

rinatkhaziev:
I'm not sure falling back is a good idea. I'd be confused by this flow.

yolih:
Good catch, @rinatkhaziev! I was focused on the text and did not catch that function.

Aside from the functionality, can the error message be formatted as:
`This command does not support @app.env notation. Use '--slug=<example>' to target a specific local environment.`

pschoffer:
Good catch! I added `UserError` error class and made it so that if that one is thrown we don't get that "please contact customer support..." line.

@yolih , for the text did you mean actual `--slug=<example>` or the way how is it now? Right now we figure out what the slug should have been based on the app.env and use that in the error message. So:
```
$ ./dist/bin/vip-dev-env-start.js @2891
Error:  This command does not support @app.env notation. Use '--slug 2891' instead to target a specific local environment.
Debug:  VIP-CLI v2.15.0, Node v14.17.5, linux 5.15.0-43-generic
```

yolih:
@pschoffer Ah! Understood. If the returned error is able to dynamically and accurately populate the `--slug` value in the error message, then my suggestion is for the error message to read like this:
`This command does not support @app.env notation. Use '--slug=2891' to target the local environment.`
Where "2891" is populated by `${ appName }`.

pschoffer:
I updated the message.

sjinks (approved):

PR #1078, opened 2022-08-17
SQL Import: Fix multisite primary domain validation

t-wright:
## Description

The primary domain validation for multisite imports fails when the provided SQL file does not contain the `wp_site` table. 

Customers will often import just these tables. E.g. if they are adding content to a new site on the network in preparation for launch and they don’t want to touch any other sites on the network.

This PR corrects the behaviour and adds relevant tests.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-sql.js @<id>.<environment> path_to_sql_file.sql` referencing a file that only contains subsite tables
1. Verify that [this error is not thrown](https://github.com/Automattic/vip/blob/214a5eccc971bc4296d5d4b0e2400079def7aa6f/src/lib/validations/site-type.js#L74-L77)

### DIFF
--- a/src/lib/validations/is-multisite-domain-mapped.js
+++ b/src/lib/validations/is-multisite-domain-mapped.js
@@ -76,10 +76,6 @@ export async function isMultisitePrimaryDomainMapped(
 	envId: number,
 	primaryDomain: string,
 ): Promise<boolean> {
-	if ( ! primaryDomain ) {
-		return true;
-	}
-
 	const track = trackEventWithEnv.bind( null, appId, envId );
 	const api = await API();
 	let res;
--- a/src/lib/validations/is-multisite-domain-mapped.js
+++ b/src/lib/validations/is-multisite-domain-mapped.js
@@ -76,8 +76,11 @@ export async function isMultisitePrimaryDomainMapped(
 	envId: number,
 	primaryDomain: string,
 ): Promise<boolean> {
-	const track = trackEventWithEnv.bind( null, appId, envId );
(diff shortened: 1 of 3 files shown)


src/lib/validations/is-multisite-domain-mapped.js
```diff
@@ -76,8 +76,11 @@ export async function isMultisitePrimaryDomainMapped(
 	envId: number,
 	primaryDomain: string,
 ): Promise<boolean> {
-	const track = trackEventWithEnv.bind( null, appId, envId );
+	if ( ! primaryDomain ) {
+		return true;
```
saroshaga: So if there is no primary domain, we assume that the primary domain is mapped?

t-wright: If the import file doesn't update the `wp_site` table, then there's nothing to validate at this stage of the import process. In this scenario we just want the validation to pass.

saroshaga: Would it be clearer to have a separate `doesPrimaryDomainExist()` function and then use that here, or in the final check in `site-type.js`

t-wright: It would! Done in [5cb8c33](https://github.com/Automattic/vip/pull/1078/commits/5cb8c33e1efdda5e8e85077fc6cc448090c8ef65)

saroshaga (approved):

PR #1083, opened 2022-08-18
Software Management: Update

pschoffer:
## Description

Command to update software. 

```
vip @app.env config software update
(fires up prompts)
 
vip @app.env config software update node
(fires up prompt to select version for node)
 
vip @app.env config software update node 18
(no prompts and sets node to 18)
```

There is a confirm step that can be suppressed by `--force`

sw update

## Steps to Test

```
npm run build && ./dist/bin/vip-config-software-update.js @2891
```


src/bin/vip-config-software-update.js
```diff
@@ -0,0 +1,84 @@
...
+import { getUpdateResult, promptForUpdate, triggerUpdate } from '../lib/config/software';
+import { ProgressTracker } from '../lib/cli/progress';
+import UserError from '../lib/user-error';
+
+const debug = debugLib( '@automattic/vip:bin:config-software' );
+
+const UPDATE_SOFTWARE_PROGRESS_STEPS = [
+	{ id: 'trigger', name: 'Trigger software update' },
+	{ id: 'process', name: 'Process software update' },
+];
```
rinatkhaziev, line 28: Do we need `success` and `error` here?

src/bin/vip-config-software-update.js
```diff
@@ -0,0 +1,84 @@
...
+
+const cmd = command( {
+	appContext: true,
+	appQuery,
+	appQueryFragments,
+	envContext: true,
+	wildcardCommand: true,
+} );
+cmd.option( 'force', 'Auto-confirm update' );
```
rinatkhaziev, line 47: Does `--force` follow the conventions we have in other places?
I know we use `-y` in some places, in WP-CLI it's `--skip-confirm`
 cc @yolih

chriszarate, line 47: don't we use `--yes` elsewhere for this behavior?

chriszarate:
Seeing some odd behavior on the get command

https://user-images.githubusercontent.com/739304/185424457-742a65c2-67db-438a-9245-347f3a14d3ac.mov

chriszarate:
@pschoffer Need to add entry to `package.json#bin` for `vip-config-software-update`

chriszarate:
@pschoffer Node.js update failed:

```
> node ./dist/bin/vip-config-software-update.js @4092 nodejs
✔ Version for Node.js to upgrade to · 16
✔ Are you sure you want to upgrade Node.js to 16? (y/N) · true
✓ Trigger software update 
⠸ Process software update 
TypeError: Cannot read property 'inProgressLock' of null
    at _getCompletedJob (/Users/czarate/Documents/Code/vip-cli/dist/lib/config/software.js:299:18)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async getUpdateResult (/Users/czarate/Documents/Code/vip-cli/dist/lib/config/software.js:315:24)
    at async /Users/czarate/Documents/Code/vip-cli/dist/bin/vip-config-software-update.js:85:7
    at async Args._args.default.argv (/Users/czarate/Documents/Code/vip-cli/dist/lib/cli/command.js:507:11)
Error:  Unexpected error
Debug:  VIP-CLI v2.15.0, Node v14.19.0, darwin 21.6.0
```

(There is no locking for Node.js upgrades, since it depends on the TC build.)

src/bin/vip-config-software-update.js
mjangda: Oh my:

```
vip sync --help | grep force
    -f, --force    Skip confirmation (disabled by default)

vip import sql --help | grep force
    -f, --force           Skip confirmation (disabled by default)

vip wp --help | grep yes
    -y, --yes      Run the command in production without a confirmation prompt

vip config envvar set --help | grep skip
    -s, --skip-confirmation  Skip manual confirmation of input (USE WITH CAUTION) (disabled by default)
```

mjangda: See https://github.com/Automattic/vip/pull/1083#discussion_r949223526

rinatkhaziev: oh my indeed 😬

src/bin/vip-config-software-update.js
pschoffer: What do you mean? The proxyTracker has a way to mark each step as success / error (it changes to :heavy_check_mark: or :x: in the print out)

src/bin/vip-config-software-update.js
pschoffer: I see more `--force`, so we are starting the great unification!

pschoffer:
> Node.js update failed

I added an exception for `nodejs`, where we skip the progress job checking:
image

pschoffer:
> Seeing some odd behavior on the get command
> 
>  Screen.Recording.2022-08-18.at.10.45.11.AM.mov

@chriszarate It seems to work fine on my end. Do you have latest dependencies (`npm i`)? Could you please try envvar-list too? That uses the same formatter:
```
./dist/bin/vip-config-envvar-list.js @4092
```

chriszarate:
> Do you have latest dependencies (npm i)?

Yes indeedy.

> Could you please try envvar-list too? That uses the same formatter:

`./dist/bin/vip-config-envvar-list.js @4092` works as expected. 🤔

chriszarate:
> That uses the same formatter

@pschoffer The difference is that `envvar list` manually calls `formatData` and `console.log`. The "built-in" behavior used by `software get` pipes the output to a pager, which I don't think is desirable. I'm not sure where my formatting issue is coming from, but I'm guessing it's a difference between our chosen pagers.

A pager doesn't add much value here, and omits the output from the console history.

pschoffer:
Thanks for the get input @chriszarate . I implemented your suggestion in the other PR that deals with GET command tweaks - https://github.com/Automattic/vip/pull/1084 please check there.

src/bin/vip-config-software-update.js
```diff
@@ -0,0 +1,102 @@
...
+		description: 'Update Node.js to v16',
+	},
+] );
+cmd.option( 'force', 'Auto-confirm update' );
+cmd.argv( process.argv, async ( arg: string[], opt ) => {
+	const { app, env } = opt;
+	const { softwareSettings } = env;
+
+	if ( softwareSettings === null ) {
+		throw UserError( 'Software settings are not supported for this environmnet.' );
```
chriszarate:
```suggestion
		throw UserError( 'Software settings are not supported for this environment.' );
```

src/bin/vip-config-software-update.js
```diff
@@ -0,0 +1,102 @@
...
+
+	const hasProcessJob = updateData.component !== 'nodejs';
+	const steps = hasProcessJob ? UPDATE_SOFTWARE_PROGRESS_STEPS : [ UPDATE_SOFTWARE_PROGRESS_STEPS[ 0 ] ];
+	const progressTracker = new ProgressTracker( steps );
+
+	progressTracker.startPrinting();
+	progressTracker.stepRunning( 'trigger' );
+
+	const triggerResult = await triggerUpdate( { appId: app.id, envId: env.id, ...updateData } );
+	debug( 'Trigered update with result', triggerResult );
```
chriszarate:
```suggestion
	debug( 'Triggered update with result', triggerResult );
```

src/bin/vip-config-software-update.js
```diff
@@ -0,0 +1,102 @@
...
+
+		if ( ok ) {
+			console.log( chalk.green( '✓' ) + ' Software update complete' );
+		} else {
+			throw Error( errorMessage );
+		}
+	} else {
+		progressTracker.print();
+		progressTracker.stopPrinting();
+		console.log( chalk.green( '✓' ) + ` A new build of the application code has been initiated and will be deployed using Node.js v${ updateData.version } when the build is successful` );
```
chriszarate: It would be helpful to link them to the deployments page here. We should be able to since we have the app and env.

src/lib/config/software.js
```diff
@@ -47,3 +65,254 @@ export const appQueryFragments = `fragment Software on AppEnvironmentSoftwareSet
...
+	} ).run();
+
+	if ( confirm ) {
+		return {
+			component,
+			version,
+		};
+	}
+
+	throw new UserError( 'Update cancelled' );
```
chriszarate: we use American English variations (for better or worse 😂)

```suggestion
	throw new UserError( 'Update canceled' );
```

src/bin/vip-config-software-update.js
pschoffer: Good idea. I added the link.

pschoffer:
I fixed the suggestions @chriszarate. I have also managed to merge some other in-flight work and rebased this PR to synchronize the `UserError` logic.

src/lib/config/software.js
```diff
@@ -48,3 +67,254 @@ export const appQueryFragments = `fragment Software on AppEnvironmentSoftwareSet
...
+	const jobs = result?.data?.app?.environments[ 0 ].jobs || [];
+	for ( const job of jobs ) {
+		if ( latestJob ) {
+			if ( job.createdAt > latestJob.createdAt ) {
+				latestJob = job;
+			}
+		} else {
+			latestJob = job;
+		}
+	}
```
WPprodigy: This feels a bit simpler:

```
const jobs = [ { createdAt: 3 }, { createdAt: 1 }, { createdAt: 4 }, { createdAt: 2  } ];
const latestJob = jobs.reduce( ( prev, current ) => ( prev.createdAt > current.createdAt ) ? prev : current );
```

WPprodigy (commented):
So the major thing I noticed is that we'll need an option for WP to enable/disable "managed updates".

src/lib/config/software.js
pschoffer: I changed it. It needed the empty array check, but still probably simpler.

pschoffer:
I added the managed version support. Thanks @WPprodigy !

WPprodigy (approved):
Works good! Should we go ahead and add tracking at the same time, ala https://github.com/Automattic/vip/pull/1086 ?

pschoffer:
Naaah, I was hoping to merge this for some time:) I'll open a separate PR for tracking next.

PR #1085, opened 2022-08-22
[dev-env] rename clientCode to appCode 

pschoffer:
## Description

To make naming easier (and consistent) we renamed clientCode/siteCode to appCode.

## Steps to Test

` npm run build && ./dist/bin/vip-dev-env-create.js --slug test --app-code demo`


pschoffer:
@yolih should we add a usage example with `--app-code demo` as well? If yes please let me know the command and the description you would like me to add.

sjinks (approved):

PR #1090, opened 2022-09-01
Add tracking to software update

pschoffer:
## Description

We would like to track usage and errors related to software update CLI

## Steps to Test

```
 npm run build && ./dist/bin/vip-config-software-update.js @2891 php --debug @automattic/vip:analytics:clients:tracks
```

And look for the tracker lines.

### DIFF
--- a/src/bin/vip-config-software-update.js
+++ b/src/bin/vip-config-software-update.js
@@ -19,6 +19,7 @@ import { appQuery, appQueryFragments } from 'lib/config/software';
 import { getUpdateResult, promptForUpdate, triggerUpdate } from '../lib/config/software';
 import { ProgressTracker } from '../lib/cli/progress';
 import UserError from '../lib/user-error';
+import { trackEvent } from '../lib/tracker';
 
 const debug = debugLib( '@automattic/vip:bin:config-software' );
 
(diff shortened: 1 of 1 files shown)


src/bin/vip-config-software-update.js
```diff
@@ -49,57 +50,74 @@ cmd.argv( process.argv, async ( arg: string[], opt ) => {
...
 		}
-		progressTracker.print();
-		progressTracker.stopPrinting();
-
-		if ( ok ) {
-			console.log( chalk.green( '✓' ) + ' Software update complete' );
+		await trackEvent( 'config_software_update_success', { ...baseTrackingInfo, ...updateData } );
+	} catch ( error ) {
+		if ( error instanceof UserError ) {
+			await trackEvent( 'config_software_update_success', { ...baseTrackingInfo, ...updateData, user_error: error?.message } );
```
WPprodigy, line 117: This shouldn't be `success` should it?

pschoffer: I think it should. We had similar problem with `dev-env` where the one of the most common error was "you need to start your dev-env before doing this ..."-kind a thing. Which is not really a failure that we need to fix.

The UserError is analog to 4xx HTTP error. So user did something wrong not the code. We still have that `user_error` code there.

I guess alternatively we could use totally different key for the user_errors? `config_software_update_user_error` maybe? Nothing like that is used in other commands as far as I know, but it might make sense here?

WPprodigy: Okay, I see what you mean. If this is intended then all good with me!

WPprodigy (approved):

PR #1091, opened 2022-09-02
Extract http helper into standalone function

nickdaugherty:
## Description

This extracts the HTTP transport code from the Apollo handler to a standalone utility. This decouples the HTTP request handling (with auto authorization, user agent, etc) from things like automatic 401 error handling (which calls `process.exit()` when it happens).

This means we can call the API from other parts of the code without the side effects of the 401 handling or any other Apollo functionality.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-app-list.js`
1. Should list all apps like normal
2. Run `./dist/bin/vip-wp.js @297.testing help`
3. Should see `wp help` output like normal


brunobasto (approved):
This is awesome! Thank you for working on it @nickdaugherty 🙇

PR #1092, opened 2022-09-02
Improve login + logout

nickdaugherty:
## Description

Adds server-side logout from the `vip logout` command. This action deactivates the token and prevents further use.

Also prevents "unknown command" errors when running `vip login`, effectively adding `vip login` as a valid command (though due to how it's architected, it still doesn't show in `vip -h`). Previously, the only way to login was to run some non-login command while you were logged out, which is unexpected. `login` is a standard command across CLI tools.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip.js logout` - start fresh
1. Run `./dist/bin/vip.js login`
2. When generating the token on the Dashboard, save it for later use
3. Run regular commands to ensure things work - `./dist/bin/vip-app-list.js` for example
4. Via curl, or Insomnia, or Postman, make an HTTP request to the API with your token in `Authorization: Bearer $TOKEN` header ... should succeed
5. Logout again .... `./dist/bin/vip.js logout`
6. In the other program, repeat the request and you should now receive a 401 indicating your token is no longer valid


src/bin/vip.js
```diff
@@ -55,11 +56,12 @@ const rootCmd = async function() {
 
 	const isHelpCommand = process.argv.some( arg => arg === 'help' || arg === '-h' || arg === '--help' );
 	const isLogoutCommand = process.argv.some( arg => arg === 'logout' );
+	const isLoginCommand = process.argv.some( arg => arg === 'login' );
 	const isDevEnvCommandWithoutEnv = process.argv.some( arg => arg === 'dev-env' ) && ! containsAppEnvArgument( process.argv );
 
 	debug( 'Argv:', process.argv );
 
-	if ( isLogoutCommand || isHelpCommand || isDevEnvCommandWithoutEnv || ( token && token.valid() ) ) {
+	if ( ! isLoginCommand && ( isLogoutCommand || isHelpCommand || isDevEnvCommandWithoutEnv || ( token && token.valid() ) ) ) {
```
mjangda, line 64: At some point in the future, we should clean up this rootCmd and the argv function :) Starting to get too unweildy!

mjangda (approved):
Looks good.

Very minor issue is that you can run `vip login` (when already logged in) and `vip logout` (when already logged out). I think that's fine and not something we need to optimize for.

PR #1101, opened 2022-09-13
Remove Flow types

chriszarate:
## Description

As a follow-up to https://github.com/Automattic/vip/pull/866, and as a first step, we will remove Flow types and the Flow-specific build steps. Later, we can introduce a TypeScript rig and restore types. Type removal is constrained to a single commit (40ee1373df7cf0c7a49520d736ebd04a983a0667) so that they can be easily relocated.

Note that this also removes the `@format` markers for Prettier. They were non-functional; formatting rules are enforced by ESLint. They exist purely as copypasta.

## Steps to Test

- `npm test`
- `npm run build`


chriszarate:
New security issue flagged by CodeQL (newly able to parse files, I think) related to RegEx performance:
https://github.com/Automattic/vip/security/code-scanning/3

PR #1102, opened 2022-09-13
Remove http proxy support

nickdaugherty:
## Description

HTTP is insecure and risks leaking access tokens and other sensitive data. We should not support a known insecure proxy configuration.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `HTTP_PROXY=foobar ./dist/bin/vip-app-list.js`
1. Verify that the HTTP_PROXY environment variable does nothing ... you should see a list of apps


mjangda (approved):
LGTM

PR #1103, opened 2022-09-14
Force the preference for WebSocket in case we see an error during connection

luismulinari:
## Description

Clients with multiple outbound IPs are experiencing connection issues since the sticky sessions are being [configured by IP](https://socket.io/docs/v3/using-multiple-nodes/#nginx-configuration).

This PR fixes this issue by setting the transport preference for WebSocket in case we see an error during the connection phase.

In case we don't see any connection errors, we keep the default transport order, [following the official documentation](https://socket.io/docs/v3/how-it-works/#upgrade-mechanism).

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-wp.js @id.production --yes -- user list`
4. Command should execute successfully


nickdaugherty (approved):
LGTM! So on `connect_error`, it'll automatically retry again (after our code has changed the transports)?

luismulinari:
Hey Nick! Thanks for reviewing this.

> LGTM! So on `connect_error`, it'll automatically retry again (after our code has changed the transports)?

Exactly. It will retry with the new transport order.

PR #1104, opened 2022-09-15
[dev-env] fetch php and wordpress versions for creation wizard

pschoffer:
## Description

Fetch `php` and `wordpress` versions from real site and inform the default values during the setup wizard

## Steps to Test

```
npm run build && ./dist/bin/vip-dev-env-create.js @2891
```


jesse-greathouse (approved):
Tested and it worked great. I went into the dashboard and changed my site to use PHP 8.0, and as soon as that changed, I created my site again, and it had PHP 8.0 already selected. Very nice job @pschoffer

PR #1106, opened 2022-09-15
Using http module instead of already decommissioned API.apiFetch function

saroshaga:
## Description

The apiFetch method of the API module was removed earlier but all references were not updated: https://github.com/Automattic/vip/pull/1091

This PR uses the `http` internal module over here which is was intended with the earlier PR 

## Steps to Test

TBD


nickdaugherty (approved):
Whoops, thank you for catching this, my bad! 😄

PR #1107, opened 2022-09-15
[dev-env] Validate dns lookup

pschoffer:
## Description

The biggest pain point in using `vip dev-env` can be networking. Especially DNS lookup of `smth.vipdev.lndo.site` as this can be suppressed by DNS used by users.

This change introduces DNS lookup which should detect the problem in advance and provide a workaround in form of changing `hosts` file to circumvent domain name lookup altogether. 

## Steps to Test

1) Change the dev-env domain to some wrong address. e.g. 
```
const testDomain = `${ slug }.vipdev.lndo.site`;
```
change to 
```
const testDomain = `${ slug }.totally.not.correct`;
```
2) Run `npm run build && ./dist/bin/vip-dev-env-create.js --slug test`
3) You should get an error with suggestion to update hosts file.
4) Follow the suggestion
5) try again and prosper (no more errors)

### DIFF
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -63,8 +63,6 @@ addDevEnvConfigurationOptions( cmd );
 
 cmd.examples( examples );
 cmd.argv( process.argv, async ( arg, opt ) => {
-	await validateDependencies();
-
 	const environmentNameOptions = {
 		slug: opt.slug,
 		app: opt.app,
@@ -72,6 +70,9 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 		allowAppEnv: true,
 	};
 	const slug = getEnvironmentName( environmentNameOptions );
+
+	await validateDependencies( slug );
+
 	debug( 'Args: ', arg, 'Options: ', opt );
 
 	const trackingInfo = { slug };
--- a/src/bin/vip-dev-env-destroy.js
+++ b/src/bin/vip-dev-env-destroy.js
@@ -43,8 +43,8 @@ command()
 	.option( 'soft', 'Keep config files needed to start an environment intact' )
 	.examples( examples )
 	.argv( process.argv, async ( arg, opt ) => {
-		await validateDependencies();
 		const slug = getEnvironmentName( opt );
+		await validateDependencies( slug );
 
 		const trackingInfo = getEnvTrackingInfo( slug );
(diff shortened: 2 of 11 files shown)


jesse-greathouse:
I tested this extensively and it worked. I wasn't sure how it worked, when the testDomain is not changed, because the proxy was not running when the DNS lookup is happening. I even went so far as to delete all of my docker containers and images because I didn't understand why `test.vipdev.lndo.site` was resolving and `test.totally.not.correct` was not resolving.

This voodoo didn't sit right with me.

I expected that while the proxy is not running (and `test.vipdev.lndo.site` is not in my `/etc/hosts` file), that there should be no reason that this url would resolve to 127.0.0.1... yet it did.

I then killed docker and went straight to my computer's networking. The result was pretty surprising:
Screen Shot 2022-09-15 at 12 00 16 PM

Then I realized that there was some persistent routing in my computer's network interface somewhere.
I then ran the following commands to flush my computer's routing tables and restart the network interface.
Screen Shot 2022-09-15 at 12 12 17 PM

Once I flushed the routing tables, it stopped recognizing that route. I tried it again:
Screen Shot 2022-09-15 at 12 16 43 PM

So you may realize now, that the only reason this worked is because my computer's network interface had already set up a route for that domain. Therefore we can conclude that if a user tries to use the vip dev-env for the very first time, without prior routing of that domain, it will always fail to resolve the DNS for that user.

I think this PR has a great concept, but I also think that we need to somehow raise the proxy before doing doing the `dns.lookup()` method, or this will cause major problems for a lot of users.

jesse-greathouse (dismissed):
This is a good start but see my comment for some potential danger in this PR as it stands right now.

mjangda:
> I didn't understand why test.vipdev.lndo.site was resolving

This is because ([#](https://docs.lando.dev/config/proxy.html)):

> *.lndo.site is an actual ON THE INTERNET wildcard DNS entry that points all *.lndo.site subdomains to localhost/127.0.0.1

```
dig anything.vipdev.lndo.site +short
127.0.0.1
```

pschoffer:
The DNS lookup doesn't work with proxy at all. As Mo is saying this works because there is public DNS [record](https://mxtoolbox.com/SuperTool.aspx?action=a%3ajesse_is_awesome.vipdev.lndo.site&run=toolpage) and I think not because:
> So you may realize now, that the only reason this worked is because my computer's network interface had already set up a route for that domain.

jesse-greathouse (approved):

PR #1110, opened 2022-09-22
[dev-env] skip validate skips env up check too

pschoffer:
## Description

When users try to import SQL to dev-env we check for validity of SQL file. We also check that `dev-env` is UP. The user has an option to skip file validations with `--skip-validate`. This change makes it that `--skip-validate` will also skip environment status.

This is useful for cases when the environment is reported to be DOWN as false negative due to an issue with communication between terminal and public endpoints but still works fine in browser.

## Steps to Test
1) Have dev-env down
2) Run
```
npm run build && ./dist/bin/vip-dev-env-import-sql.js ~/Downloads/test.sql -S
```

### DIFF
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -68,18 +68,21 @@ command( {
 				} );
 			}
 
+			const execOptions = {
+				force: !! opt.skipValidate,
+			};
 			const importArg = [ 'wp', 'db', 'import', inContainerPath ];
-			await exec( slug, importArg );
+			await exec( slug, importArg, execOptions );
 
 			if ( searchReplace && searchReplace.length && ! inPlace ) {
 				fs.unlinkSync( resolvedPath );
 			}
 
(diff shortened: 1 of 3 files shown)


rinatkhaziev:
Ummm, I'm not sure about this, I guess it won't hurt, but should we look into maybe doing an alternative check when determining whether the env is UP or DOWN? Namely, we can hit the web URL and see if it's 200s or the request fails.

pschoffer:
> Ummm, I'm not sure about this, I guess it won't hurt, but should we look into maybe doing an alternative check when determining whether the env is UP or DOWN? Namely, we can hit the web URL and see if it's 200s or the request fails.

That wouldn't cover that scenario when the URL is not accessible from terminal. I've seen HTTP_PROXY set in terminal that can cause that, while it still works fine in browser (which is the same symptom described in the ZD tickets I saw). So basically doing `curl my-site.vip.lndo.site` fails but it works fine in browser.

 The import command itself is not using the domain or the network for the work it invokes cli directly in the container.

sjinks:
Added this functionality to #1201

PR #1112, opened 2022-09-23
Handle comand interruptions during software update

pschoffer:
## Description

When prompts are `ESC`ed or `Ctrl+C`ed during sw update we should handle it gracefully.

## Steps to Test

```
./dist/bin/vip-config-software-update.js @2891
✔ Component to update · wordpress
✖ Version for WordPress to upgrade to · 
Error:  Command cancelled by user.
Debug:  VIP-CLI v2.18.0, Node v14.18.0, linux 5.15.0-46-generic
```


jesse-greathouse (approved):
LGTM 👍

PR #1049, opened 2022-06-16
Fix security issues

sjinks:
## Description

This PR updates several vulnerable packages to fix critical and high-severity security issues.

* Inefficient Regular Expression Complexity in chalk/ansi-regex - GHSA-93q8-gq69-wqmw [High]
* Template injection in ejs - https://github.com/advisories/GHSA-phwq-j96m-2c2q [Critical]
* Prototype Pollution in minimist - https://github.com/advisories/GHSA-xvch-5gv4-984 [critical]
* Exposure of Sensitive Information to an Unauthorized Actor in node-fetch - https://github.com/advisories/GHSA-r683-j2x4-v87g [high]
* Exposure of Sensitive Information in simple-get - https://github.com/advisories/GHSA-wpg7-2c88-r8xv [high]

## Steps to Test

The CI should pass.

PR #1113, opened 2022-09-23
2.19 prep

pschoffer:
### 2.19.0 (23 Sep 2022)
- #1108 [dev-env] Validate import file and add debug lines
- #1107 [dev-env] Validate dns lookup
- #1104 [dev-env] fetch php and wordpress versions for creation wizard
- #1049 Fix security issues


andrea-sdl (approved):
🚢

PR #1114, opened 2022-09-26
[dev-env] Verify current user is a docker user

pschoffer:
## Description

If user is not in docker group. They can't interact with docker. Some of the symptoms are:

```
Error:  connect EACCES /var/run/docker.sock

-- or --

Could not automatically start the Docker Daemon. Please manually start it to continue.

```

This change adds a validation of the ability of fetching docker info. You would not be able to fetch that if you are not in docker group. Also this is done After we verified docker is installed.

## Steps to Test

```
./dist/bin/vip-dev-env-info.js 
Error:  Failed to connect to docker. Please follow the following instructions to add yourself to the docker group - https://docs.docker.com/engine/install/linux-postinstall/
Debug:  VIP-CLI v2.18.0, Node v10.19.0, linux 5.15.0-46-generic
```

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -353,3 +353,16 @@ export async function validateDockerInstalled() {
 		throw Error( 'docker-compose could not be located! Please follow the following instructions to install it - https://docs.docker.com/compose/install/' );
 	}
 }
+
+export async function validateDockerUser() {
+	const lando = new Lando( getLandoConfig() );
+	await lando.bootstrap();
+
+	const docker = lando.engine.docker;
+	lando.log.verbose( 'Fetching docker info to verify user is in docker group' );
+	try {
+		await docker.info();
+	} catch ( error ) {
+		throw Error( 'Failed to connect to docker. Please follow the following instructions to add yourself to the docker group - https://docs.docker.com/engine/install/linux-postinstall/' );
+	}
+}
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
@pschoffer, how does this work on Windows and Mac? Linking to Linux instructions will be confusing for Win users

pschoffer:
I think Docker Desktop takes care of it for you so this issue is only on linux when not using docker desktop, hence docker has only documentation for linux on how to do this.

https://docs.docker.com/engine/install/linux-postinstall/

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -352,3 +352,16 @@ export async function validateDockerInstalled() {
...
+export async function validateDockerUser() {
+	const lando = new Lando( getLandoConfig() );
+	await lando.bootstrap();
+
+	const docker = lando.engine.docker;
+	lando.log.verbose( 'Fetching docker info to verify user is in docker group' );
+	try {
+		await docker.info();
+	} catch ( error ) {
+		throw Error( 'Failed to connect to docker. Please follow the following instructions to add yourself to the docker group - https://docs.docker.com/engine/install/linux-postinstall/' );
```
sjinks: What do you think about this fix?

```suggestion
        if ( os.platform() === 'linux' ) {
    		throw Error( 'Failed to connect to docker. Please follow the following instructions to add yourself to the docker group - https://docs.docker.com/engine/install/linux-postinstall/' );
    	} else {
    	    throw Error( 'Please contact the support' ); // Obviously, we need a different text here
    	}
```

pschoffer:
I updated the PR with the validation outputs added recently + rephrase the error message to be a bit more generic as we are not sure that the user not being in docker group is actually the cause for them not being able to access docker commands.

rinatkhaziev (approved):

PR #1116, opened 2022-09-26
Reverted breaking change in PR #1049

andrea-sdl:
## Description

This PR reverts the code on PR #1049 while we investigate it further.

## Steps to Test

Example:

1. Check out PR.
1. Run `npm install`
1. Run `npm run build`
1. Run `node dist/bin/vip-app list` runs fine.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@automattic/vip",
-  "version": "2.19.0",
```
pschoffer: It shouldn't revert this I think.

pschoffer (changes requested):

package.json
andrea-sdl: you're right, updating the code

andrea-sdl: @pschoffer fixed

pschoffer (approved):
Thank you!

PR #1123, opened 2022-09-29
[dev-env] Add VSCode workspace config on start

pschoffer:
## Description

In order to make local setup easier for users of `dev-env` we autogenerate the workspace config usable for VSCode.

The workspace includes:
- folder with dev-env location
- folder where mu-plugins are sourced (only if not image)
- folder where clientCode is sourced (only if not image)
- xdebug config including correct path mappings

We also added a prompt after start offering to open vscode for them.

Peek 2022-10-05 07-52

## Steps to Test

```
$ npm run build && ./dist/bin/vip-dev-env-start.js
...
Already existing wp-config.php file
Checking for database connectivity...
Copying dev-env-plugin.php to mu-plugins
Checking for WordPress installation...
WordPress already installed
 SLUG              vip-local                                                                      
 LOCATION          /home/pavel/.local/share/vip/dev-environment/vip-local                         
 SERVICES          devtools, nginx, php, database, memcached, wordpress                           
 NGINX URLS        http://vip-local.vipdev.lndo.site/                                             
                   https://vip-local.vipdev.lndo.site/                                            
 DATABASE          127.0.0.1:49212                                                                
 STATUS            UP                                                                             
 LOGIN URL         http://vip-local.vipdev.lndo.site/wp-admin/                                    
 DEFAULT USERNAME  vipgo                                                                          
 DEFAULT PASSWORD  password                                                                       
 DOCUMENTATION     https://docs.wpvip.com/technical-references/vip-local-development-environment/ 
VSCode workspace generated to use it run:
code /home/pavel/.local/share/vip/dev-environment/vip-local/vip-local.code-workspace
✔ Would you like to run the command now? (Y/n) · true
```


rinatkhaziev:
Discussed in Slack, but I think we should take it further a notch and generate and save the `launch.json` in the env instance folder.

pschoffer:
We switched to generate the whole VSCode workspace instead of only `launch.json`. This now becomes valuable even if the user is not using xdebug to open all the relevant code in one space.

rinatkhaziev (commented):
Thanks for this, code looks fine, but should we check whether a code space already exists? 

The scenario that I'm a bit worried about - what if I modified the workspace file and then the next start command just overrides it.

pschoffer:
Fixed

pschoffer:
I am not sure about the prompt ... It is probably annoying, especially if you don't use VSCode. Should we switch to use of a switch? Maybe `vip dev-env start --vscode` will start the vscode. Otherwise we just create the workspace file but not prompt. What do you think @rinatkhaziev ?

rebeccahum:
Is it possible to detect if a VS codespace file exists before automatically creating it?

rinatkhaziev:
The prompt could be like 
`We've detected the workspace file is already present, do you want to overwrite? `
`Yes/No/Don't ask again`

pschoffer:
I meant the prompt to open vscode not if we should overwrite (I am fine with not overwriting it).

I updated the code to what I meant. Please give it a try and let me know.

```
npm run build && ./dist/bin/vip-dev-env-start.js --vscode
```

pschoffer:
This is now rebased and ready for review again!

pschoffer:
About your note from slack @rinatkhaziev 
>  I think we should abstract out the part that generates the mapping, so we can re-use it later for other editors as well.

I aggree, but it is already pretty well-encapsulated [function](https://github.com/Automattic/vip-cli/blob/add/xdebug_config/src/lib/dev-environment/dev-environment-core.js#L737). It is VSCode specific yes, but it would be pretty easy lift to take it out when and **if** we build in support for other tool.

Or did you mean to actually expose a way to generate say `json` file only with the mappings?

rinatkhaziev:
@pschoffer 

> Or did you mean to actually expose a way to generate say `json` file only with the mappings?

I meant something like (pseudo): 

```javascript

const generatePathMappings = instanceData => ({
   '/wp/wp-content/client-mu-plugins' : path.resolve( instanceData.appCode.dir, 'client-mu-plugins' ),
// omitted
})
```

Then, when we decide to add a PHPStorm (or whatever IDE) support we will already have that piece

pschoffer:
OK Done @rinatkhaziev

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -711,3 +713,22 @@ export function getEnvTrackingInfo( slug: string ): any {
...
+
+	if ( fs.existsSync( workspacePath ) ) {
+		console.log( `VSCode workspace already exists, skipping creation. To use it run:\ncode ${ workspacePath }` );
+	} else {
+		generateVSCodeWorkspace( slug );
+		console.log( `VSCode workspace generated to use it run:\ncode ${ workspacePath }` );
+	}
+
+	if ( options.openVSCode ) {
+		spawn( 'code', [ workspacePath ] );
```
rinatkhaziev: I think this needs a safe-guard - `code` is not guaranteed to be in path, in case it's missing this will happen: 

```
Error: spawn code ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Error:  Unexpected error
```

rinatkhaziev (changes requested):
Left an inline comment. 

Functionally there's one thing we should address. It seems that we're generating the VSCode project no matter what, I don't think that's desirable - what if I as a user don't use VSCode? How is this relevant to me?

src/lib/dev-environment/dev-environment-cli.js
pschoffer: I added call to `code --version` which should work on all OS. 

And skip that if they dont have vscode.

Also added a note: "VSCode not detected, skipping opening VSCode" if they would provide `--vscode` switch but dont have vscode

sjinks: Those who use VSCode Insiders, have a different binary (`code-insiders` vs `code`). And there is [`codium`](https://github.com/VSCodium/vscodium) if we care.

[Ref](https://github.com/devcontainers/features/blob/main/src/common-utils/bin/code)

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -732,3 +741,12 @@ export async function postStart( slug: strin, options: PostStartOptions ) {
 		spawn( 'code', [ workspacePath ] );
 	}
 }
+
+const isVSCodeInstalled = () => {
+	try {
+		const result = spawnSync( 'code', [ '--version' ] );
```
sjinks: There is a [more lightweight and reliable way](https://github.com/shelljs/shelljs#whichcommand).

`shelljs` is a dependency of `lando`; we can use it "for free".

rinatkhaziev (dismissed):

src/lib/dev-environment/dev-environment-cli.js
rinatkhaziev: @pschoffer How about:

1. when the flag `--vscode` is passed, we generate the workspace no matter what.
2. If we can't determine the proper binary we just print `VSCode workspace has been generated at ${workspacePath`
3. If we can determine the binary we spawn it.

rinatkhaziev: @sjinks is `codium` binary at `codium`? I think I agree with you we should support `code-insiders` as well.

src/lib/dev-environment/dev-environment-cli.js
pschoffer: nice

src/lib/dev-environment/dev-environment-cli.js
pschoffer: Yep its codium I checked.

pschoffer:
Here is how I changed it @rinatkhaziev  / @sjinks . What do you think?

* No `--vscode` switch, nothing is generated or outputed related to worksapce
* `--vscode` switch present
   * I always generate workspace
   * If any of the `[ 'code', 'code-insiders', 'codium' ]` is found I use that to open workspace
   * Otherwise I output: `VSCode not detected in path, please open ${ workspacePath } with VSCode`

rinatkhaziev:
I tested it, works under Mac OS (with VSCode installed) and under WSL (no `code` installed).

However, under native Windows it successfully detects the VSCode but spawning fails with ENOENT, even though typing `code` in cmd works.

sjinks:
@rinatkhaziev could you please change this line: https://github.com/Automattic/vip-cli/pull/1123/files#diff-ab286c4e349fcc4e35d4c6afc0e2a1f948c379423372eb116384b760f200d09dR752

from `return candidate;` to `return result.toString();` and try again?

rinatkhaziev:
@sjinks that was different issue - https://stackoverflow.com/questions/37459717/error-spawn-enoent-on-windows

rinatkhaziev:
Soo it's a bit useless under WSL, but ChatGPT tells me it's possible: 

Screenshot 2023-04-20 at 1 58 45 PM

rinatkhaziev (changes requested):
We're almost there but should try to fix WSL mappings before this is merged.

pschoffer:
Nice fix for the WSL! 

What is wrong with the mappings @rinatkhaziev ? 

I was not using WSL on my Windows machine (powershell all the way). It seemed to work fine. I was able to run debug too. Paths were correct.

I am trying to set up WSL to talk to docker now, to reproduce your environment. I'll continue on that today.

rinatkhaziev:
Yeah I know that non-WSL version works fine. I guess I'm going to retry with having code at `/mnt/c/` rather than inside WSL. I'll follow-up here.

pschoffer:
I got it working on WSL. I didn't encounter any big issues. The one thing I had was that the mappings were not interpretted correctly. Apparently order matters, and it tried to look up the clint-code files inside the `wordpress` folder.

The last commit that adds the `wp` mapping as a last one solved it for me.

Was that the same issue you had @rinatkhaziev ? 

image

rinatkhaziev:
@pschoffer no, my Xdebug is not working. I.e. I put a breakpoint in code, and I pointed the client-code to `/mnt/c/my/code/cantina-test-sites` and it's not stopping on breakpoints. I'm still looking. Does debug work for you?

pschoffer:
Yeah it does. I tried client-mu-plugin and WordPress code breakpoints.

 You can also check what's up in the xdebug server log.

To get there, open the terminal in the `php` container and tail `/tmp/xdebug.log`. It tells you if it connects to client. If not which endpoint it is trying. When setting breakpoint what path it will use inside php.

You also mentioned `discover_client_host` in the p2. Have that solved it for you?

rinatkhaziev:
Yeah it does fix it for me and doesn't break Mac, @sjinks tested on Linux and it seems to work for him there too.

pschoffer:
Cool. Here is the PR to add that to the default setup - https://github.com/Automattic/vip-container-images/pull/471

Should we wrap up this PR then @rinatkhaziev ?

rinatkhaziev (approved):

anotherlatenightcoder:
Hi, I got referred here by our VIP support manager. Could you please tell me whether there are any plans for an integration into PHPStorm?

PR #1124, opened 2022-09-30
[dev-env] Supress Setup wizard during update

pschoffer:
## Description

This change makes is so that when any option that changes dev-env config is provided to `vip dev-env update` the setup wizard is not triggered. The values that are not provided are kept the same as they were.

If no option is provided the wizard still starts like it did.

## Steps to Test

```
 npm run build && ./dist/bin/vip-dev-env-update.js --phpmyadmin false
```

### DIFF
--- a/src/bin/vip-dev-env-update.js
+++ b/src/bin/vip-dev-env-update.js
@@ -76,7 +76,12 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			title: '',
 		};
 
-		const instanceData = await promptForArguments( preselectedOptions, defaultOptions );
+		const providedOptions = Object.keys( opt )
+			.filter( option => option.length > 1 ) // Filter out single letter aliases
+			.filter( option => ! [ 'debug', 'help', 'slug' ].includes( option ) ); // Filter out options that are not related to instance configuration
+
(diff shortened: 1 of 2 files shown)


jesse-greathouse (approved):
👍

src/bin/vip-dev-env-update.js
```diff
@@ -76,7 +76,12 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 			title: '',
 		};
 
-		const instanceData = await promptForArguments( preselectedOptions, defaultOptions );
+		const providedOptions = Object.keys( opt )
+			.filter( option => option.length > 1 ) // Filter out single letter aliases
```
rinatkhaziev, line 80: Why are we filtering out the single-letter aliases?

rinatkhaziev (commented):
Left a note inline.

src/bin/vip-dev-env-update.js
pschoffer: The library we use to parse arguments always populates both no matter if you actually used single letter or full world one.

It doesn't matter too much but I would had to add also the one letter version variants to the arguments we should ignore without it.

PR #1125, opened 2022-10-03
add validation output

pschoffer:
## Description

We already have some validations in place before each command. This change makes the validation visible on every run. This should help troubleshoot problems by users and support.

## Steps to Test

```
$ ./dist/bin/vip-dev-env-info.js 
Running validation steps...
✓ Check for docker instalation 
✓ Check for docker-compose instalation 
✓ Check DNS resolution 
 SLUG              vip-local                                                                      
 LOCATION          /home/pavel/.local/share/vip/dev-environment/vip-local                         
...
```

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -17,7 +17,7 @@ import os from 'os';
 /**
  * Internal dependencies
  */
-import * as exit from 'lib/cli/exit';
+import { ProgressTracker } from 'lib/cli/progress';
 import { trackEvent } from '../tracker';
 import {
 	DEV_ENVIRONMENT_FULL_COMMAND,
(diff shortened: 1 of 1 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -95,14 +95,31 @@ const verifyDNSResolution = ( slug: string ) => {
 		} );
 	} );
 };
-
+const VALIDATION_STEPS = [
+	{ id: 'docker', name: 'Check for docker instalation' },
```
sjinks:
```suggestion
	{ id: 'docker', name: 'Check for docker installation },
```

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -95,14 +95,31 @@ const verifyDNSResolution = ( slug: string ) => {
...
+	{ id: 'compose', name: 'Check for docker-compose instalation' },
```
sjinks:
```suggestion
	{ id: 'compose', name: 'Check for docker-compose installation },
```

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -95,14 +95,31 @@ const verifyDNSResolution = ( slug: string ) => {
...
 	} catch ( exception ) {
 		throw new UserError( exception.message );
 	}
+	progressTracker.stepSuccess( 'docker' );
+	progressTracker.stepSuccess( 'compose' );
+	progressTracker.print();
+
 	await verifyDNSResolution( slug );
+
+	progressTracker.stepSuccess( 'dns' );
```
sjinks, line 119: I didn't look into `progressTracker` but I would like to confirm that we don't need

```js
progressTracker.stepRunning( 'dns' );
```

before `await verifyDNSResolution( slug );`

sjinks (changes requested):

src/lib/dev-environment/dev-environment-cli.js
pschoffer: Thanks for the fixes!

Yeah `progressTracker.stepSuccess( <stepId> );` moves it to the next step so no need to call stepRunning.

sjinks (approved):

PR #1127, opened 2022-10-03
Run wp-cli as www-data

sjinks:
## Description

This PR updates the Lando template to run WP CLI as `www-data` user for the reasons specified in Automattic/vip-container-images#293

## Steps to Test

Check out this PR, create a dev env with ElasticSearch turned on, enable ES, run

```sh
vip dev-env exec -- wp vip-search index --setup --skip-confirm
```

Ensure that there are no permission errors out there.

Without this PR, the output will look like this:
```
Warning: error_log(/tmp/logstash.log): Failed to open stream: Permission denied in /wp/wp-content/mu-plugins/logstash/class-logger.php on line 541 [$ /usr/local/bin/wp --allow-root vip-search index --setup --skip-confirm] [wp-content/mu-plugins/logstash/class-logger.php:541 error_log(), wp-includes/class-wp-hook.php:307 Automattic\VIP\Logstash\Logger::process_entries_on_shutdown(), wp-includes/class-wp-hook.php:331 WP_Hook-&gt;apply_filters(), wp-includes/plugin.php:476 WP_Hook-&gt;do_action(), wp-includes/load.php:1102 do_action('shutdown'), shutdown_action_hook()]

Warning: error_log(/tmp/logstash.log): Failed to open stream: Permission denied in /wp/wp-content/mu-plugins/logstash/class-logger.php on line 541 [$ /usr/local/bin/wp --allow-root vip-search index --setup --skip-confirm] [wp-content/mu-plugins/logstash/class-logger.php:541 error_log(), wp-includes/class-wp-hook.php:307 Automattic\VIP\Logstash\Logger::process_entries_on_shutdown(), wp-includes/class-wp-hook.php:331 WP_Hook-&gt;apply_filters(), wp-includes/plugin.php:476 WP_Hook-&gt;do_action(), wp-includes/load.php:1102 do_action('shutdown'), shutdown_action_hook()]

[...]

PR #1129, opened 2022-10-04
Fix ESLint/Flow/SonarScan issues

sjinks:
## Description

This PR fixes the issues found by [ESLint](https://github.com/Automattic/vip-cli/pull/1127/files) and, for the affected files, Flow and SonarScan.

## Steps to Test

There are no breaking changes; the CI should pass.

### DIFF
--- a/__tests__/lib/cli/config.js
+++ b/__tests__/lib/cli/config.js
@@ -47,7 +47,7 @@ describe( 'utils/cli/config', () => {
 			const config = require( 'lib/cli/config' );
 			expect( config.default ).toMatchObject( expected );
 		} catch ( error ) {
-			expect( hasError ).toBeTruthy();
+			expect( hasError ).toBe( true );
 		}
 	} );
 } );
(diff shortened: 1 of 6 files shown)


rinatkhaziev:
@sjinks are we sure that removing `await`  is really non-breaking? As in, did we test it?

sjinks:
An async function always returns a Promise. If you are trying to return something else, that something else is internally converted into a Promise (smth like Promise.resolve(retval).

Thus, when you do return await, you wait for the promise to fullfil to immediately wrap it into another promise.

The only difference is when a promise is rejected. If you do await in the function, the function will throw an exception; if you return a Promise instead, the exception will happen in the caller.

BTW, eslint can fix this automatically, this is a safe conversion.

sjinks:
We did the same in mu-plugins' e2e tests, everything works 🙂

rinatkhaziev (approved):

PR #1131, opened 2022-10-06
FORNO-1296: SQL Import - Add ALTER TABLE statement validation

t-wright:
## Description

DB exports from tools like phpMyAdmin sometimes define constraints like primary keys in `ALTER TABLE` statements at the end of the file. When customers import these files, it can lead to a scenario where another process inserts data into a table before its constraints are set. Since we require the SQL file being imported to contain `DROP TABLE` and `CREATE TABLE` statements, this PR encourages defining the table structure in the `CREATE TABLE` statements by disallowing `ALTER TABLE`.

## Documentation

https://veetoop2.wordpress.com/2022/10/06/disallowing-alter-table-statements-in-sql-imports/

## Steps to Test

1. Produce an SQL file containing an `ALTER TABLE` statement:

```

  `ID` bigint(20) UNSIGNED NOT NULL,
  `timestamp` bigint(20) UNSIGNED NOT NULL,
  `args` longtext NOT NULL,
  `interval` int(10) UNSIGNED DEFAULT 0,
  `created` datetime NOT NULL,
  `last_modified` datetime NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

ALTER TABLE `wp_a8c_cron_control_jobs`
  ADD PRIMARY KEY (`ID`),
  ADD UNIQUE KEY `ts_action_instance_status` (`timestamp`,`action`(191),`instance`,`status`),
  ADD KEY `status` (`status`);
```

3. Check out PR.
4. Run `npm run build`
5. Run `./dist/bin/vip-import-validate-sql.js path/to/file.sql`
6. Verify that validation fails

### DIFF
--- a/__fixtures__/validations/bad-sql-dump.sql
+++ b/__fixtures__/validations/bad-sql-dump.sql
@@ -32,3 +32,8 @@ INSERT INTO wp_options (option_name, option_value, autoload)
         ('home', 'https://super-empoyees.com', 'yes');
 
 SET @@SESSION.SQL_LOG_BIN= 0;
+
+ALTER TABLE wp_options
+	ADD PRIMARY KEY (`option_id`),
+	ADD UNIQUE KEY `option_name` (`option_name`),
+	ADD KEY `autoload` (`autoload`);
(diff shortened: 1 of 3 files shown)


saroshaga (approved):
Tested this out and it looks good! Thanks for working on this!!

I have tested this only on the `validate-sql` command. The same validations will also run on the actual import command as well, right?

t-wright:
> I have tested this only on the `validate-sql` command. The same validations will also run on the actual import command as well, right?

It will 👍

PR #1134, opened 2022-10-10
[dev-env] add app and env to tracks when creating env

pschoffer:
## Description

When creating dev-env we can pass app and env from user to know if they used `@target` or not.

## Steps to Test

```
npm run build && ./dist/bin/vip-dev-env-create.js @2891  --debug @automattic/vip:analytics:clients:tracks
```

### DIFF
--- a/src/bin/vip-dev-env-create.js
+++ b/src/bin/vip-dev-env-create.js
@@ -75,7 +75,11 @@ cmd.argv( process.argv, async ( arg, opt ) => {
 
 	debug( 'Args: ', arg, 'Options: ', opt );
 
-	const trackingInfo = { slug };
+	const trackingInfo = {
+		slug,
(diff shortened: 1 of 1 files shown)


brunobasto (approved):
Looks good! Thank you @pschoffer

simonwheatley (changes requested):
Could we have a `has_target` property which is `true` if `@thing.thing` is present? The immediate question I have is "how many times is `vip dev-env create` run with a target?" and there may be other similar questions in the future. Does this make sense? Could I solve or this with these additional or with existing properties, do you think?

pschoffer:
Yes, app will only be provided if it was used via `@<app_id>` or `--app <app-id>` select. So hasTarget is the same as `app != false` @simonwheatley

PR #1141, opened 2022-10-18
SQL validation: Change formatting to make it more readable

rebeccahum:
## Description

I personally found it confusing that the errors were listed after the error message displaying that the validation failed.

All this change does is:
- moves the ordering of the list of errors to be in line with the grouping of the list of warnings (above the message displaying validation error)
- bolds and reds the entire validation error message

Before:
Screen Shot 2022-10-18 at 12 03 07 PM

After:
Screen Shot 2022-10-18 at 12 00 43 PM

## Steps to Test

1) Use a SQL file with both warnings and errors
2) `npm run build`
3) `./dist/bin/vip-dev-env-import-sql.js file.sql`

### DIFF
--- a/src/lib/validations/sql.js
+++ b/src/lib/validations/sql.js
@@ -403,10 +403,7 @@ const postValidation = async ( options: ValidationOptions ) => {
 			error: errorSummary,
 		} );
 
-		const errorOutput = [
-			`SQL validation failed due to ${ chalk.red( problemsFound ) } error(s)`,
-			'',
-		];
+		const errorOutput = [];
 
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
Should this be the other way around, though?

SQL validation failed due to 1 error(s)
Error explanation and recommendation

rebeccahum:
I don't think so because what if the error list is really long? Then you miss that status message saying how many errors it failed from. When I run a command, I expect to see the bottom message with a failure/success message and the log above.

rinatkhaziev (approved):

PR #1142, opened 2022-10-19
Update Slack channel on which to (optionally) ping before release

saroshaga:
## Description 

If it is actually required, we should ping on the VIP Platform Slack channel when releasing


mjangda (approved):
👍

PR #1149, opened 2022-10-23
refactor(dev-env): Fix Flow types

sjinks:
## Description

This PR fixes most of the Flow errors in the dev env component.

## Steps to Test

The CI should pass.

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -28,13 +28,26 @@ import {
 	DEV_ENVIRONMENT_PHP_VERSIONS,
 } from '../constants/dev-environment';
 import { getVersionList, readEnvironmentData } from './dev-environment-core';
-import type { AppInfo, ComponentConfig, InstanceOptions, EnvironmentNameOptions, InstanceData } from './types';
+import type {
+	AppInfo,
+	ComponentConfig,
+	InstanceOptions,
+	EnvironmentNameOptions,
+	InstanceData,
+	WordPressConfig,
+} from './types';
 import { validateDockerInstalled, validateDockerAccess } from './dev-environment-lando';
 import UserError from '../user-error';
+import typeof Command from 'lib/cli/command';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
 const DEFAULT_SLUG = 'vip-local';
+
+// Forward declaratrion to avoid no-use-before-define
+declare function promptForComponent( component: 'wordpress', allowLocal: false, defaultObject: ComponentConfig | null ): Promise<WordPressConfig>;
+// eslint-disable-next-line no-redeclare
+declare function promptForComponent( component: string, allowLocal: boolean, defaultObject: WordPressConfig | null ): Promise<ComponentConfig>;
 
 export async function handleCLIException( exception: Error, trackKey?: string, trackBaseInfo?: any = {} ) {
 	const errorPrefix = chalk.red( 'Error:' );
@@ -196,14 +209,14 @@ export function getOptionsFromAppInfo( appInfo: AppInfo ): InstanceOptions {
  * Prompt for arguments
  * @param {InstanceOptions} preselectedOptions - options to be used without prompt
  * @param {InstanceOptions} defaultOptions - options to be used as default values for prompt
- * @param {boolean} supressPrompts - supress prompts and use default values where needed
+ * @param {boolean} suppressPrompts - supress prompts and use default values where needed
  * @returns {any} instance data
  */
-export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: $Shape<InstanceOptions>, supressPrompts: boolean = false ): Promise<InstanceData> {
+export async function promptForArguments( preselectedOptions: InstanceOptions, defaultOptions: InstanceOptions, suppressPrompts: boolean = false ): Promise<InstanceData> {
 	debug( 'Provided preselected', preselectedOptions, 'and default', defaultOptions );
 
-	if ( supressPrompts ) {
-		preselectedOptions = { ...defaultOptions, ...preselectedOptions };
+	if ( suppressPrompts ) {
+		preselectedOptions = { ...( defaultOptions: Object ), ...( preselectedOptions: Object ) };
 	} else {
 		console.log( DEV_ENVIRONMENT_PROMPT_INTRO );
 	}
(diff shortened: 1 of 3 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -59,14 +64,22 @@ type SQLImportPaths = {
...
+
+export async function startEnvironment( slug: string, options: StartEnvironmentOptions ): Promise<void> {
 	debug( 'Will start an environment', slug );
 
 	const instancePath = getEnvironmentPath( slug );
 
 	debug( 'Instance path for', slug, 'is:', instancePath );
 
-	const environmentExists = fs.existsSync( instancePath );
+	const environmentExists = await fs.existsSync( instancePath );
```
pschoffer: What is this change for? I would assume sync variant of he comman doesn't need await?

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -310,12 +323,14 @@ async function prepareLandoEnv( instanceData, instancePath ) {
 	const nginxFileTargetPath = path.join( nginxFolderPath, nginxFileName );
 	const instanceDataTargetPath = path.join( instancePath, instanceDataFileName );
 
-	fs.mkdirSync( instancePath, { recursive: true } );
-	fs.mkdirSync( nginxFolderPath, { recursive: true } );
+	await fs.promises.mkdir( instancePath, { recursive: true } );
```
pschoffer, line 326: Why do you prefer the async api here?

I think `fs.promises.mkdir` is supported on less node versions than `fs.mkdirSync`.

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -626,23 +641,18 @@ function isVersionListExpired( cacheFile, ttl ) {
...
-			res = await fetchVersionList();
-			fs.writeFileSync( cacheFile, res );
-		}
-
 		// If the cache is expired, refresh it
-		if ( isVersionListExpired( cacheFile, DEV_ENVIRONMENT_WORDPRESS_VERSION_TTL ) ) {
+		if ( ! fs.existsSync( cacheFile ) || isVersionListExpired( cacheFile, DEV_ENVIRONMENT_WORDPRESS_VERSION_TTL ) ) {
 			res = await fetchVersionList();
-			fs.writeFileSync( cacheFile, res );
+			await fs.promises.writeFile( cacheFile, res );
```
pschoffer, line 655: Same question as above sync vs promises

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -652,7 +662,8 @@ export async function getVersionList() {
 
 	// Try to parse the cached file if it exists.
 	try {
-		return JSON.parse( fs.readFileSync( cacheFile ) );
+		const data = await fs.promises.readFile( cacheFile, 'utf8' );
```
pschoffer, line 665: and here

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -28,14 +28,27 @@ import {
...
 import { validateDockerInstalled, validateDockerAccess } from './dev-environment-lando';
 import UserError from '../user-error';
+import typeof Command from 'lib/cli/command';
 
 const debug = debugLib( '@automattic/vip:bin:dev-environment' );
 
 const DEFAULT_SLUG = 'vip-local';
 
+// Forward declaratrion to avoid no-use-before-define
+declare function promptForComponent( component: 'wordpress', allowLocal: false, defaultObject: ComponentConfig | null ): Promise<WordPressConfig>;
```
pschoffer, line 48: Maybe let's have a different function for WordPress vs other components (where you get to choose the mode) altogether.

Apart from cleaner type definitions, the functions would get simpler too.

What do you think?

pschoffer (changes requested):
This is great. I left a few notes.

src/lib/dev-environment/dev-environment-core.js
sjinks: Yeah, I was planning to use `fs.promises`, but `exists`/`existsSync` are deprecated and are not a part of the Promises API.

src/lib/dev-environment/dev-environment-core.js
sjinks: https://github.com/Automattic/vip-cli/blob/develop/package.json#L77

The Promises API is available since [v10.0.0](https://nodejs.org/api/fs.html#promises-api).

As for the async API - it probably does not matter much for a CLI application, but in general, it is [not recommended to block the event loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/).

With the async API, we can use [`Promise.all`](https://github.com/Automattic/vip-cli/pull/1149/files/f4edb944f3d17d9e04a829177a6106a9ac0e8f87#diff-c7d8beaf4cbe55eeb49f1a4ae46c4b0efd1ea6f21d1863eeea06554e89bbc57dR329-R333) to make things a bit faster.

src/lib/dev-environment/dev-environment-core.js
sjinks: My personal preference is not to mix async and sync calls. The function already uses `await`, so it makes sense to use the async API.

src/lib/dev-environment/dev-environment-core.js
sjinks: Same.

src/lib/dev-environment/dev-environment-cli.js
sjinks: Yes, I agree. But this requires more changes to the code, so I left that for another PR to make it easier to review this one.

src/lib/dev-environment/dev-environment-core.js
sjinks: Also, Linux natively supports the [AIO API](https://man7.org/linux/man-pages/man7/aio.7.html), which means that the async versions could be less resource-consuming than the synchronous APIs (TBH, I did not check whether `libuv` uses the AIO or a thread pool on Linux).

pschoffer: Oh right:
> VIP-CLI requires a minimum installed version of Node.js v14 and npm v6
From the [docs](https://docs.wpvip.com/technical-references/vip-cli/installing-vip-cli/#h-update-node-js-and-npm)

So we should be all good here.

pschoffer (approved):

PR #1151, opened 2022-10-25
refactor(dev-env): Modify `verifyDNSResolution()` to use Promises API

sjinks:
## Description

Tiny refactoring: this PR updates the `verifyDNSResolution()` function to use the [DNS Promises API](https://nodejs.org/docs/latest-v14.x/api/dns.html#dns_dns_promises_api) instead of the old callback-style API.

## Steps to Test

Apply the patch and make sure that `vip dev-env` commands still work.


pschoffer (approved):
Cool!

PR #1156, opened 2022-10-25
Validate env before resolving path

pschoffer:
## Description

We would not validate environment if doing search-replace during the SQL import soon enough. This would result in confusing error.

This change adds extra validation to catch this scenario.

## Steps to Test

```
npm run build && ./dist/bin/vip-dev-env-import-sql.js --slug=wrong ~/Downloads/test --skip-validate --search-replace=a,b

Error: Environment doesn't exist.


To create a new environment run:

vip dev-env create

```

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -446,6 +446,16 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
 	debug( `Will try to resolve path - ${ fileName }` );
 	let resolvedPath = resolvePath( fileName );
 
+	const instancePath = getEnvironmentPath( slug );
+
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -446,6 +446,16 @@ export async function resolveImportPath( slug: string, fileName: string, searchR
 	debug( `Will try to resolve path - ${ fileName }` );
 	let resolvedPath = resolvePath( fileName );
 
+	const instancePath = getEnvironmentPath( slug );
+
+	debug( 'Instance path for', slug, 'is:', instancePath );
```
mjangda:
```suggestion
	debug( `Instance path for ${ slug } is ${ instancePath }` );
```

jesse-greathouse (approved):

PR #1163, opened 2022-10-31
Pull correct environment's details from API when displaying Multisite SQL Import preflight details

saroshaga:
## Description

In the context of a Multisite's SQL Import, when importing data into a non-production environment, the command's preflight checks memo does not reflect the correct information about the data to be imported. 

This PR fixes this issue by pulling data from the API for the environment selected by the user.


aagam-shah (approved):
🚢 Thanks for looking into this and fixing the issue!

PR #1164, opened 2022-11-01
Fix regExp to categorize multisite tables during SQL Imports

saroshaga:
## Description

The Regular Expressions for selecting the tables to display in the SQL Import prompt, for a multisite is incorrect and fails for an edge case.

Here's an illustration:

image

In the example above, the current regular expression is incorrectly selecting `wp_3_somethingwp_index` as a table of Blog 1 because it matches the end of the string.


aagam-shah (approved):
🚢

PR #1165, opened 2022-11-01
Add caching type policy for `WPSite` data

saroshaga:
## Description

The ApolloClient's in-memory cache will copy objects into other objects if they have the same value in the `id` or `_id` field.

ApolloClient issue: https://github.com/apollographql/apollo-client/issues/605

Docs: https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids


aagam-shah (approved):
Nice find 🚀

PR #1167, opened 2022-11-03
dev-env: Fix bug in getVersionList() when path doesn't exist

rebeccahum:
## Description

> fetchWordPressVersionList failed to retrieve an updated version list

This occurs if the `vip` folder hasn't been created yet.

## Steps to Test
0) `rm -rf /Users/<user>/.local/share/vip`
1) `npm run build`
2) `./dist/bin/vip-dev-env-create.js`
3) Do the below options:
```
✔ WordPress site title · VIP Dev
✔ Multisite (y/N) · false
✔ PHP version to use · 8.0
```
4) At `? WordPress - Which version would you like …`, ensure there's no error being thrown


jesse-greathouse (approved):
LGTM 👍

PR #1170, opened 2022-11-10
[dev-env] Substitute the check for forward slash to with a cross-platform regex when processing component argument value

rinatkhaziev:
## Description

In #157122-z a customer reported that they don't see their code being properly mounted when using `vip dev-env create --app-code C:\Users\User\code`. 

Upon closer examination it turned out that `processComponentOptionInput` that processes wordpress ,mu-plugins, app-code used a simplistic check for forward slash to establish whether the path was valid or not. Windows supports both types of slashes. 

E.g. 

`C:\path\to\something`  and `C:/path/to/something` are equal and would resolve to `C:\path\to\something`.

This PR modifies the check to include both types of slashes.
Additionally this PR adds the output for pre-selected option.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `npm link`
1. On Windows: Run `vip dev-env create --app-code C:\My\Windows\Path\to\repo`
1. On Windows: Run `vip dev-env create --app-code C:/My/Windows/Path/to/repo`
1. On Mac/Linux: Run `vip dev-env create --app-code ~/path/to/repo`
1. Verify that .lando.yml has the correct mounts defined for client-code

A basic unit test case was added for Windows paths

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-cli.js
+++ b/__tests__/lib/dev-environment/dev-environment-cli.js
@@ -7,6 +7,7 @@
  */
 import { prompt, selectRunMock, confirmRunMock } from 'enquirer';
 import nock from 'nock';
+import os from 'os';
 /**
  * Internal dependencies
  */
@@ -150,14 +151,22 @@ describe( 'lib/dev-environment/dev-environment-cli', () => {
 					tag: '/tmp/wp',
 				},
 			},
-			{ // if local is  allowed
+			// if local is allowed
+			( os.platform() === 'win32' ? {
+				param: 'C:\path',
+				allowLocal: true,
+				expected: {
+					mode: 'local',
+					dir: 'C:\path',
+				},
+			} : {
 				param: '~/path',
 				allowLocal: true,
 				expected: {
 					mode: 'local',
 					dir: '~/path',
 				},
-			},
+			} ),
 		] )( 'should process options and use defaults', async input => {
 			const result = processComponentOptionInput( input.param, input.allowLocal );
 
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -179,7 +190,8 @@ export function printTable( data: Object ) {
 export function processComponentOptionInput( passedParam: string, allowLocal: boolean ): ComponentConfig {
 	// cast to string
 	const param = passedParam + '';
-	if ( allowLocal && param.includes( '/' ) ) {
+	// This is a bit of a naive check
+	if ( allowLocal && param.includes( path.sep ) ) {
```
sjinks: What do you think about this:

```suggestion
	if ( allowLocal && /[\\\/]/.test(param) ) {
```

because on Windows, you can use both forward and back slashes :-) `C:/path/to/something` also works.

__tests__/lib/dev-environment/dev-environment-cli.js
```diff
@@ -150,14 +151,22 @@ describe( 'lib/dev-environment/dev-environment-cli', () => {
...
 				},
 			},
-			{ // if local is  allowed
+			// if local is allowed
+			{ ( os.platform() == 'win32' ? { 
+				param: 'C:\path',
+				allowLocal: true,
+				expected: {
+					mode: 'local',
+					dir: 'C:\path',
```
sjinks:
```suggestion
				param: 'C:\\path',
				allowLocal: true,
				expected: {
					mode: 'local',
					dir: 'C:\\path',
```

because

```
console.log('C:\path');
C:path
```

src/lib/dev-environment/dev-environment-cli.js
rinatkhaziev: That totally missed me, I'm gonna try and test it tomorrow

sjinks (approved):

PR #1173, opened 2022-11-17
Add dev-env configuration file

alecgeatches:
vip-cli-basic-demo

## Description

Adds a configuration file for `vip dev-env` like [`.wp-env.json` for `wp-env`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#wp-env-json). Here's an example configuration file (altered since  the video above):

```yaml
configuration-version: 0.preview-unstable
slug: go-skeleton
title: VIP Go Skeleton
php: 8.0
wordpress: 6.1
xdebug: false
mailhog: false
multisite: false
mu-plugins: image
app-code: ./
elasticsearch: false
phpmyadmin: false
```

This feature is intended to solve a few use-cases:

- When writing documentation for `vip dev-env`, a configuration file can be provided to skip the wizard and reduce setup friction.
- Teams using `vip dev-env` for development can share a source-controlled set of development environment settings
- Makes managing multiple slugs for multiple environments easier

## Design choices

I tried to limit the scope of this PR to provide a useful configuration file, but not one that provides all the features I'd like to see. I've documented those in the next section for future improvement.

[See the discussion below](https://github.com/Automattic/vip-cli/pull/1173#issuecomment-1318083531) for how configuration file options are used. In short, during `vip dev-env create` and `vip dev-env update` the configuration file options from `.vip-dev-env.yml` are merged with command-line parameters and treated the same. If any options are missing, the wizard will prompt for missing options during `vip dev-env start`. If command-line parameters are used with a configuration file, command-line parameters override configuration file options.

For all other `vip dev-env` commands, the `slug` parameter is pulled from the configuration file if present in the current directory, unless a `--slug` is manually specified.

Currently the majority of configuration file functionality is provided in `dev-environment-configuration-file.js`, rather than extending the existing `-cli`/`-core` modules. This was easier for me during development and keeps files smaller, but if desired I'd be happy to move the changes into the existing modules.

## Future feature ideas

This PR is missing some features that might be good for future inclusion:

- Combine `vip dev-env create` and `vip dev-env start`. It would be great if a single command could do both parts when a configuration file is present like `wp-env start`.
- When a new `create` option is added but not yet present in the configuration file, this causes the wizard to prompt during `vip dev-env create`. Ideally, it would be great to have a `useDefaults: true` option or just default to automatically use sane defaults for features not explicitly versioned. This would be helpful for quickly starting up an environment and refining options as needed (also like `wp-env`).
- Provide environment variable configuration via configuration file, like:

    ```yaml
    slug: my-site
    env:
      SOME_VAR: some-value
    ```
- Similar to above, provide `wp-config.php` constant values like [`wp-env.json`'s `config` option](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#wp-env-json).
    ```yaml
    slug: my-site
    config:
      WP_SCRIPT_DEBUG: true
    ```
- In `vip dev-env update`, show specifically which configuration values have changed.
- (maybe) VIP could auto-generate these files for a customer's environment to make initial setup easy.
- (maybe) Search for configuration file in all parent directories like `npm` and `docker-compose`. The current version only searches the current directory.

## Steps to Test

These changes can be tested globally via `npm run build && npm link` using the `add/configuration-file` branch, or commands can be run directly from compiled files like this:

```bash
$ cd vip-cli
$ npm run build
$ node dist/bin/vip-dev-env-start.js
```

### `vip dev-env create`

1. In an existing VIP Go site source or using [vip-go-skeleton](https://github.com/Automattic/vip-go-skeleton), create a `.vip-dev-env.yml` file. Add a top-level slug option and at least one of the following options:

    ```yaml
    configuration-version: 0.preview-unstable
    slug: test-slug
    title: VIP Go Skeleton
    php: 8.0
    wordpress: 6.1
    xdebug: false
    mailhog: false
    multisite: false
    mu-plugins: image  # "image" or path to local mu-plugins/ directory
    app-code: ./       # "image" or path to local app code
    elasticsearch: false
    phpmyadmin: false
    ```

2. Run `vip dev-env create`. This should show all configuration from `.vip-dev-env.yml` that passes validation:

    ```bash
    $ vip dev-env create
    Using environment go-skeleton from .vip-dev-env.yml

    # ...

    Using configuration from file.
    configuration-version: 0.preview-unstable
    slug: wp-content-hub
    title: Content Hub
    multisite: false
    php: 8.0
    wordpress: 6.1
    mu-plugins: image
    app-code: ./
    elasticsearch: false
    phpmyadmin: false
    xdebug: false
    mailhog: false
    ```

3. Run `vip dev-env destroy`, and try removing some values from the configuration file.
4. Run `vip dev-env create` again, and confirm missing options are prompted by the wizard.
5. Run `vip dev-env start`, and ensure options (e.g. `elasticsearch: true`) are respected in the resulting provisioning.

### `vip dev-env update`

If changes are made to the configuration file after a `vip dev-env start`, these will not be propegated in the instance configuration until a `vip dev-env update` is run. This works exactly like passing command-line parameters to `vip dev-env update` - only options specified in the configuration file will be updated to match. If command-line parameters are also passed when a configuration file is present, those parameters will override the configuration file. To test:

1. Create a `.vip-dev-env.yml` file with configuration `phpmyadmin: false` in a VIP Go site source folder. See step 2 above for an example configuration file.
2. Run `vip dev-env create` and `vip dev-env start` to start the environment.
3. View the contents of the local `instance_data.json` file for the created environment:

    ```bash
    $ cat ~/.local/share/vip/dev-environment/<site-slug>/instance_data.json
    {"wpTitle":"VIP Go Skeleton","multisite":false,"elasticsearch":false,"php":"ghcr.io/automattic/vip-container-images/php-fpm:8.0","mariadb":"10.3","mediaRedirectDomain":"","wordpress":{"mode":"image","tag":"6.1"},"muPlugins":{"mode":"image","tag":"image"},"appCode":{"mode":"image","tag":"."},"statsd":false,"phpmyadmin":false,"xdebug":false,"xdebugConfig":"","siteSlug":"go-skeleton","mailhog":false}
    ```

    See the `"phpmyadmin"` key is set to `false`.

4. Edit `.vip-dev-env.yml` and turn on PHPMyAdmin with `phpmyadmin: true`.
5. Run `vip dev-env update`.
6. Reread instance data in `instance_data.json`:

    ```bash
    $ cat ~/.local/share/vip/dev-environment/<site-slug>/instance_data.json
    ```

    Ensure the `"phpmyadmin"` key is now set to `true`.

7. Run `vip dev-env start` to recreate the environment. Ensure that a `phpmyadmin` container is started and the info table contains working `PHPMYADMIN URLS`.

### Other commands

All other `vip dev-env` commands that require a slug (`vip dev-env info`, `vip dev-env exec`) will now pull the slug from a local configuration file if present via `getEnvironmentName()`. Ensure commands like `vip dev-env exec -- wp option get siteurl` work without a slug. If a slug is explicitly provided via command-line parameter, the configuration file slug should be ignored.


alecgeatches:
Update on my progress so far. Here's what it currently looks like:

dev-env-create-from-configuration

(note the "This is a wizard" line needs to be removed when nothing is being wizard configured)

---

Here's how I'm generally approaching this right now:

- Currently `vip dev-env create` can pull configuration normally selected in the wizard from `.vip-dev-env.yml`. The code works by [merging the configuration file with `preselectedOptions`](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/bin/vip-dev-env-create.js#L122), essentially working the same as filling in all settings values via command-line options.
- I've put all of the configuration-specific code in [`dev-environment-configuration-file.js`](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/lib/dev-environment/dev-environment-configuration-file.js) rather than adding to `dev-environment-cli`/`dev-environment-core`. I'm happy to move the code elsewhere if preferred.
- There is only [one slug allowed per `.vip-dev-env.yml`](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/lib/dev-environment/dev-environment-configuration-file.js#L83-L85), and if a slug is specified on the command line (`--slug`) [the configuration file is skipped](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/bin/vip-dev-env-create.js#L69-L74). I'm assuming if a command is run slugless it's referring to the current environment, but it's still possible to run commands against other slugs which bypasses the current configuration file.
- For parameters other than the slug, the configuration file [is overridden](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/lib/dev-environment/dev-environment-configuration-file.js#L155-L159) when specifying command-line values. The idea is that a temporary change can be made to the environment (e.g. adding `--xdebug`) and this should take precedence over the configuration file.
- The configuration file is processed in two stages:
    - [Sanitization](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/lib/dev-environment/dev-environment-configuration-file.js#L76), where settings are parsed from YAML before being displayed back to the user. This stage displays any logic / syntax errors.
    - [Merging](https://github.com/Automattic/vip-cli/blob/212a25d0adf565b24bd569528346c99995f6f6af/src/lib/dev-environment/dev-environment-configuration-file.js#L116), where the sanitized settings are mapped and merged with `preselectedOptions` to setup wizard defaults.
 
    This is a bit repetitive in code, but I've split this up into two sections so the configuration file format does not need to exactly match the internal `preselectedOptions` structure. `InstanceOptions` and `ConfigurationFileOptions` are related but different types. For a simple example, we can have `mu-plugins` as an option instead of `muPlugins` by remapping during merging. We also have the ability to have one configuration file option map to multiple internal options if desired.

- The configuration file code sanitizes but doesn't validate any of the values provided to it, and leans on existing `dev-env` code for PHP version/WordPress version/directory structure validation. If configuration values are incorrect, they fail validation during `vip dev-env create` or `vip dev-env start` and the wizard kicks in to provide a valid option. This avoids needing to duplicate validation code for parsing configuration files and reuses the existing error paths.

src/lib/dev-environment/types.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 export interface InstanceOptions {
-	title: string;
-	multisite: boolean;
+	title?: string;
```
alecgeatches, line 3: Changed these `InstanceOptions` properties to optional since they [can be missing and are treated as optional by the code](https://github.com/Automattic/vip-cli/blob/4de9a92/src/lib/dev-environment/dev-environment-cli.js#L277-L278). This made merging with `ConfigurationFileOptions` easier and sated `flow` squiggles.

sjinks:
I am going to add some E2E tests and then will merge this.

mjangda:
This is so awesome! Nice work @alecgeatches!

> then will merge this

@alecgeatches @sjinks What do you think about introducing this as a "preview" / "beta" feature? Specifically, noting that the naming / structure / format of the file and options might change based on feedback we get.

My biggest worry is that once users start adopting the this functionality, any changes we need to make to the config file will be much harder. By flagging it as a preview feature, we can do some feedback / outreach sessions to make sure this works in the way our users expect.

Tagging in @Crixu (since you've been looking into how to make setup / onboarding for Dev Env faster and easier) and @rinatkhaziev (since you had good thoughts on this).

sjinks:
I would add a "version" field (like docker-compose does) so that we can introduce new features relatively painlessly

__tests__/lib/dev-environment/dev-environment-cli.js
```diff
@@ -115,20 +115,18 @@ describe( 'lib/dev-environment/dev-environment-cli', () => {
...
+		it( 'should throw an exception if used the app.env when not allowed', () => {
 			const options = {
 				allowAppEnv: false,
 				app: '123',
 				env: 'bar',
 			};
 
 			const expectedErrorMessage = "This command does not support @app.env notation. Use '--slug=123-bar' to target the local environment.";
-			await expect( getEnvironmentName( options ) ).rejects.toThrow( expectedErrorMessage );
+			return expect( getEnvironmentName( options ) ).rejects.toThrow( expectedErrorMessage );
```
alecgeatches, line 139: @sjinks Thank you for cleaning these tests up!

alecgeatches:
> @alecgeatches @sjinks What do you think about introducing this as a "preview" / "beta" feature?

> I would add a "version" field (like docker-compose does) so that we can introduce new features relatively painlessly

This is a great idea. I think we'll need to make this required to avoid any sort of version auto-detection, so a baseline `.vip-dev-env.yml` would look like this:

```yaml
my-site:
  version: 1
```

Maybe we can start with `version: preview` to make that explicit? I'll add something like this today.

alecgeatches:
Added a required `version` key in https://github.com/Automattic/vip-cli/pull/1173/commits/0464b1d45552dc2ec1170db928a0fd54c5ed9a9e. The current version is called `0.preview-unstable`. My thought here is that we can drop this "valid" version after we have a bit more testing/feedback, and later replace it with a version `1`.

Here are the error messages for a missing/invalid version respectively:

Screenshot 2022-12-30 at 10 43 11 AM

---

Screenshot 2022-12-30 at 10 42 45 AM

sjinks:
@alecgeatches maybe we should make version a top level key, what do you think? In my opinion, it will be less confusing - technically, the version key relates to the while file, not the environment itself...

alecgeatches:
> maybe we should make version a top level key, what do you think?

@sjinks You're right, that makes more sense, I'll change it.

alecgeatches:
@sjinks Okay! I've pushed up the changes to make `version` a top-level key. A basic configuration file now looks like this:

```yaml
version: 0.preview-unstable
decoupled:
  title: Decoupled
  xdebug: false
  mailhog: false
  multisite: false
  php: 8.0
  wordpress: 6.1
  mu-plugins: image
  app-code: ./wordpress-site
  elasticsearch: false
  phpmyadmin: false
```

Another way I considered would be to add a second layer of hierarchy to avoid [figuring out which key is the slug](https://github.com/Automattic/vip-cli/blob/7be3078600ce0f2c591edab97d1458b1d145ba01/src/lib/dev-environment/dev-environment-configuration-file.js#L96), something like this:

```yaml
version: 0.preview-unstable
environments:
  decoupled:
    title: Decoupled
    xdebug: false
    # ...
```

However, I don't think that gives us any immediate advantages (as we're only handling a single configuration per-slug at this point), and it also strays further from the [super simple `.wp-env.json` idea](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/#a-complete-testing-environment) this is based on. I'm hoping with some careful choices (and the `version` field) we can keep this configuration file simple as we add more features.

alecgeatches:
@sjinks @mjangda After some thought, I pushed up a change that flattens the configuration file like this:

```yaml
version: 0.preview-unstable

slug: wp-content-hub
title: Content Hub
app-code: ./
mu-plugins: image
elasticsearch: false
# ...
```

I think this makes the structure easier to remember, and will allow for the future option to have defaults automatically applied. For example, this should be a valid bare-bones configuration file:

```yaml
version: 0.preview-unstable
slug: wp-content-hub
```

However, there's no great way to do this with the previous format:

```yaml
version: 0.preview-unstable
wp-content-hub:
  # ?
```

I think flattening the whole file makes more sense here, it maps directly to the existing `--slug` option, and better defines the goal of keeping the file structure simple.

The downside is that the `version` key is now at the same level as environment-specific keys, but I think we can avoid confusion by always putting it at the top of the file with a newline in examples.

Let me know if you think we should revert this change. Thanks!

alecgeatches:
One more tweak, changed the `version` key to `configuration-version`, which should help differentiate it in the flattened format.

alecgeatches:
> I am going to add some E2E tests and then will merge this.

@sjinks If you haven't worked on this yet, I'm happy to have a go at these tests to get this ready for merging. Can I take these on?

sjinks:
Yes, please - I am on a sick leave atm ☹️

alecgeatches:
@sjinks I finally made some time and wrote some e2e tests! They are ready for review.

The new tests are in [011-configuration-file.spec.js](https://github.com/Automattic/vip-cli/blob/29c332b890200a48e2e1bcc8721ba103bdb7cc61/__tests__/devenv-e2e/011-configuration-file.spec.js). I chose to put these in a separate spec file instead of adding to the existing e2e tests because these use configuration-file specific `tmpWorkingDirectoryPath` setup and teardown for `.vip-dev-env.yml` files used in the e2e tests.

Right now there are only 4 tests mostly based around `vip dev-env create`. Technically the configuration file could be used to test every other command (`destroy`/`start`/`stop`/etc) but the majority of the special logic for the configuration file happens in `create`. Most other commands just pull the slug from the file. I didn't want to add too many e2e tests for this one feature, but I'm happy to remove or add additional tests if you have better examples to use. Thank you!

alecgeatches:
Added one more test here for `vip dev-env update` from a configuration file.

sjinks:
OK, let's have this merged then.

sjinks (approved):

PR #1175, opened 2022-11-18
Dev-env: Throw error if more than one environment found when starting

rebeccahum:
## Description

When we create more than one environment and use `vip dev-env start`, it is confusing because it starts by the default (aka first) environment.  
Screenshot 2022-11-18 at 2 55 48 PM

## Steps to Test

1) Create more than one environment with `vip dev-env create`
2) `npm run build`
3) `./dist/bin/vip-dev-env-start.js`

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -73,12 +73,6 @@ type WordPressTag = {
 }
 
 export async function startEnvironment( slug: string, options: StartEnvironmentOptions ): Promise<void> {
-	const allEnvNames = getAllEnvironmentNames();
-	if ( allEnvNames.length > 1 ) {
-		throw new Error( `More than one environment found: ${ chalk.blue.bold( allEnvNames.join( ', ' ) ) }.
-		Please re-run command with the --slug parameter: ${ chalk.bold( 'vip dev-env start --slug <env-name>' ) }` );
-	}
-
 	debug( 'Will start an environment', slug );
 
 	const instancePath = getEnvironmentPath( slug );
@@ -347,7 +341,7 @@ async function prepareLandoEnv( instanceData: InstanceData, instancePath: string
 	debug( `Instance data file created in ${ instanceDataTargetPath }` );
 }
 
-function getAllEnvironmentNames() {
+export function getAllEnvironmentNames() {
 	const mainEnvironmentPath = xdgBasedir.data || os.tmpdir();
 
 	const baseDir = path.join( mainEnvironmentPath, 'vip', 'dev-environment' );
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-core.js
```diff
@@ -73,6 +73,12 @@ type WordPressTag = {
 }
 
 export async function startEnvironment( slug: string, options: StartEnvironmentOptions ): Promise<void> {
+	const allEnvNames = getAllEnvironmentNames();
+	if ( allEnvNames.length > 1 ) {
```
sjinks: What do you think about this (I am not insisting on it):

```suggestion
	if ( allEnvNames.length > 1 && ! allEnvNames.includes( 'vip-local' ) ) {
```

The reason is that when I create a slugless environment, I do not expect to be forced to specify `--slug vip-local` parameter

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -73,6 +73,12 @@ type WordPressTag = {
...
+		throw new Error( `More than one environment found: ${ chalk.blue.bold( allEnvNames.join( ', ' ) ) }.
+		Please re-run command with the --slug parameter: ${ chalk.bold( 'vip dev-env start --slug <env-name>' ) }` );
+	}
+
```
sjinks: Actually… Isn't `startEnvironment()` a bit too late for this check?

In [`vip-dev-env-start.js`](https://github.com/Automattic/vip-cli/blob/2beafce9eaff32a8d472afed9a0e19f7f976ce04/src/bin/vip-dev-env-start.js#L42):
```js
...
await startEnvironment( slug, options );
```

The slug is resolved in [`getEnvironmentName()`](https://github.com/Automattic/vip-cli/blob/2beafce9eaff32a8d472afed9a0e19f7f976ce04/src/lib/dev-environment/dev-environment-cli.js#L145) and is then passed to `startEnvironment()`. If we do the check here, we will reject any name if the user has more than one environment.

sjinks: I'd say that we probably need to move this check to `getEnvironmentName()`. What do you think?

rebeccahum: Yes...I was actually thinking about this last night before I went to bed. It's def too late for it. Will re-visit.

src/lib/dev-environment/dev-environment-core.js
rebeccahum: I think we should have to specify it if there's more than one environment...or else how would we know which one to target?

sjinks: Well, I mean I can have:

```bash
vip dev-env create --slug=test1
vip dev-env create --slug=test2
```

Those are smth like "permanent environments" where I work on something. Then, when I want to test something quickly, I sometimes do `vip dev-env create && vip-dev-env start`. This starts my "anonymous" (`vip-local`) environment.

sjinks:
```
vip dev-env create --slug test1
vip dev-env start
```

The second command will fail if there is no "anonymous" environment

sjinks (approved):

PR #1176, opened 2022-11-18
[dev-env] Don't try to pull Dev Env images if connectivity is missing or last pull happened less than a week ago

rinatkhaziev:
## Description

We were forcing image pull every time on `start`, in situations where ghcr.io (where we store images) is not resolvable it led to start command dying, even though the images could be downloaded and available locally.

This PR addresses this behavior by only forcing pull if the previous pull was more than a week ago and `ghcr.io` is resolvable. 

We're using `pre-engine-build` hook to force Lando to thing the defined services are local, rather than pullable. 

In practice this means that if an image was recently downloaded we won't try to download it. If an image is missing it would try to pull. If a registry is resolvable this "just works", if it's not resolvable it'll die with an error on the image that's missing. 

## Steps to Test

1. Check out the PR, build and `npm link`
2. Create an env
3. Start it, wait for it to be fully set up
4. Disconnect from the internet
5. Restart the environment and verify that it starts just fine even without the connection.

## Props

@sjinks for proposing a simpler solution not requiring a Lando fork patch.


package.json
```diff
@@ -127,9 +127,10 @@
     "graphql-tag": "2.12.5",
     "https-proxy-agent": "^5.0.1",
     "ini": "2.0.0",
+    "is-online": "^9.0.1",
     "json2csv": "5.0.7",
     "jwt-decode": "2.2.0",
-    "lando": "git+https://github.com/Automattic/lando-cli.git#v3.5.2-patch2022_09_05",
+    "lando": "git+https://github.com/Automattic/lando-cli.git#22d28f3de90814536636f8547a3966805343e693",
```
rinatkhaziev: Ref to https://github.com/Automattic/vip-cli/pull/1176/files

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -100,6 +103,32 @@ export async function landoRebuild( instancePath: string ) {
...
+
+	/**
+	 * Prevent the rebuild from fetching the containers all the time:
+	 * Only do so if we're online and last pull was more than 1 week ago
+	*/
+	app.events.on( 'pre-rebuild', 1, async () => {
+		const instanceData = readEnvironmentData( app._name );
+		const week = 7 * 24 * 60 * 60 * 1000;
+		let shouldPull = null;
+		let overrideOpts = { pull: false, nocache: false };
```
sjinks:
```suggestion
		let shouldPull = false;
		const overrideOpts = { pull: false, nocache: false };
```

sjinks (approved):

PR #1184, opened 2022-11-28
Validate Preflight: Allow preflight checks to execute without explicit app and/or env

vaurdan:
## Description

This PR intends to make the preflight checks available to be executed without an explicit application/env, for when the user don't have an application for the specific codebase (new apps, for example), or if it doesn't have the required permissions.

## Steps to Test

1. Run `npm run build`
1. Run `./dist/bin/vip-validate-preflight.js`
1. The command should output a warning, and show a selector for the Node version (the node version can also be manually specified with `-n <major>`)

image


chriszarate (approved):
Looks good

PR #1189, opened 2022-12-05
refactor(dev-env): Bootstrap Lando only once

sjinks:
## Description

This PR‌ refactors the way we instantiate Lando to speed things up.

How it used to be:
* we instantiated and bootstrapped Lando and its ecosystem twice during sanity checks (in `validateDockerInstalled()` and `validateDockerAccess()`);
* then we instantiated and bootstrapped Lando at least once again when doing what the user has asked us to do (e.g., start or stop the environment);
* if we had to run a command inside a container (`vip dev-env import sql`), we bootstrapped Lando for every command.

How we do it now:
* we instantiate and bootstrap Lando only once: right before we run the sanity checks;
* we pass the bootstrapped Lando instance to the functions used to bootstrap Lando.

The benefits:
* things are faster now. For example, `vip dev-env list` with one environment takes on average 0.5 seconds less to execute.

As a side effect, this PR fixes mocks in `dev-environment-core.js` so that the tests do not depend on each other when it comes to mocked implementations of `fs` functions.

## Steps to Test

Apply the patch and try dev-env commands.

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-core.js
+++ b/__tests__/lib/dev-environment/dev-environment-core.js
@@ -25,22 +25,28 @@ import { getEnvironmentPath,
 import { searchAndReplace } from '../../../src/lib/search-and-replace';
 import { resolvePath } from '../../../src/lib/dev-environment/dev-environment-cli';
 import { DEV_ENVIRONMENT_NOT_FOUND } from '../../../src/lib/constants/dev-environment';
+import { bootstrapLando } from '../../../src/lib/dev-environment/dev-environment-lando';
 
 jest.mock( 'xdg-basedir', () => ( {} ) );
-jest.mock( 'fs' );
 jest.mock( '../../../src/lib/api/app' );
 jest.mock( '../../../src/lib/search-and-replace' );
 jest.mock( '../../../src/lib/dev-environment/dev-environment-cli' );
 
 describe( 'lib/dev-environment/dev-environment-core', () => {
+	const cleanup = () => fs.rmSync( path.join( os.tmpdir(), 'lando' ), { recursive: true, force: true } );
+
+	beforeAll( cleanup );
+	afterAll( cleanup );
+
 	beforeEach( () => {
-		jest.clearAllMocks();
+		jest.resetAllMocks();
+		jest.restoreAllMocks();
 	} );
 
 	describe( 'createEnvironment', () => {
 		it( 'should throw for existing folder', async () => {
 			const slug = 'foo';
-			fs.existsSync.mockReturnValue( true );
+			jest.spyOn( fs, 'existsSync' ).mockReturnValueOnce( true );
 
 			const promise = createEnvironment( { siteSlug: slug } );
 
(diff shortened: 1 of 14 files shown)


__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -25,22 +25,28 @@ import { getEnvironmentPath,
 import { searchAndReplace } from '../../../src/lib/search-and-replace';
 import { resolvePath } from '../../../src/lib/dev-environment/dev-environment-cli';
 import { DEV_ENVIRONMENT_NOT_FOUND } from '../../../src/lib/constants/dev-environment';
+import { bootstrapLando } from '../../../src/lib/dev-environment/dev-environment-lando';
 
 jest.mock( 'xdg-basedir', () => ( {} ) );
-jest.mock( 'fs' );
```
sjinks, line 30: Now that we call `bootstrapLando()` before calling our Lando wrappers, we cannot safely mock the `fs` module because Lando needs it.

Instead, we are going to mock only those functions we are interested in.

__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -25,22 +25,28 @@ import { getEnvironmentPath,
...
 jest.mock( '../../../src/lib/api/app' );
 jest.mock( '../../../src/lib/search-and-replace' );
 jest.mock( '../../../src/lib/dev-environment/dev-environment-cli' );
 
 describe( 'lib/dev-environment/dev-environment-core', () => {
+	const cleanup = () => fs.rmSync( path.join( os.tmpdir(), 'lando' ), { recursive: true, force: true } );
+
+	beforeAll( cleanup );
+	afterAll( cleanup );
```
sjinks, line 39: When we have an existing `/tmp/lando` directory, this sometimes leads to interesting situations when tests pass locally and fail on CI.

Make sure we always start from a clean slate to be able to catch mocking bugs.

__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -25,22 +25,28 @@ import { getEnvironmentPath,
...
+
 	beforeEach( () => {
-		jest.clearAllMocks();
+		jest.resetAllMocks();
+		jest.restoreAllMocks();
```
sjinks, line 43: Ensure that we clear what we have mocked after every test. This way, our tests won't depend on which functions the previously run tests have mocked. All tests must be isolated. While this leads to slightly more code, we get reproducible and predictable behavior in return.

__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -52,9 +58,18 @@ describe( 'lib/dev-environment/dev-environment-core', () => {
...
+			const expectedPath = getEnvironmentPath( slug );
+			fs._originalExistsSync = fs.existsSync;
+			jest.spyOn( fs, 'existsSync' ).mockImplementation( fpath => {
+				if ( fpath === expectedPath ) {
+					return false;
+				}
 
-			const promise = startEnvironment( slug );
+				return fs._originalExistsSync( fpath );
+			} );
```
sjinks, line 69: This complication happens because we bootstrap Lando before we call `startEnvironment()`.

`startEnvironment()` used to run some checks before it instantiated Lando, and that is why the tests worked. But now we need Lando to bootstrap first; therefore, we mock only the case when the path checked is our dev environment.

__tests__/lib/dev-environment/dev-environment-core.js
```diff
@@ -302,7 +327,8 @@ describe( 'lib/dev-environment/dev-environment-core', () => {
 
 		it( 'should handle windows path correctly', async () => {
 			path.sep = '\\';
-			fs.existsSync.mockReturnValue( true );
+			jest.spyOn( fs, 'existsSync' ).mockReturnValue( true );
+			jest.spyOn( fs, 'lstatSync' ).mockReturnValue( { isDirectory: () => false } );
```
sjinks, line 331: This is the case when the test depended on the previous test. The previous test set up a mock for `lstatSync()`, and because there was no cleanup logic, this test inherited the mock.

rinatkhaziev (approved):
I didn't get a chance to test it but the change makes sense.

PR #1181, opened 2022-11-21
fix(dev-env): Set SQL_MODE to match our production environment

sjinks:
## Description

This PR sets the `SQL_MODE` to match what we have in our production environment.

See p9zhOE-2yQ for details.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-dev-env-create.js` and `./dist/bin/vip-dev-env-start.js`
4. Ensure the environment is up and running.
5. Try to run `insert into wp_posts set post_author = -1; \W`. If everything goes well, there will be no errors, just warnings.

PR #1180, opened 2022-11-21
fix(dev-env): Fix a typo in parameter description

sjinks:
## Description

This PR fixes a typo in the description of the `--skip-wp-versions-check` parameter

## Steps to Test

No testable changes.

PR #1179, opened 2022-11-21
feat(dev-env): Do not prompt for missing values if stdin is not a TTY

sjinks:
## Description

When we run `vip dev-env create` in a non-interactive environment (e.g., CI), it dies when it needs to prompt for a non-specified parameter:

```
$ vip dev-env create < /dev/
Running validation steps...
✓ Check for docker installation
✓ Check for docker-compose installation
✓ Check access to docker for current user
✓ Check DNS resolution
This is a wizard to help you set up your local dev environment.

Sensible default values were pre-selected for convenience. You may also choose to create multiple environments with different settings using the --slug option.


? WordPress site tvolodymyr@democratic-circus:~/$
```

This PR adds a check whether `stdin` is a TTY and if not, disables interactive prompts.

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. Run `./dist/bin/vip-dev-env-create < /dev/null`
4. Verify the environment has been created and can be started.

PR #1177, opened 2022-11-19
Logged out `vip --version`

mjangda:
Allow running `vip --version` when logged out.

Witthout this PR, we get thrown into the login flow:

```
vip --version

   _    __ ________         ________    ____
  | |  / //  _/ __        / ____/ /   /  _/
  | | / / / // /_/ /______/ /   / /    / /
  | |/ /_/ // ____//_____/ /___/ /____/ /
  |___//___/_/           ____/_____/___/

  VIP-CLI is your tool for interacting
```

With this PR, it shows the version:

```
node ./dist/bin/vip --version
2.23.0
```

## Steps to Test

1. Check out PR.
1. Run `npm run build`
2. Make sure you're logged out.
1. Test out various commands with `--version`: `node ./dist/bin/vip --version && node ./dist/bin/vip app --version && node ./dist/bin/vip app list --version` 
1. All should list out the version.
2. Try running a command without version (e.g. `node ./dist/bin/vip app` -- this should prompt you to log in.
3. Now log in with `node ./dist/bin/vip login`.
4. Now try `--version` again: `node ./dist/bin/vip --version && node ./dist/bin/vip app --version && node ./dist/bin/vip app list --version` 
5. All should still list the version.

PR #1190, opened 2022-12-06
Release 2.24.0

sjinks:
## 2.24.0 (06 Dec 2022)

- #1139 fix(dev-env): Fall back to copy when rename fails due to EXDEV
- #1170 [dev-env] Substitute the check for forward slash to with a cross-platform regex when processing component argument value
- #1174 Override WordPress image cache key in tests
- #1175 Dev-env: Throw error if more than one environment found when starting
- #1177 Logged out `vip --version`
- #1178 chore(deps): Fix some of vulnerable dependencies
- #1179 feat(dev-env): Do not prompt for missing values if stdin is not a TTY
- #1180 fix(dev-env): Fix a typo in parameter description
- #1181 fix(dev-env): Set SQL_MODE to match our production environment
- #1182 Dev-env: Throw UserError in getEnvironmentName() instead and default to first one if only one env
- #1184 Validate Preflight: Allow preflight checks to execute without explicit app and/or env
- #1185 feat(dev-env): Add support for MailHog
- #1186 feat(dev-env): Do not mount home directory into containers
- #1188 fix: Make bin files executables
- #1189 refactor(dev-env): Bootstrap Lando only once


vaurdan (approved):
LGTM :+1:

Issue #1192, opened 2022-12-08
Development environment SQL import fails


rbackert:
## Expected/Desired Behavior

Importing sql files into a local development environment [as documented](https://docs.wpvip.com/how-tos/dev-env-add-content/) succeeds.

## Actual Behavior

Importing sql files into a local development environment fails.

## Steps to Reproduce the Problem

1. Place a sql file at `$HOME/db.sql`
2. Run `vip dev-env --slug=project-slug import sql ~/db.sql`
3. See error: `Import file missing or not readable: /user/db.sql`

## Additional notes

`$HOME`is no longer mounted into `/user` as per https://github.com/Automattic/vip-cli/blob/develop/src/lib/dev-environment/dev-environment-lando.js#L68, `dev-env import sql` expects the files to be available at `/user` in the container.

sjinks:
Fixed in 2.24.1.

PR #1193, opened 2022-12-09
Fix regression in vip dev-env import sql related to the fact we don't mount User's home dir anymore

rinatkhaziev:
## Description

https://github.com/Automattic/vip-cli/pull/1186 introduced a regression where the import file became not accessible because we used the user mount rather than the path in the container. 

The root cause was that we tried to use the mounted path inside the container, even though we copied the file into the `app` mount. We've addressed this by using the proper in-container path when importing. 

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. `vip dev-env start`
1. `vip dev-env import ~/path/to/sql`
1. Wait for it to finish and make sure it was imported successfully.


sjinks (approved):

sjinks:
Fixes: #1192

PR #1195, opened 2022-12-10
fix(dev-env): increase max number of event listeners for AsyncEvents

sjinks:
## Description

This PR fixes these warnings:

```
(node:29693) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 65 post-init listeners added to [AsyncEvents]. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
(node:29693) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 65 post-start listeners added to [AsyncEvents]. Use emitter.setMaxListeners() to increase limit
```

which fire when running `vip dev-env import sql` command.

## Steps to Test

Apply the patch and make sure that `vip dev-env import sql` command does not produce these warnings anymore.

### DIFF
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -54,6 +54,7 @@ command( {
 		const slug = getEnvironmentName( opt );
 
 		const lando = await bootstrapLando();
+		lando.events.constructor.prototype.setMaxListeners( 100 );
 		await validateDependencies( lando, slug );
 
 		const trackingInfo = getEnvTrackingInfo( slug );


src/bin/vip-dev-env-import-sql.js
```diff
@@ -54,6 +54,7 @@ command( {
 		const slug = getEnvironmentName( opt );
 
 		const lando = await bootstrapLando();
+		lando.events.constructor.prototype.setMaxListeners( 100 );
```
sjinks, line 57: AsyncEvents [hardcodes](https://github.com/Automattic/lando-cli/blob/93b9137e25/lib/events.js#L149) the number of event listeners to 64.

We need a higher value because we issue several calls to exec().

rinatkhaziev (approved):

PR #1197, opened 2022-12-10
feat(dev-env): Pipe files to Docker instead of copying them when doing SQL import

sjinks:
## Description

`vip dev-env sql import` copies the SQL dump into the PHP container. This can be undesirable when the file is large: it creates unnecessary I/O, duplicates the file, and can potentially upset Docker if the copy operation exceeds the disk quota.

In this PR, we pipe the file to Docker instead.

## Steps to Test

Apply the patch and test all `vip dev-env sql import` scenarios.

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-core.js
+++ b/__tests__/lib/dev-environment/dev-environment-core.js
@@ -289,14 +289,8 @@ describe( 'lib/dev-environment/dev-environment-core', () => {
 		} );
 	} );
 	describe( 'resolveImportPath', () => {
-		afterEach( () => {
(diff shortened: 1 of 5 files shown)


rinatkhaziev:
I think if we stream directly to MariaDB it should be even better because we'll be able to skip the intermediary PHP process

sjinks:
True, but we don't know the MariaDB credentials in Lando.

sjinks:
@rinatkhaziev OK, please take a look :-)

rinatkhaziev (approved):
Test on Mac and Windows, works great :)

Issue #1191, opened 2022-12-07
SSL certificate is not verified inside dev environment


goldenapples:
I'm using VIP dev-env for a project, and noticing that the SSL cvertificate is not verified inside the dev environment. I can access the site over https from my host machine after accepting the security exception, but when making https requests inside the container, they always fail with the error message

```
cURL error 60: SSL certificate problem: unable to get local issuer certificate
```
The [Lando docs](https://docs.lando.dev/config/security.html#trusting-the-ca) seem to say that by default Lando should be automatically trusting the CA internally:

> While Lando will automatically trust this CA internally, it is up to you to trust it on your host machine. Doing so will alleviate browser warnings regarding certs we issue.

I notice that vip dev-env specifies a custom path for the Lando config which is inside the temp directory `/tmp/lando` as opposed to `~/.lando` which could be related. In any case, it means that the certificate is regenerated every time the container is started, which makes it more difficult to trust on the host machine.

I'm wondering if there's a workaround for this, or what it would take to fix this behavior?
 
---

## Expected/Desired Behavior

After starting a dev environment with `vip dev-env start`, the ssl certificate for all domains served by the site should be automatically trusted inside the PHP container, so that REST requests can work without setting `sslverify => false`. 

## Actual Behavior

Making REST requests from application code to the same site fail with a cURL SSL certificate problem, unless the sslverify option is set to false. 

## (Optional) Additional notes

System: Ubuntu 22.04
VIP-CLI version: 2.23.0
Node version: 16.18.1

f0rthelulz:
The following steps might be helpful:

- Get the ID of the Ngnix container from the list of containers or from the Docker app:
  - `docker ps -aqf "name=<slug>_nginx_1"`
- Copy the `undefined.pem` file out of the container
  - `docker cp <container_id>:/lando/certs/undefined.pem ~/`
- Trust the CA cert via the method of your choice: 
  - https://docs.lando.dev/config/security.html#trusting-the-ca
  - Mac, for example: `sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/undefined.pem`
- Restart the local dev environment 
  - `vip dev-env stop --slug=wp-nexstar-platform && vip dev-env start --slug=wp-nexstar-platform`

PR #1198, opened 2022-12-12
feat(dev-env): Use VIP CA to sign all X509 certificates

sjinks:
## Description

Related: #1191 

This PR‌ does a few things:
* First, it generates the CA data when the `postinstall` script is run;
* Then, it copies the generated CA files into Lando when it starts (currently, this happens upon `vip dev-env start`).

This allows us to use the same certificate authority for all development environments, eliminating the need to trust the CA every time `/tmp/lando` disappears.

## Steps to Test

**Existing environments**

1. Make sure you have a test environment created
2. Apply the patch, make sure to run `node helpers/generate-keys.js`; this should create `lndo.site.pem` and `lndo.site.key` in `$XDG_BASE_DIR/vip`
3. Start and restart your test environment; everything should work
4. Stop the environment, remove `/tmp/lando` (or wherever it is located on Macs), then start and restart the environment. Everything should work.

**New environments**

Same steps as above except you need to create a new environment after you apply the patch.

Also, if you view the certificate data, the certificate issuer should mention VIP now:

ssl-cert


sjinks:
Closing in favor of #1228

PR #1200, opened 2022-12-13
fix(dev-env): Reuse Lando App instead of recreating it every time

sjinks:
## Description

This PR brings in some performance improvements.

Every time we call a function from `dev-environment-lando.js`, that function usually gets a fresh copy of Lando's App and initializes it. This is fine if we don't have to call more than one function, but is problematic when we have to call several functions (e.g., our script calls `landoExec` to do something).

This PR ensures that the app is created and initialized only once (per instance path). This reduces I/O and memory consumption (as Lando won't create event listeners like crazy).

How it used to be:

```
$ DO_NOT_TRACK=1 DEBUG="@automattic/vip:bin:dev-environment" ./dist/bin/vip-dev-env-import-sql.js test.sql 2>&1 | grep -F 'initiatilizing app'
vip-local 01:53:55 VERBOSE ==> initiatilizing app at /home/volodymyr/.local/share/vip/dev-environment/vip-local...
vip-local 01:53:58 VERBOSE ==> initiatilizing app at /home/volodymyr/.local/share/vip/dev-environment/vip-local...
vip-local 01:53:59 VERBOSE ==> initiatilizing app at /home/volodymyr/.local/share/vip/dev-environment/vip-local...
vip-local 01:54:07 VERBOSE ==> initiatilizing app at /home/volodymyr/.local/share/vip/dev-environment/vip-local...

$ DO_NOT_TRACK=1 time ./dist/bin/vip-dev-env-import-sql.js test.sql
...
1.17user 0.15system 0:06.54elapsed 20%CPU (0avgtext+0avgdata 143748maxresident)k
0inputs+86568outputs (0major+118314minor)pagefaults 0swaps
```

Now:
```
$ DO_NOT_TRACK=1 DEBUG="@automattic/vip:bin:dev-environment" ./dist/bin/vip-dev-env-import-sql.js test.sql 2>&1 | grep -F 'initiatilizing app'
vip-local 02:09:36 VERBOSE ==> initiatilizing app at /home/volodymyr/.local/share/vip/dev-environment/vip-local...

$ DO_NOT_TRACK=1 time ./dist/bin/vip-dev-env-import-sql.js test.sql
...
0.73user 0.18system 0:06.02elapsed 15%CPU (0avgtext+0avgdata 141040maxresident)k
160inputs+82976outputs (1major+92091minor)pagefaults 0swaps
```

Also,

```
DO_NOT_TRACK=1 DEBUG="@automattic/vip:bin:dev-environment" ./dist/bin/vip-dev-env-import-sql.js test.sql > log.txt 2>&1
```

The size of `log.txt`:
* Before: 135,362 bytes
* After: 35,624 bytes

## Steps to Test

Apply the patch and ensure that dev env commands still work.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -70,6 +70,24 @@ function getLandoConfig() {
 	};
 }
 
+const appMap: Map<string, App> = new Map();
+
+async function getLandoApplication( lando: Lando, instancePath: string ): Promise<App> {
+	if ( appMap.has( instancePath ) ) {
+		return Promise.resolve( appMap.get( instancePath ) );
+	}
+
+	const app = lando.getApp( instancePath );
+	addHooks( app, lando );
+	appMap.set( instancePath, app );
+
+	if ( ! app.initialized ) {
+		await app.init();
+	}
+
+	return app;
+}
+
 export async function bootstrapLando(): Promise<Lando> {
 	const lando = new Lando( getLandoConfig() );
 	await lando.bootstrap();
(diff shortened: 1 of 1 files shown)


src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -70,6 +70,24 @@ function getLandoConfig() {
 	};
 }
 
+const appMap: Map<string, App> = new Map();
+
+async function getLandoApplication( lando: Lando, instancePath: string ): Promise<App> {
+	if ( appMap.has( instancePath ) ) {
+		return Promise.resolve( appMap.get( instancePath ) );
```
rinatkhaziev, line 77: Since this is an async function wouldn't `return appMap.get( instancePath );` be sufficient?

sjinks: async functions return promises no matter what 🙂

Although a JS engine will convert a "plain" value into a promise, type checkers like Flow and TypeScript complain if we return something else than a promise.

rinatkhaziev: > async functions return promises no matter what

right, that's exactly what I meant :) 

> type checkers like Flow and TypeScript complain if we return something else than a promise.

ah I see

rinatkhaziev (approved):
I didn't do a thorough test but the code looks fine.

PR #1201, opened 2022-12-13
fix(dev-env): Do not check env is up for every action

sjinks:
## Description

This PR further improves #1200.

`landoExec()` checks if the environment is up every time it is called. This creates unnecessary load, I/O, and slows things down.

This PR extracts the `isEnvUp()` logic into a dedicated function (the Single Responsibility Principle is great) and gets rid of the `force` option.

This allows us to speed the application up even more. With this change, the size of the log file goes down to 29,011 bytes.

## Steps to Test

Apply the patch and verify that dev-env commands still work.

### DIFF
--- a/__tests__/devenv-e2e/009-import-media.spec.js
+++ b/__tests__/devenv-e2e/009-import-media.spec.js
@@ -53,7 +53,7 @@ describe( 'vip dev-env import media', () => {
 		expect( result.rc ).toBe( 0 );
 		expect( checkEnvExists( slug ) ).toBe( true );
 
-		result = await cliTest.spawn( [ process.argv[ 0 ], vipDevEnvImportMedia, '--slug', slug, __dirname ], { env } );
+		result = await cliTest.spawn( [ process.argv[ 0 ], vipDevEnvImportMedia, '--slug', slug, __dirname ], { env }, true );
 		expect( result.rc ).toBe( 0 );
 
 		const uploadsPath = path.join( getEnvironmentPath( slug ), 'uploads' );
--- a/__tests__/devenv-e2e/helpers/utils.js
+++ b/__tests__/devenv-e2e/helpers/utils.js
@@ -7,7 +7,7 @@ import { expect } from '@jest/globals';
 /**
  * Internal dependencies
  */
-import { doesEnvironmentExist } from '../../../src/lib/dev-environment/dev-environment-core';
+import { doesEnvironmentExist, getEnvironmentPath } from '../../../src/lib/dev-environment/dev-environment-core';
 import { vipDevEnvCreate, vipDevEnvDestroy, vipDevEnvStart } from './commands';
 
 let id = 0;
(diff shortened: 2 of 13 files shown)


rinatkhaziev:
Cursory testing shows that it doesn't work. A running environment is not recognized:

```
vip dev-env --slug what-mu exec -- wp option
get home
Running validation steps...
✓ Check for docker installation
✓ Check for docker-compose installation
✓ Check access to docker for current user
✓ Check DNS resolution
Error: Environment doesn't exist.


To create a new environment run:

vip dev-env create
```

but it was running, and now restarting:

```
 vip dev-env --slug what-mu start
Running validation steps...
✓ Check for docker installation
✓ Check for docker-compose installation
✓ Check access to docker for current user
✓ Check DNS resolution
The most recent WordPress version available is: 6.1
Environment WordPress version is: 6.1  ... 😎 nice!
Killing whatmu_nginx_1         ... done
Killing whatmu_php_1           ... done
Killing whatmu_memcached_1     ... done
Killing whatmu_wordpress_1     ... done
Killing whatmu_elasticsearch_1 ... done
Killing whatmu_database_1      ... done
Killing whatmu_devtools_1      ... done
```

If we wanna speed things up we probably should look into optimizing isEnvUp and also the pre-flight checks. Do we really need the dependency checks there and the DNS check?

rinatkhaziev (changes requested):
Please see the comment.

sjinks:
> Base automatically changed from fix/multi-init to develop 4 hours ago

Not all PRs are compatible :-)

sjinks:
> If we wanna speed things up we probably should look into optimizing isEnvUp

This is exactly what we are trying to do. In `develop`, `exec()` calls `isEnvUp()` on every invocation. And, before #1200 was merged, it (re)initialized Lando every time. Now, we call `isEnvUp()` in three places: `vip dev-env exec`, `vip dev-env import sql`, and `vip dev-env info`

> Do we really need the dependency checks there and the DNS check?

TBH, I would only leave these checks in `vip dev-env start`, assuming that the user needs to start the environment first before doing something else with it. However, if the user runs any other command before starting the environment, and there is a misconfiguration, they will probably see a cryptic message (which was the reason why we introduced these checks).

However, the preflight checks are pretty fast:
```
@automattic/vip:bin:dev-environment Validation checks completed in 10 ms
```

`isEnvUp()` is slower (the actual duration will greatly depend on the load):
```
@automattic/vip:bin:dev-environment isEnvUp took 169 ms
```

sjinks:
See also: #1209

sjinks:
Waiting for #1203 to be merged first.

rinatkhaziev (approved):
Code looks fine but will do some manual testing.

rinatkhaziev:
I seemed to have ran into an issue due to checking out the branch but not doing `npm ci` but after I did that I'm not able to reproduce after doing `npm ci`

PR #1203, opened 2022-12-13
test(dev-env): Add E2E tests for dev env

sjinks:
## Description

This PR adds E2E tests for dev environment.

  vip dev-env create
    ✓ should create a new environment
    ✓ should fail on duplicate slugs
    ✓ should use sane defaults
    ✓ should be configurable via command line

  vip dev-env destroy
    ✓ should fail if environment does not exist
    ✓ should remove existing environment
    ✓ should remove existing environment even without landofile
    ✓ should keep the files when asked to
    if the environment is running
      ✓ should stop and destroy it

  vip dev-env start
    ✓ should fail if environment does not exist
    ✓ should start an environment

  vip dev-env stop
    ✓ should fail if environment does not exist
    ✓ should stop a running environment
    ✓ should not fail if the environment is stopped

  vip dev-env update
    ✓ should fail if the environment does not exist
    ✓ should not update the environment if there are no changes
    ✓ should update the environment
    ✓ should not update multisiteness

  vip dev-env list
    ✓ should handle no environments
    ✓ should list multiple environments
    for started environments
      ✓ should list them as UP

  vip dev-env info
    ✓ should fail on a non-existing environment
    ✓ should display the info
    for started environments
      ✓ should list them as UP

  vip dev-env exec
    if the environment does not exist
      ✓ should fail
    if the environment exists
      ✓ should run known commands
      ✓ should fail on unknown commands
      ✓ should not produce superfluous output in silent mode
      ✎ todo vip dev-env exec --force

  vip dev-env import media
    ✓ should fail if environment does not exist
    ✓ should copy files if environment exists

  vip dev-env import sql
    if the environment does not exist
      ✓ should fail
    if the environment exists
      ✓ should fail if the file fails validation
      ✓ should allow to bypass validation
      ✓ should correctly perform replace

## Steps to Test

The CI should pass.


__tests__/devenv-e2e/helpers/utils.js
```diff
@@ -0,0 +1,88 @@
...
+	expect( result.stdout ).toMatch( /STATUS\s+UP/u );
+}
+
+/**
+ * @param {import('./cli-test').CliTest} cliTest CLI Test instance
+ * @param {string} slug Environment slug
+ * @param {NodeJS.ProcessEnv} env Environment
+ * @param {bool} shouldSucceed Whether destruction should succeed
+ */
+export async function destroyEnvironment( cliTest, slug, env, shouldSucceed = true ) {
```
rebeccahum: I don't see the use case of `false` being passed into `shouldSucceed` anywhere. Is this for future?

sjinks: Yeah, incomplete refactoring :-( `CliTest.spawn()` uses `shouldSucceed` to see if it should print stderr on failure. I extracted the function but never checked if `false` is ever passed to it.

Fixed, thanks!

rebeccahum (approved):
Thanks for fixing!

PR #1209, opened 2022-12-15
refactor(dev-env): Run health checks in parallel

sjinks:
## Description

This PR improves performance of our healthcheck hook by running all checks in parallel.

Before: Healthcheck completed in 2789ms
After: Healthcheck completed in 478ms

(times may vary wildly)

## Steps to Test

Check that `vip dev-env start` still works.


rinatkhaziev (approved):
Thanks! Tested, works.

PR #1211, opened 2022-12-15
fix(dev-env)!: remove support for statsd

sjinks:
## Description

Because we are replacing statsd with Prometheus, we don't need stated anymore.

Ref: https://github.com/Automattic/vip-go-mu-plugins/pull/3960

## Steps to Test

Once the patch is applied, dev env should no longer start `statsd` container.


rinatkhaziev (approved):
YESSSSS

PR #1215, opened 2022-12-21
fix(dev-env): Import media should fail if the env does not exist

sjinks:
## Description

If the environment does not exists, `vip dev-env import media` still copies the files, messing up `vip dev-env list`.

This PR adds a check whether the environment exists before starting to copy the files.

## Steps to Test

`vip dev-env import media` should say "Error: Environment doesn't exist" if the environment does not exist.


sjinks:
This patch was already in #1203

PR #1216, opened 2022-12-21
fix(dev-env): Do not bootstrap Lando for `vip dev-env import media`

sjinks:
## Description

This PR removes the Lando bootstrap code for the `vip dev-env import media` command, becuase it does not use Lando not Docker.

## Steps to Test

`vip dev-env import media` should work the same way as it used to.


sjinks:
This patch was in #1203

PR #1225, opened 2022-12-24
chore: Update dot files

sjinks:
## Description

This PR:
* replaces `master` with `trunk` in our files;
* updates `.gitignore` and `.npmignore` (mostly with the missing dotfiles and `.vscode`);
* adds `publishConfig` to `package.json` (so that we don't have to type `--access public` when publishing the package);
* limits the number of workers for Flow (on 22 cores, this is a disaster).

## Steps to Test

No testable changes.


pandah3 (approved):
Awesome, thank you 👍

PR #1228, opened 2023-01-04
fix(dev-env): Do not use a temporary directory for Lando

### LINKED ISSUE #1191: SSL certificate is not verified inside dev environment

sjinks:
## Description

Related: #1191, #1198

This PR updates Lando configuration so that Lando's configuration perists reboots. Among other things (like faster warm bootstrap), this allows us to use the same certificate authority for all development environments, eliminating the need to trust the CA every time `/tmp/lando` disappears.

## Steps to Test

1. Create and start a test environment; stop it.
2. Apply the patch and start the test environment again; it should work.

*NB:* it may be necessary to restart all running environments after applying the patch.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -6,6 +6,7 @@
  */
 import debugLib from 'debug';
 import os from 'os';
+import fs from 'fs';
 import path from 'path';
 import Lando from 'lando/lib/lando';
 import landoUtils from 'lando/plugins/lando-core/lib/utils';
@@ -13,13 +14,14 @@ import landoBuildTask from 'lando/plugins/lando-tooling/lib/build';
 import chalk from 'chalk';
 import App from 'lando/lib/app';
 import dns from 'dns';
+import xdgBasedir from 'xdg-basedir';
 
 /**
  * Internal dependencies
  */
 import { doesEnvironmentExist, readEnvironmentData, writeEnvironmentData } from './dev-environment-core';
-import { DEV_ENVIRONMENT_NOT_FOUND } from '../constants/dev-environment';
 import UserError from '../user-error';
+import { DEV_ENVIRONMENT_NOT_FOUND } from '../constants/dev-environment';
 
 /**
  * This file will hold all the interactions with lando library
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -54,6 +44,16 @@ function getLandoConfig() {
...
 
+	const vipDir = path.join( xdgBasedir.data || os.tmpdir(), 'vip' );
+	const landoDir = path.join( vipDir, 'lando' );
+	const fakeHomeDir = path.join( landoDir, 'home' );
+
+	try {
+		await fs.promises.mkdir( fakeHomeDir, { recursive: true } );
+	} catch ( err ) {
+		// Ignore
+	}
```
sjinks, line 54: We do this to be compatible with `@lando/core` - until they accept our patch, at least.

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -108,8 +108,9 @@ function getLandoConfig() {
 			'@lando/wordpress',
 		],
 		proxyName: 'vip-dev-env-proxy',
-		userConfRoot: getLandoUserConfigurationRoot(),
-		home: '',
+		userConfRoot: landoDir,
+		home: fakeHomeDir,
+		domain: 'lndo.site',
```
sjinks, line 112: The `domain` key is used to name the certificate file; if we don't pass it, we'll get `undefined.pem`, which is a bit confusing.

src/lib/dev-environment/dev-environment-lando.js
rinatkhaziev: What's a fakeHomeDir?

sjinks: Empty directory to be mounted as `/home`. `@lando/core` will create the `.ssh` directory inside it. If we pass an empty string as `home` (see below), Lando will create `.ssh` directory in the current directory, which is not what we want.

In brief, we are trying to solve here the issue we solved in https://github.com/Automattic/lando-cli/pull/15, except that we don't want to fork `@lando/core`.

rinatkhaziev (approved):

PR #1229, opened 2023-01-05
fix(dev-env): Perform ES healthcheck properly

sjinks:
## Description

The ES service in the dev env is named `elasticsearch`, not `vip-search`. This PR updates healthchecks to reference the proper service.

## Steps to Test

It's not trivial as this really depends on many factors including how powerful the computer is. Ideally you should be able to see the "Waiting for service elasticsearch ..." message when starting the dev environment with ES enabled.


rinatkhaziev (approved):

rinatkhaziev:
thanks for the catch

PR #1232, opened 2023-01-06
fix(dev-env): Make environment existence check more robust

sjinks:
## Description

This PR makes the environment existence check more robust.

Previously, for an environment to exist, it was enough that a directory with the name matching the environment exists. It was enough to create an empty directory to crash the application.

Now we check that the `instance_data.json` file exists as well.

## Steps to Test

```
$ mkdir ~/.local/share/vip/dev-environment/xxx
$ vip dev-env list
...
WARN ==> could not find app in this dir or a reasonable amount of directories above it! 
Error: Cannot read properties of undefined (reading 'init')
```

```
$ npm run build
$ ./dist/bin/vip-dev-env-list.js
...
WARNING: "xxx" is not a valid environment
```

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -74,14 +74,14 @@ declare function promptForComponent( component: string, allowLocal: boolean, def
 
 export async function handleCLIException( exception: Error, trackKey?: string, trackBaseInfo?: any = {} ) {
 	const errorPrefix = chalk.red( 'Error:' );
-	if ( exception instanceof UserError && DEV_ENVIRONMENT_NOT_FOUND !== exception.message ) {
-		// User errors are handled in global error handler
-		throw exception;
-	} else if ( DEV_ENVIRONMENT_NOT_FOUND === exception.message ) {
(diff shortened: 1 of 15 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -74,7 +74,7 @@ declare function promptForComponent( component: string, allowLocal: boolean, def
 
 export async function handleCLIException( exception: Error, trackKey?: string, trackBaseInfo?: any = {} ) {
 	const errorPrefix = chalk.red( 'Error:' );
-	if ( exception instanceof UserError ) {
+	if ( exception instanceof UserError && DEV_ENVIRONMENT_NOT_FOUND !== exception.message ) {
```
rebeccahum: Minor: What about just moving up line 80 `} else if ( DEV_ENVIRONMENT_NOT_FOUND === exception.message ) {` to be the first check for readability?

sjinks: Done

rebeccahum (approved):

PR #1237, opened 2023-01-12
Skip requoting if the arg is a valid JSON Object

andrea-sdl:
## Description

This PR improves how CLI handles commands that include JSON objects with spaces.

An example of this is this command
```
vip <environment> -- wp option update test '{"json":"json with spaces"}' --format=json
+ command: wp option update test "{"json":"json with spaces"}" --format=json
Error: Too many positional arguments: with spaces}
```

This PR introduces the same behaviour we have on [the cron runner](https://github.com/Automattic/cron-control-runner/blob/trunk/remote/remote.go#L436) by checking if the argument might be a JSON object, and only if it is a valid JSON object, it skips the quoting.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-wp <environment> -- option update test '{"json":"json with spaces"}' --format=json`
1. No error should be raised

### DIFF
--- a/__tests__/lib/cli/format.js
+++ b/__tests__/lib/cli/format.js
@@ -1,0 +1,34 @@
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import { requoteArgs } from 'lib/cli/format';
+
+describe( 'utils/cli/format', () => {
+	describe( 'requoteArgs', () => {
+		it.each( [
+			{
+				input: [ 'text with spaces' ],
+				expected: [ '"text with spaces"' ],
+			},
+			{
+				input: [ 'textnospaces' ],
+				expected: [ 'textnospaces' ],
+			},
+			{
+				input: [ '{"json":"json with spaces"}' ],
+				expected: [ '{"json":"json with spaces"}' ],
+			},
+			{
+				input: [ '{"json":broken json with spaces}' ],
+				expected: [ '"{"json":broken json with spaces}"' ],
+			},
+		] )( 'should requote args when needed - %o', ( { input, expected } ) => {
+			const result = requoteArgs( input );
+			expect( result ).toStrictEqual( expected );
+		} );
+	} );
+} );
--- a/__tests__/lib/cli/format.js
+++ b/__tests__/lib/cli/format.js
@@ -23,12 +23,12 @@ describe( 'utils/cli/format', () => {
 				expected: [ '{"json":"json with spaces"}' ],
 			},
 			{
-				input: [ '{ "json" : "json with spaces outside strings" }' ],
-				expected: [ '{ "json" : "json with spaces outside strings" }' ],
+				input: [ '{ "json"     :    "json with spaces outside strings"     }' ],
+				expected: [ '{ "json"     :    "json with spaces outside strings"     }' ],
 			},
 			{
-				input: [ '{ "json" : "spaces-outside-strings-only" }' ],
-				expected: [ '{ "json" : "spaces-outside-strings-only" }' ],
+				input: [ '{ "json" : "spaces-outside-strings-only"      }' ],
+				expected: [ '{ "json" : "spaces-outside-strings-only"      }' ],
 			},
 			{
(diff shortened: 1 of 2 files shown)


__tests__/lib/cli/format.js
```diff
@@ -0,0 +1,34 @@
...
+		it.each( [
+			{
+				input: [ 'text with spaces' ],
+				expected: [ '"text with spaces"' ],
+			},
+			{
+				input: [ 'textnospaces' ],
+				expected: [ 'textnospaces' ],
+			},
+			{
```
rinatkhaziev, line 21: Just for the sake of being. thorough we may want to do other spaces variations, like `{ "json"  :  "string"    }`

andrea-sdl: @rinatkhaziev That's some good idea. I've added some extra space variations for JSON based on this comment.  🙇

src/lib/cli/format.js
```diff
@@ -122,14 +122,27 @@ export function requoteArgs( args: Array<string> ): Array<string> {
...
+		if ( arg.includes( ' ' ) && ! isJsonObject( arg ) ) {
 			return `"${ arg }"`;
 		}
 
 		return arg;
 	} );
 }
 
+export function isJsonObject( str: string ): boolean {
+	return typeof str === 'string' && str.startsWith( '{' ) && isJson( str );
```
abdullah-kasim: An edge case but, `JSON.parse( '       {}       ' )` returns `{}`

andrea-sdl: @abdullah-kasim Good catch. I've added some trimming before checking the `startsWith`.  This way if it's a valid JSON Object, `isJson` will return true.

Updated in https://github.com/Automattic/vip-cli/pull/1237/commits/cd4f1161024a3b2098bdcf414634dbc95dac9dc6

src/lib/cli/format.js
```diff
@@ -122,14 +122,27 @@ export function requoteArgs( args: Array<string> ): Array<string> {
...
-		if ( arg.includes( ' ' ) ) {
+		if ( arg.includes( ' ' ) && ! isJsonObject( arg ) ) {
 			return `"${ arg }"`;
 		}
 
 		return arg;
 	} );
 }
 
+export function isJsonObject( str: string ): boolean {
```
maxschmeling, line 133: This is incredibly minor, but I don't love variable / argument names like `str`. I would vote for following the `JSON.parse` naming of `text`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#parameters

But feel free to ignore me. Just personal preference.

src/lib/cli/format.js
```diff
@@ -122,14 +122,27 @@ export function requoteArgs( args: Array<string> ): Array<string> {
...
+	return typeof str === 'string' && str.trim().startsWith( '{' ) && isJson( str );
```
maxschmeling, line 134: Is doing a `typeof str` here necessary / helpful? This isn't an application boundary where we're checking something that can't be statically verified, and we've already done a `.includes` on this value before it is passed to this function.

src/lib/cli/format.js
```diff
@@ -122,14 +122,27 @@ export function requoteArgs( args: Array<string> ): Array<string> {
...
+}
+
+export function isJson( str: string ): boolean {
+	try {
+		JSON.parse( str );
+	} catch ( error ) {
+		return false;
+	}
+	return true;
```
maxschmeling, line 143: We probably handle all of the edge cases by checking if the input text starts with `{`, but [this example from Stack Overflow](https://stackoverflow.com/a/33369954) does an additional check after the `JSON.parse` to make sure the result is an object and not null, because some non-object values (such as `"null"` or a number) will not throw an error.

Probably not necessary because of our input check, but also, could be a slightly more bullet-proof way to do the check.

maxschmeling (approved):
Left a few comments, but nothing that would be a blocker. LGTM 🚢

PR #1243, opened 2023-01-14
fix: Fix peer dependencies and code style issues

sjinks:
## Description

This PR attempts to fix the broken peer dependencies. As a consequence, it has to update a bunch of eslint-related stuff. However, this is not painless: for CI to pass, we have to fix a lot of issues found by ESLint, which contributes to the size of this PR.

Good news is that the most of the changes fix absolute import paths into relative ones and fix misspelled JSDoc tags.

However, there are some changes to the code, they need to be carefully reviewed.

## Steps to Test

* CI should pass;
* Probably it makes sense to run some things manually as well


.eslintrc.json
```diff
@@ -1,10 +1,24 @@
...
+    "plugin:@automattic/wpvip/testing",
+    "plugin:flowtype/recommended"
   ],
   "rules": {
     "no-console": 0,
-    "no-process-exit": 0
+    "no-process-exit": 0,
+    "jsdoc/check-tag-names": [1, {
+      "definedTags": ["flow", "format"]
+    }],
```
sjinks, line 16: We will need to replace all JSDoc comments with these tags with inline comments

.eslintrc.json
```diff
@@ -1,10 +1,24 @@
...
+    "security/detect-object-injection": 0,
+    "security/detect-non-literal-fs-filename": 0,
+    "security/detect-unsafe-regex": 0,
+    "security/detect-non-literal-regexp": 0,
+    "security/detect-non-literal-require": 0
```
sjinks, line 21: These ones generate way too much noise :-( Maybe someday someone has enough courage to try to annotate all those false-positives.

__tests__/bin/vip-config-envvar-delete.js
```diff
@@ -1,16 +1,19 @@
+// @flow
+
 /**
  * External dependencies
  */
+import { describe, expect, it, jest } from '@jest/globals';
```
sjinks: Flow is unaware of Jest; wherever we have to add the `@flow` annotation, we also have to import this file.

package.json
```diff
@@ -55,7 +55,7 @@
     "postinstall": "node ./helpers/check-version.js",
     "build": "babel src -d dist",
     "build:watch": "babel src -d dist --watch --source-maps",
-    "flow": "flow",
+    "flow": "flow || true",
```
sjinks: Flow generates many errors. Will take a look later.

src/bin/vip-import-sql.js
```diff
@@ -266,7 +265,7 @@ If you are confident the file does not contain unsupported statements, you can r
 }
 
 function validateFilename( filename ) {
-	const re = /^[a-z0-9\-\_\.]+$/i;
+	const re = /^[a-z0-9\-_.]+$/i;
```
sjinks, line 268: Unnecessary escapes

src/bin/vip-import-sql.js
```diff
@@ -351,7 +350,7 @@ const displayPlaybook = ( {
 				);
 			}
 			console.log( chalk.yellow( 'The following sites will be affected by the import:' ) );
-			multiSiteBreakdown.map( siteObject => {
+			multiSiteBreakdown.forEach( siteObject => {
```
sjinks, line 353: When the callback passed to `map` does not return anything, `map` should be replaced with `forEach`.

src/bin/vip-logs.js
```diff
@@ -113,7 +118,8 @@ export async function followLogs( opt ): Promise<void> {
 			delay = Math.max( ( logs?.pollingDelaySeconds || DEFAULT_POLLING_DELAY_IN_SECONDS ), MIN_POLLING_DELAY_IN_SECONDS );
 		}
 
-		await new Promise( resolve => setTimeout( resolve, delay * 1000 ) );
+		// eslint-disable-next-line no-await-in-loop
+		await setTimeout( delay * 1000 );
```
sjinks, line 122: `timers/promises` is a nice alternative available since Node 15.

src/lib/analytics/clients/stub.js
```diff
@@ -6,6 +8,6 @@ import type { AnalyticsClient } from './client';
 export default class AnalyticsClientStub implements AnalyticsClient {
 	// eslint-disable-next-line no-unused-vars
 	trackEvent( name: string, props: {} ): Promise<Response> {
-		return new Promise( resolve => resolve( true ) );
+		return Promise.resolve( ( ( true: any ): Response ) );
```
sjinks, line 11: It is not nice to return `true` when a `Response` is exected, but I don't feel like I want to fix broken tests :-(

src/lib/cli/apiConfig.js
```diff
@@ -1,7 +1,5 @@
-/**
- * @flow
- * @format
- */
+// @flow
+// @format
```
sjinks, line 2: Ideally, this is what we have to do to all files abusing `@flow` and `@format` as JSDoc tags.

src/lib/cli/apiConfig.js
```diff
@@ -23,25 +21,24 @@ export async function checkFeatureEnabled(
 	// for now, let's see if the user of the CLI is VIP
 	await trackEvent( 'checkFeatureEnabled_start', { featureName, exitOnFalse } );
 
-	const isVIP = await new Promise( async resolve => {
```
sjinks, line 26: Promise executors should not be async.

src/lib/cli/command.js
```diff
@@ -357,26 +355,28 @@ args.argv = async function( argv, cb ): Promise<any> {
...
+			case 'import-sql': {
 				const site = options.env;
 				if ( site && site.primaryDomain ) {
 					const primaryDomainName = site.primaryDomain.name;
 					info.push( { key: 'Primary Domain Name', value: primaryDomainName } );
 				}
 
 				// Site launched details
-				const haveLaunchedField = site.hasOwnProperty( 'launched' );
+				const haveLaunchedField = Object.prototype.hasOwnProperty.call( site, 'launched' );
```
sjinks, line 366: ESLint [recommends this](https://eslint.org/docs/latest/rules/no-prototype-builtins) as a safe way to call `hasOwnProperty`.

src/lib/cli/format.js
```diff
@@ -5,34 +5,32 @@
...
 
-type Options = {};
-
 export type Tuple = {
 	key: string,
 	value: string,
 };
 
-export function formatData( data: Array<any>, format: string, opts: ?Options ): string {
+export function formatData( data: Array<any>, format: string ): string {
```
sjinks, line 13: `opts` is not used by any of the functions.

src/lib/validations/is-multi-site.js
```diff
@@ -11,15 +11,13 @@ import gql from 'graphql-tag';
...
-import { trackEventWithEnv } from 'lib/tracker';
-import * as exit from 'lib/cli/exit';
+import API from '../../lib/api';
+import { trackEventWithEnv } from '../../lib/tracker';
+import * as exit from '../../lib/cli/exit';
 
 const isMultiSite = new WeakMap();
 
 export async function isMultiSiteInSiteMeta( appId: number, envId: number ): Promise<boolean> {
-	const track = trackEventWithEnv.bind( null, appId, envId );
```
sjinks, line 21: ESLint suggests to lower the scope

src/lib/validations/sql.js
```diff
@@ -361,19 +378,7 @@ const postValidation = async ( options: ValidationOptions ) => {
 		// there was a duplciate table
 		problemsFound++;
 
-		function findDuplicates( arr ) {
```
sjinks, line 364: ESLint does not like nested functions.

src/lib/validations/sql.js
```diff
@@ -323,6 +327,19 @@ const checks: Checks = {
...
 const DEV_ENV_SPECIFIC_CHECKS = [ 'useStatement', 'siteHomeUrlLando' ];
 
+function findDuplicates( arr: Array<*>, where: Set<*> ) {
+	const filtered = arr.filter( item => {
+		if ( where.has( item ) ) {
+			where.delete( item );
+			return false;
+		}
+
+		return true;
```
sjinks, line 337: `filter()` should return a boolean value

src/lib/validations/is-multi-site-sql-dump.js
```diff
@@ -28,23 +28,21 @@ export function sqlDumpLineIsMultiSite( line: string ): boolean {
...
-		await new Promise( resolveBlock => readInterface.on( 'close', resolveBlock ) );
-		resolve( false );
+	readInterface.on( 'error', () => {
+		exit.withError(
+			'An error was encountered while reading your SQL dump file.  Please verify the file contents.'
+		);
 	} );
+	// Block until the processing completes
+	await new Promise( resolveBlock => readInterface.on( 'close', resolveBlock ) );
+	return false;
```
chriszarate: This refactor is broken. How does it return anything other than false?

chriszarate: There are no tests covering it either

chriszarate: Hmmm ... is this function used at all??

chriszarate:
Can we use [`eslint-import-resolver-alias`](https://www.npmjs.com/package/eslint-import-resolver-alias) to keep the absolute imports, at least for this PR? It's kind hard to read esp since those changes aren't in their own commit

chriszarate:
> Can we use [eslint-import-resolver-alias](https://www.npmjs.com/package/eslint-import-resolver-alias) to keep the absolute imports, at least for this PR? It's kind hard to read esp since those changes aren't in their own commit

Nevermind on this, unless you think preserving the absolute imports is important. I reviewed it all together and it wasn't too bad. :)

chriszarate (approved):
Just the one problem with broken / maybe unused(?) function, otherwise looks good

src/lib/validations/is-multi-site-sql-dump.js
sjinks: Yes, you are right - I have completely missed that the promise was resolved from a callback.

Removed the function, as there are no references to it. It looks like the code that used to use it was refactored to use `sqlDumpLineIsMultiSite()` instead.

sjinks:
Thank you!

PR #1247, opened 2023-01-17
Add library to facilitate SQL backup export

aswasif007:
## Description

In this PR, we are adding library to facilitate exporting the latest SQL backup of an application environment. It can be done by attaching the following piece of code to a command:

```javascript
const exportCommand = new SQLExportCommand( app, env, outputPath );
await exportCommand.runSequence();
```
An object of the `SQLExportCommand` class acts as a show-runner for exporting an SQL backup. It runs the following in sequence:
- Initiates export generation for the latest backup
- Generates a pre-signed URL for the exported file
- Downloads the export to the specified path

The implementation also accounts for the following cases:
- There is already an export running for the latest backup
- An export for the latest backup already exists

The implementation has assumed the following limitations, which will be addressed in subsequent work:
- Doesn't check for permission (and rely on the API to fail if the user does not have sufficient permission)
- Doesn't allow user to select network-sites
- Doesn't allow user to make table selection

This class is to be used in the upcoming `vip dev-env sync sql` and `vip export sql` commands. 

## Steps to Test

To manually test this change, we need to add a command first.

1. Add a new file `src/bin/vip-export-sql.js` and paste the following code:
```js
import command from 'lib/cli/command';
import { SQLExportCommand } from 'lib/sql-export';

const examples = [];

const appQuery = `
  id,
  name,
  type,
  organization { id, name },
  environments{
    id
    appId
    type
    name
    primaryDomain { name }
  }
`;

command( {
  appContext: true,
  appQuery,
  envContext: true,
  requiredArgs: 0,
  module: 'export-sql',
} )
  .option(
    'output',
    'Specify the location where you want to save the export file',
  )
  .examples( examples )
  .argv( process.argv, async ( arg: string[], { app, env } ) => {
    const exportCommand = new SQLExportCommand( app, env );
    await exportCommand.runSequence();
  } );

```
2. Add `"vip-export-sql": "dist/bin/vip-export-sql.js",` line to `package.json#bin`
3. Add `.command( 'export-sql', 'Export SQL files ' )` line [here](https://github.com/Automattic/vip-cli/blob/8a068a7560c845ad377bd913d048287cd9533885/src/bin/vip.js#L46)
4. Run `rm -rf dist && npm link node ./dist/bin/vip export-sql @<app>.<env>` (if you are an automattician and using proxy, you may have to add `VIP_PROXY=socks://localhost:8080`)


src/lib/common.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Polls a function until its return value satisfies a condition
+ * @param       {function} fn A function to poll
+ * @param       {int} interval Poll interval in milliseconds
+ * @param       {function} isDone A function that returns true if the polling should stop
+ * @resolves		{Promise} A promise which resolves when the polling is done
+ */
+export async function pollUntil( fn, interval, isDone ) {
+	return new Promise( async ( resolve, reject ) => {
```
abdullah-kasim: You'll want `return await new Promise` here. See [ref under "Code example Anti-Pattern: Returning a promise without tagging the function as async"](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md#code-example-anti-pattern-returning-a-promise-without-tagging-the-function-as-async), pretty surprising but that's how it is.

src/lib/sql-export/export-command.js
```diff
@@ -0,0 +1,329 @@
...
+	 * @param {object} env The environment object
+	 * @param {string} outputFile The output file path
+	 * @constructor
+	 */
+	constructor( app, env, outputFile ) {
+		this.app = app;
+		this.env = env;
+		this.outputFile = outputFile;
+		this.progressTracker = new ProgressTracker( [
+			{ id: 'prepare', name: 'Preparing' },
```
abdullah-kasim: nit: Maybe we should use constants for the id? i.e. `const PROGRESS_PREPARE = 'prepare'` defined somewhere in this file. Helps future readers that aren't familiar with `ProgressTracker` and minimize typo chances.

src/lib/common.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Polls a function until its return value satisfies a condition
+ * @param       {function} fn A function to poll
+ * @param       {int} interval Poll interval in milliseconds
+ * @param       {function} isDone A function that returns true if the polling should stop
```
abdullah-kasim: nit:

```suggestion
 * @param       {function} isDone A function that accepts the return of `fn`. Stops the polling if it returns true
```

abdullah-kasim (approved):
I don't really see any issue, though I'm not too familiar with `vip-cli` myself.

src/lib/common.js
aswasif007: Wow! TIL

@Automattic/vip-platform-forno

src/lib/sql-export/export-command.js
```diff
@@ -0,0 +1,337 @@
+/**
```
saroshaga: No need to name it as `export-command` , I think `sql-export.js` should suffice to keep in line with other modules

src/lib/common.js
```diff
@@ -0,0 +1,23 @@
+/**
```
saroshaga: probably `src/lib/utils.js`

src/lib/sql-export/export-command.js
```diff
@@ -0,0 +1,337 @@
...
+
+	if ( ! success ) {
+		throw new Error();
+	}
+}
+
+/**
+ * Class representing an export command workflow
+ */
+export class ExportCommand {
```
saroshaga:
```suggestion
export class SQlExportCommand {
```
or 

```suggestion
export class DatabaseExportCommand {
```

src/lib/common.js
saroshaga: Would it be cleaner to just not explicitly return a new Promise?

src/lib/sql-export/export-command.js
saroshaga: If we're using `DatabaseExport` then we should reflect that everywhere else

saroshaga (commented):
Looks good!! You ended up building most of the command 😄 

Left a few comments but they are minor

src/lib/common.js
sjinks: See also: https://eslint.org/docs/latest/rules/no-async-promise-executor

I think if you update the branch, ESLint will complain.

aswasif007: > Would it be cleaner to just not explicitly return a new Promise?

That's a very good idea. Wonder how I missed that :smile:

sjinks:
@aswasif007 could you please address these issues, as they break CI: https://github.com/Automattic/vip-cli/actions/runs/3949306457?

Thank you.

aswasif007:
Thanks for flagging @sjinks : https://github.com/Automattic/vip-cli/pull/1251

PR #1249, opened 2023-01-18
refactor(dev-env): Do not use Lando's memcached plugin

sjinks:
Yo dawg, I heard you like reverts…

Reverts Automattic/vip-cli#1248

Now that we have #1250, we can easily upgrade old Landofiles without crashes.


sjinks:
Needs #1250

rinatkhaziev (approved):
Tested, works - as in migration successfully happens.

PR #1250, opened 2023-01-18
feat(dev-env): Implement upgrade path for changes affecting .lando.yml

sjinks:
## Description

Ref: #1245, #1248

This PR implements recovery logic in case Lando has troubles reading its `.lando.yml` file. We simply recreate it using the instance data.

## Steps to Test

Create a dev environment and corrupt its `.lando.yml`, for example, like this:

```diff
 services:
 
   devtools:
-    type: compose
+    type: composer
     services:
```

Don't limit your creativity; the only condition is that Lando should fail when reading the file (you can use `vip dev-env list` to verify that Lando crashes).

The error you see may look like this:
```
WARN ==> composer is not a supported service type. 
TypeError: Cannot read properties of undefined (reading 'builder')
```

Apply the patch. Try `vip dev-env list` again. You should see something like this:
```
WARN ==> composer is not a supported service type. 
There was an error initializing Lando, trying to recover...
Recovery successful, trying to initialize again...
 SLUG           vip-local                                                                           
 LOCATION       /home/volodymyr/.local/share/vip/dev-environment/vip-local                          
 SERVICES       devtools, nginx, php, database, memcached, wordpress, vip-mu-plugins, demo-app-code 
 STATUS         DOWN                                                                                
 DOCUMENTATION  https://docs.wpvip.com/technical-references/vip-local-development-environment/      
```

UPD: another scenario is to remove `.lando.yml` and ensure that the system regenerates it.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -120,6 +120,45 @@ async function getLandoConfig() {
 
 const appMap: Map<string, App> = new Map();
 
+async function regenerateLandofile( instancePath: string ): Promise<void> {
+	const landoFile = path.join( instancePath, '.lando.yml' );
+
+	try {
+		const now = new Date().toISOString().replace( /[^\d]/g, '' ).slice( 0, -3 );
+		const backup = `${ landoFile }.${ now }`;
+		await fs.promises.rename( landoFile, backup );
+		console.warn( chalk.yellow( 'Backed up %s to %s' ), landoFile, backup );
+	} catch ( err ) {
+		// Rename failed - possible the file does not exist. Silently ignoring.
+	}
+
+	const slug = path.basename( instancePath );
+	const currentInstanceData = readEnvironmentData( slug );
+	await updateEnvironment( currentInstanceData );
+}
+
+async function landoRecovery( lando: Lando, instancePath: string, error: Error ): Promise<App> {
+	debug( 'Error initializing Lando app', error );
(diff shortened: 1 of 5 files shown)


src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -124,14 +129,36 @@ async function getLandoApplication( lando: Lando, instancePath: string ): Promis
...
+	let app;
 
-	if ( ! app.initialized ) {
+	try {
+		app = lando.getApp( instancePath );
+		addHooks( app, lando );
 		await app.init();
+	} catch ( error ) {
+		debug( 'Error initializing Lando app', error );
+		console.warn( chalk.yellow( 'There was an error initializing Lando, trying to recover...' ) );
```
rebeccahum: Not sure if we would ever need it, but should we take a back-up if the failed yml file? Especially if it fails regenerating it and for debugging purposes, might be nice to see what the original failed YML file was.

rinatkhaziev: I kinda like the idea, better for debugging too.

sjinks: Done.

rebeccahum (commented):
Nice! 
```
WARN ==> composer is not a supported service type.
There was an error initializing Lando, trying to recover...
Backed up /Users/rebasaurus/.local/share/vip/dev-environment/test-site/.lando.yml to /Users/rebasaurus/.local/share/vip/dev-environment/test-site/.lando.yml.20230120022326
```

rebeccahum (approved):

PR #1253, opened 2023-01-19
Remove use-database statement check

aswasif007:
## Description

Currently, we have check for the `USE DATABASE` statement for `vip dev-env import sql` commands, though we don't have it for `vip import sql` command. In this PR, we are removing that check.

## Steps to Test

1. Check out PR.
2. Run `rm -rf dist && npm link`
3. Run `node ./dist/bin/vip dev-env create` to create a new dev environment with slug `vip-local`. You may have to do `export VIP_PROXY=socks://localhost:8080` if you are an Automattician and using proxy.
4. Run `node ./dist/bin/vip dev-env import sql --slug vip-local __fixtures__/validations/bad-sql-dev-env.sql`
5. The output should not contain the following lines:
```
Error:  SQL Error: USE <DATABASE_NAME> statement on line(s) 4.
Recommendation: Remove these lines
```
6. Also ensure that, `CREATE` and `USE DATABASE` statements have been removed from the file `__fixtures__/validations/bad-sql-dev-env.sql`

### DIFF
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -103,7 +103,7 @@ command( {
 				process.stdin.isTTY = origIsTTY;
 			}
 
-			if ( searchReplace && searchReplace.length && ! inPlace ) {
+			if ( ! inPlace ) {
 				fs.unlinkSync( resolvedPath );
 			}
 
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -10,6 +10,7 @@
  */
 import fs from 'fs';
 import chalk from 'chalk';
+import ChildProcess from 'child_process';
 
 /**
  * Internal dependencies
@@ -63,6 +64,9 @@ command( {
 
 		try {
 			const resolvedPath = await resolveImportPath( slug, fileName, searchReplace, inPlace );
+
+			// Remove the `USE` and `CREATE DATABASE` statements from the SQL file.
+			ChildProcess.execSync( `sed --in-place -E '/^(USE|CREATE DATABASE)/Id' ${ resolvedPath }` );
 
 			if ( ! opt.skipValidate ) {
 				if ( ! isEnvUp( lando, getEnvironmentPath( slug ) ) ) {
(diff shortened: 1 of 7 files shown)


abdullah-kasim (approved):

abdullah-kasim:
Slack discussion: https://a8c.slack.com/archives/CLKJLBFB8/p1674116212483679

Adding these as I move around as I've greatly benefitted from looking through git history for a specific line to see why it was written as such.

Maybe I'll see this again in years to come :)

src/bin/vip-dev-env-import-sql.js
```diff
@@ -64,6 +65,9 @@
 		try {
 			const resolvedPath = await resolveImportPath( slug, fileName, searchReplace, inPlace );
 
+			// Remove the `USE` and `CREATE DATABASE` statements from the SQL file.
+			ChildProcess.execSync( `sed --in-place -E '/^(USE|CREATE DATABASE)/Id' ${ resolvedPath }` );
```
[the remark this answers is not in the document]
aswasif007: Thank you good bot. Addressed: https://github.com/Automattic/vip-cli/pull/1253/commits/04d6bd392ec4882206c75dbaeb6a424f69246947

src/bin/vip-dev-env-import-sql.js
```diff
@@ -64,6 +65,12 @@ command( {
 		try {
 			const resolvedPath = await resolveImportPath( slug, fileName, searchReplace, inPlace );
 
+			// Remove the `USE` and `CREATE DATABASE` statements from the SQL file.
+			ChildProcess.execFileSync(
+				'sed',
```
saroshaga, line 70: Would this work always or would you have to provide the path to the actual executable file? `

aswasif007: `sed` is a standard linux command, so this should work out of the box.

aswasif007: Ah.... this will be executed on user's OS. And `sed` is not available on windows. :grimacing:

saroshaga: yup 😄

aswasif007: On another thought, we don't support vip-cli on windows natively. One has to install windows subsystem for linux to be able to vip-cli on windows: https://docs.wpvip.com/technical-references/vip-cli/installing-vip-cli/

And `sed` is available on WSL.

PR #1254, opened 2023-01-19
refactor(dev-env): Remove healthcheck hook

sjinks:
## Description

This PR‌ removes health checks from the code. These checks belong to `vip-container-images`, and there is not much sense to duplicate them here.

Ref: https://github.com/Automattic/vip-container-images/pull/405

## Steps to Test

Create a dev environment with ES enabled and make sure that ES works right after `vip dev-env start` finishes.


rebeccahum (approved):

rinatkhaziev:
Hmm I think this should be tested well, across all platforms before merging. Not feeling confident enough this won't break stuff.

PR #1258, opened 2023-01-20
fix(dev-env): Get more verbose logs from Lando in debug mode

sjinks:
## Description

The current log level is not very useful when it comes to debugging issues with `docker`, `docker-compose` or Lando internals, because Lando does not display the most useful low level things.

This PR sets the log level to `silly`, which allows us to capture a lot of useful things.

The logic is:
  * `DEBUG=*`: log level is `silly`;
  * ` --debug @automattic/vip:bin:dev-environment`: log level is `debug`;
  * otherwise, log level is `warn`.

## Steps to Test

Run `DEBUG=* vip dev-env list` and observe `SILLY` in the output.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -44,8 +44,8 @@ async function getLandoConfig() {
 
 	debug( `Getting lando config, using paths '${ landoPath }' and '${ atLandoPath }' for plugins` );
 
-	const isLandoDebugSelected = ( process.env.DEBUG || '' ).includes( DEBUG_KEY );
-	const isAllDebugSelected = process.env.DEBUG === '*';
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
Can we make it a flag? I think having `silly` on by default is a great thing from the UX standpoint.

sjinks:
@rinatkhaziev It is behind the `--debug` flag. Or do you mean to introduce another flag?

rinatkhaziev:
IDK maybe an env var, like LANDO_DEBUG_LEVEL=silly. 

My point was mostly around that the current debug log is already verbose and I'm not sure if it's helpful to make it more verbose by default.

sjinks:
@rinatkhaziev
* `DEBUG=*`: silly
* `--debug @automattic/vip:bin:dev-environment`: debug
* otherwise: warn

rinatkhaziev (approved):

PR #1261, opened 2023-01-21
fix(dev-env): Gracefully handle failures to fetch WP versions

sjinks:
## Description

This PR does two things:
1. It adds the logic to handle the case when the list of WP versions is empty (e.g., due to a failure).
2. It makes sure that we don't store non-JSON responses from WP API as a version cache

## Steps to Test

1. Corrupt the `.local/share/vip/wordpress-versions.json` file
2. Try to start an environment

Without this path, it will fail with `Error: Cannot read properties of undefined (reading 'tag')`.

With the patch, it will skip this step:
```
The most recent WordPress version available is: 6.1
Environment WordPress version is: 6.1  ... 😎 nice! 
```

(Assuming that you corrupt the file right before starting the environment, VIP CLI will see the version file as fresh and won't try to refetch it. I have added a check to prevent corrupt data from being cached but, unfortunately, we can be in a situation when we already have a bad file).

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.js
+++ b/src/lib/dev-environment/dev-environment-core.js
@@ -631,7 +631,7 @@ async function updateWordPressImage( slug: string ): Promise<boolean> {
 /**
  * Makes a web call to raw.githubusercontent.com
  */
-export async function fetchVersionList(): Promise<any> {
+export function fetchVersionList(): Promise<any> {
 	const url = `https://${ DEV_ENVIRONMENT_RAW_GITHUB_HOST }${ DEV_ENVIRONMENT_WORDPRESS_VERSIONS_URI }`;
 	return fetch( url ).then( res => res.json() );
 }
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
I think we should use trunk/HEAD in those cases rather than an empty file?

sjinks:
@rinatkhaziev done.

src/lib/dev-environment/dev-environment-core.js
```diff
@@ -662,11 +669,10 @@ export async function getVersionList(): Promise<WordPressTag[]> {
 			await fs.promises.mkdir( cacheFilePath, { recursive: true } );
 		}
 
-		// If the cache doesn't exist, create it
-		// If the cache is expired, refresh it
-		if ( ! fs.existsSync( cacheFile ) || isVersionListExpired( cacheFile, DEV_ENVIRONMENT_WORDPRESS_VERSION_TTL ) ) {
+		// If the cache does not exist or has expired, refresh it
+		if ( await isVersionListExpired( cacheFile, DEV_ENVIRONMENT_WORDPRESS_VERSION_TTL ) ) {
```
sjinks, line 673: We do `stats()` in `isVersionListExpired()`, there's no sense to issue `existsSync()`, as it will be another call to `stats()`.

rinatkhaziev (approved):

PR #1262, opened 2023-01-24
Dev-env: Fix WordPress version inputs for major releases ending in .0

rebeccahum:
## Description
When we pass in major versions ending in `.0`, it doesn't parse it correctly to the matching tag. It also logs it as `✓ Path to your local WordPress:`. This PR fixes that for `--wordpress` and `--php` parameters by adding `processVersionOption()`

Also updates `getTagChoices()` to use the latest WP Versions (if none found) since it was outdated.

## Steps to Test

0) Check out PR
1) `npm run build`
2) `./dist/bin/vip-dev-env-create.js --app-code ~/vipdev/vip-go-skeleton --mu-plugins ~/vipdev/vip-go-mu-plugins --elasticsearch true --mailhog=false --wordpress=6.0 --php=8.0 --xdebug false --phpmyadmin false --multisite false`
3) Observe no `✓ Path to your local WordPress: 6` in the output
4) `vip dev-env start` and ensure that it doesn't throw an error about the image not existing

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -359,8 +359,13 @@ async function processComponent( component: string, preselectedValue: string, de
 	const allowLocal = component !== 'wordpress';
 	const defaultObject = defaultValue ? processComponentOptionInput( defaultValue, allowLocal ) : null;
 	if ( preselectedValue ) {
+		if ( component === 'wordpress' && preselectedValue.indexOf( '.' ) === -1 && ! isNaN( parseFloat( preselectedValue ) ) ) {
+			preselectedValue = parseFloat( preselectedValue ).toFixed( 1 ); // Major releases ending in .0 aren't parsed correctly. This converts it back to a valid version.
+		}
 		result = processComponentOptionInput( preselectedValue, allowLocal );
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -481,8 +483,7 @@ function resolvePhpVersion( version: string ): string {
 	const versions = Object.keys( DEV_ENVIRONMENT_PHP_VERSIONS );
 	const images = ( ( Object.values( DEV_ENVIRONMENT_PHP_VERSIONS ): any[] ): string[] );
 
-	// eslint-disable-next-line eqeqeq -- use loose comparison because commander resolves '8.0' to '8'
-	const index = versions.findIndex( value => value == version );
+	const index = versions.findIndex( value => value === version );
```
rebeccahum, line 491: No need to use loose comparison anymore

sjinks (approved):
LGTM. Bonus points if you write a test for this case :-)

rinatkhaziev:
@rebeccahum so in option definition, why do we default to undefined specifically?

rebeccahum:
@sjinks Challenge accepted.

@rinatkhaziev Defaults added. Mistakenly thought it would prompt if undefined was passed in.

sjinks (approved):

PR #1264, opened 2023-01-27
feature(dev-env): Add dev-env-sync-sql command

aswasif007:
## Description

In this PR, we are adding a new command which synchronizes a local development environment into the latest SQL backup of a production site.

```bash
vip dev-env sync sql --slug=my-site @test-site.develop
```

The above is an example command which does the following:
- Checks if `my-site` exists and the containers are running in user's computer. If not, the command will exit.
- Generate and download the export of the latest backup 
- Extracts site-urls from the downloaded SQL export.
- Runs a search-replace operation to replace the site-urls with the lando domain.
- Imports the curated SQL file to the local database.

The implementation has assumed the following limitations, which will be addressed in subsequent work:

- Doesn't check for permission (and rely on the API to fail if the user does not have sufficient permission)
- Doesn't allow user to select network-sites

## Steps to Test

1. If you are an automattician and using proxy, run `VIP_PROXY=socks://localhost:8080`.
1. You need to have a local development instance of a site running in your computer. Run the following command if you don't have one:
`vip dev-env create --slug=<your-site> @<app>.<env> && vip dev-env start --slug=<your-site>` 
1. Run the following to test this PR:
`rm -rf dist && npm link && node ./dist/bin/vip dev-env sync sql --slug=<your-site> @<app>.<env>`.

### DIFF
--- a/src/bin/vip-dev-env.js
+++ b/src/bin/vip-dev-env.js
@@ -28,5 +28,5 @@ command( {
 	.command( 'import', 'Import data into a local WordPress environment' )
 	.command( 'shell', 'Spawns a shell in a dev environment' )
 	.command( 'logs', 'View logs from a local WordPress environment' )
-	.command( 'sync', 'Sync a local development environment with a production environment' )
+	.command( 'sync', 'Pull data from production to local development environment' )
 	.argv( process.argv );
--- a/src/bin/vip-dev-env.js
+++ b/src/bin/vip-dev-env.js
@@ -28,4 +28,5 @@ command( {
 	.command( 'import', 'Import data into a local WordPress environment' )
 	.command( 'shell', 'Spawns a shell in a dev environment' )
 	.command( 'logs', 'View logs from a local WordPress environment' )
+	.command( 'sync', 'Sync a local development environment with a production environment' )
 	.argv( process.argv );
--- a/__tests__/commands/dev-env-sync-sql.js
+++ b/__tests__/commands/dev-env-sync-sql.js
@@ -85,11 +85,11 @@ describe( 'commands/DevEnvSyncSQLCommand', () => {
 	describe( '.runSearchReplace', () => {
 		it( 'should run search-replace operation on the SQL file', async () => {
 			const cmd = new DevEnvSyncSQLCommand( app, env, 'test-slug' );
-			cmd.siteUrls = [ 'test.go-vip.com' ];
+			cmd.siteUrls = [ '//test.go-vip.com' ];
 			cmd.slug = 'test-slug';
 
 			await cmd.runSearchReplace();
-			expect( replace ).toHaveBeenCalledWith( mockReadStream, [ 'test.go-vip.com', 'test-slug.vipdev.lndo.site' ] );
+			expect( replace ).toHaveBeenCalledWith( mockReadStream, [ '//test.go-vip.com', '//test-slug.vipdev.lndo.site' ] );
 		} );
 	} );
 
--- a/__tests__/commands/dev-env-sync-sql.js
+++ b/__tests__/commands/dev-env-sync-sql.js
@@ -44,6 +44,7 @@ const mockWriteStream = getMockStream( [ { name: 'finish' } ], 20 );
 
 jest.spyOn( fs, 'createReadStream' ).mockReturnValue( mockReadStream );
 jest.spyOn( fs, 'createWriteStream' ).mockReturnValue( mockWriteStream );
+jest.spyOn( fs, 'renameSync' ).mockImplementation( () => {} );
 jest.mock( '@automattic/vip-search-replace', () => {
 	return {
 		replace: jest.fn(),
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -73,6 +73,8 @@
         "vip-dev-env-shell": "dist/bin/vip-dev-env-shell.js",
         "vip-dev-env-start": "dist/bin/vip-dev-env-start.js",
         "vip-dev-env-stop": "dist/bin/vip-dev-env-stop.js",
+        "vip-dev-env-sync": "dist/bin/vip-dev-env-sync.js",
+        "vip-dev-env-sync-sql": "dist/bin/vip-dev-env-sync-sql.js",
         "vip-dev-env-update": "dist/bin/vip-dev-env-update.js",
         "vip-export": "dist/bin/vip-export.js",
         "vip-export-sql": "dist/bin/vip-export-sql.js",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
     "vip-dev-env-logs": "dist/bin/vip-dev-env-logs.js",
     "vip-export": "dist/bin/vip-export.js",
     "vip-export-sql": "dist/bin/vip-export-sql.js",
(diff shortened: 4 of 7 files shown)


src/lib/client-file-uploader.js
```diff
@@ -98,6 +99,20 @@ export const gzipFile = async ( uncompressedFileName: string, compressedFileName
...
+ *
+ * @param {string} inputFilename  The file to unzip
+ * @param {string} outputFilename The file where the unzipped data will be written
+ * @return {Promise} A promise that resolves when the file is unzipped
+ */
+export const unzipFile = async ( inputFilename: string, outputFilename: string ) => {
+	const source = fs.createReadStream( inputFilename );
+	const destination = fs.createWriteStream( outputFilename );
+	const pipe = promisify( pipeline );
+	await pipe( source, createGunzip(), destination );
```
sjinks:
```
import { pipeline } from 'node:stream/promises';
...
await pipeline( source, createGunzip(), destination );
```

No need to `promisify()`.

src/lib/search-and-replace.js
```diff
@@ -58,7 +52,7 @@ export function getReadAndWriteStreams( {
 	let outputFileName;
 
 	if ( inPlace ) {
-		const midputFileName = path.join( makeTempDir(), path.basename( fileName ) );
+		const midputFileName = path.join( makeTempDir( 'vip-search-replace' ), path.basename( fileName ) );
```
sjinks: Could we please clean up the temporary directory when we are done?

aswasif007: We do clean them actually. Just don't have any specific framework in doing so: https://github.com/Automattic/vip-cli/blob/63e125d491f761b6ab8b1f907c7cb548431f81e1/src/lib/dev-environment/dev-environment-cli.js#L753-L755

sjinks: Yes, we clean up the files, but I actually was talking about the temporary directory itself :-) Say, when I run tests locally, I end up with many temporary directories created by search/replace.

I was thinking on smth like this (not tested; probably there is a better alternative):

```js
export function makeTempDir( prefix = 'vip-cli' ) {
    const tempDir = fs.mkdtempSync( path.join( os.tmpdir(), `${ prefix }-` ) );
    debug( `Created a directory to hold temporary files: ${ tempDir }` );

    process.on( 'exit', () => fs.rmSync( tempDir, { recursive: true } );
    return tempDir;
}
```

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -694,3 +695,79 @@ export function getEnvTrackingInfo( slug: string ): any {
 		};
 	}
 }
+
+export class DevEnvSQLImportCommand {
```
sjinks: Would you mind moving this class into a separate file?

sjinks: IDK, maybe create `commands` directory and put it there? Then we can refactor other commands as well, and we will have them in a single directory. What do you think?

src/bin/vip-dev-env.js
```diff
@@ -26,4 +26,5 @@ command( {
 	.command( 'list', 'Provides basic info about all local dev environments' )
 	.command( 'exec', 'Execute an operation on a dev environment' )
 	.command( 'import', 'Import data into a local WordPress environment' )
+	.command( 'sync', 'Sync a local development environment with a production environment' )
```
saroshaga:
```suggestion
	.command( 'sync', 'Pull data from production to local development environment' )
```

Could we align with similar terminology and language as set in the Dashboard?
image

src/lib/dev-environment/dev-environment-cli.js
aswasif007: Sounds like a neat idea :+1:

src/lib/utils.js
```diff
@@ -20,3 +39,24 @@ export async function pollUntil( fn, interval, isDone ) {
...
+ * @return {string}				Path to the temporary directory
+ * @throws {Error} 				If the directory cannot be created
+ */
+export function makeTempDir( prefix = 'vip-cli' ) {
+	const tempDir = fs.mkdtempSync( path.join( os.tmpdir(), `${ prefix }-` ) );
+	debug( `Created a directory to hold temporary files: ${ tempDir }` );
+
+	process.on( 'exit', () => {
+		if ( fs.existsSync( tempDir ) ) {
+			fs.rmSync( tempDir, { recursive: true } );
```
saroshaga:
```suggestion
			fs.rmSync( tempDir, { force: true, recursive: true } );
```
No need for `fs.existsSync()` if we add the force option
https://nodejs.org/api/fs.html#fsrmsyncpath-options

src/lib/utils.js
```diff
@@ -20,3 +39,24 @@ export async function pollUntil( fn, interval, isDone ) {
...
+
+/**
+ * Create a temporary directory in the system's temp directory
+ *
+ * @param {string} prefix Prefix for the directory name
+ * @return {string}				Path to the temporary directory
+ * @throws {Error} 				If the directory cannot be created
+ */
+export function makeTempDir( prefix = 'vip-cli' ) {
+	const tempDir = fs.mkdtempSync( path.join( os.tmpdir(), `${ prefix }-` ) );
```
saroshaga: Just curious, why `mkdtempSync`, if we're cleaning up later anyway?

saroshaga (commented):
Partial review! If we're releasing a "next" version, do we merge this into `develop` or some other branch?

src/lib/utils.js
aswasif007: Just wanted to follow OS's default system for temporary directory handling. It also gives you a unique name for the directory, so you don't have to worry about conflicts.
We didn't strictly need the manual cleanup, but did it anyway because it doesn't hurt to have it.

aswasif007:
> If we're releasing a "next" version, do we merge this into develop or some other branch?

I think merging into `develop` branch will ship this out with other releases. What I am thinking is, we will merge the `DevEnvSyncSQLCommand` class to develop, and then maintain a separate branch for the rest of the code to release a "next" version.

aswasif007:
Heads up: I have moved the library functions of the command to another PR: https://github.com/Automattic/vip-cli/pull/1273 and update the current PR's base branch to the new PR (i.e. https://github.com/Automattic/vip-cli/pull/1273 <- https://github.com/Automattic/vip-cli/pull/1264)

aswasif007:
We have already released this as `2.27.0-dev2` and can be installed with `npm install -g @automattic/vip@next`.

src/lib/dev-environment/dev-environment-cli.js
saroshaga: 👋 I've been thinking about this and I feel like `commands` aren't a true representation of what this code is. This code is helper functions/code wrapped in a class. It is not a command. 

I'm okay (I think) with refactoring into a separate directory but I think we should come up with a better name that commands. People (like me) will confuse it with the actual commands that are in the `bin` folder. 

I would recommend moving these into a subfolder with the `lib` directory (that already contains helper functions/modules) and name it to represent exactly what it is `lib-classes` or `helper-classes`.

rinatkhaziev:
LGTM, @sjinks do you have any feedback on this?

sjinks (approved):

PR #1266, opened 2023-01-27
fix(dev-env): Fix incorrect working directory when vip dev-env exec is run from a 8 level deep directory

sjinks:
## Description

This PR provides the fix for the case when `vip dev-env exec` is run from a 8+ level deep directory.

## Steps to Test

```sh
mkdir -p ~/3/4/5/6/7/8
cd ~/3/4/5/6/7/8
vip dev-env exec --quiet -- wp option get home
```

Will fail without the patch, succeed with the patch.


rinatkhaziev (approved):
Tested, works

PR #1267, opened 2023-01-27
feat(dev-env): Add `vip dev-env shell` command

sjinks:
## Description

This PR adds `vip dev-env shell` command.

Supported options:
  * `--slug name`: environment to operate on
  * `--root`: spawn a root shell (by default, log in as the `www-data` user)

It is also possible to run a custom command in a shell, like this:

```sh
vip dev-env shell -- whoami
```

## Steps to Test

* Create an start an environment, run `vip dev-env shell`, ensure you are in the container.
* Run `vip dev-env shell -- whoami`, ensure the output is `www-data`.
* Run `vip dev-env shell --root -- whoami`, ensure the output is `root`.

E2E tests check last two scenarios; the first one has to be tested manually.


rinatkhaziev:
Thanks! Tested both in interactive and non-interactive modes + root. Works great. Made a couple of changes in examples.

rinatkhaziev (dismissed):

rinatkhaziev:
Got an interesting thing

If you try to call the command for an env that hasn't started up it'll start the container. 

```
vip dev-env shell --slug search
WARN ==> memcached is not a supported service type.
There was an error initializing Lando, trying to recover...
Backed up /Users/rinat/.local/share/vip/dev-environment/search/.lando.yml to /Users/rinat/.local/share/vip/dev-environment/search/.lando.yml.20230201051603
Recovery successful, trying to initialize again...
Starting search_php_1 ... done
```

It's kinda cool but also very confusing and useless. Thoughts?

sjinks:
@rinatkhaziev It's not a bug, it's a feature :-) For me, it was useful when I wanted to test something in the container without spinning up the entire environment.

But, if you think this is confusing, we can check whether the environment is up first.

rinatkhaziev:
> It's not a bug, it's a feature :-)

That's why I said "an interesting thing" :).

But from the customer standpoint:
1) What is the value of spinning up just the PHP container? They wouldn't be able to do anything WP-related?
2) I as a developer would prefer to have my env autostarted for me - maybe with or without a confirmation (E.g. `Environment not started, do you want to start it up?` cc @Crixu for input.

rinatkhaziev (approved):

sjinks:
I think I will rework this a bit: add an option to allow to specify the service to log into. This may be useful for mariadb => mysql transition.

When the environment is down, we log into the database container and dump the database.
Then we kill the database container and the associated volume
Then we update the files to use mysql instead of mariadb
Finally, we log into the new database container and import the database.

sjinks:
@rinatkhaziev could you please take a look again?

src/bin/vip-dev-env-shell.js
```diff
@@ -0,0 +1,101 @@
...
+		debug( 'Args: ', args, 'Options: ', opt );
+
+		const isRoot = !! opt.root;
+		const service = opt.service || 'php';
+		const user = isRoot ? 'root' : userMap[ service ] || 'www-data';
+		const cmd = getCommand( args );
+		try {
+			await landoShell( lando, getEnvironmentPath( slug ), service, user, cmd );
+			await trackEvent( 'dev_env_shell_command_success', trackingInfo );
+		} catch ( error ) {
```
rinatkhaziev, line 93: So this block is not entirely clear to me:

1) Where does error.hide come from?
2) Why do we report an error as success on #L98?

sjinks: 1. From Lando. Lando uses that flag to distinguish between normal errors and non-zero exit code from commands.
2. Because we don't care if a command in the shell failed. We spawned the shell successfully, but what the user runs inside the shell may fail (say, exit 1)

rinatkhaziev: Ok can we add that reasoning to a comment above the block as to not confuse the next person looking at this code?

rinatkhaziev: Updated the comment

PR #1271, opened 2023-02-01
refactor(dev-env): Switch to mysql:8

sjinks:
## Description

Replace mariadb with mysql:8 to match the production environment.

Note: we need a command to regenerate old `.lando.yml` files.

## Steps to Test

E2E tests should pass.


rinatkhaziev:
> Note: we need a command to regenerate old .lando.yml files.

This would be a great case for the migration routines we chatted about in the past, where we'd run some actions on version change

sjinks:
Leaving upgrade for another PR, as it is not trivial: MySQL and MariaDB log files are incompatible, and to swap images is not a viable solution.

The proper upgrade path will require dumping all databases, wiping the old database volume, swapping images, importing the saved dump, probably with a couple of restarts in between.

sjinks:
Docs: we have removed the `--mariadb` option from `vip dev-env create` and `vip dev-env update` commands.

rinatkhaziev (approved):
Tested, works great, including the scenario where Lando ffile is from an older version (needs to be recovered first).

PR #1273, opened 2023-02-03
Add library for the dev-env-sync-sql command

aswasif007:
## Description

The objective of this PR is to split out the PR  https://github.com/Automattic/vip-cli/pull/1264 into two, so that we are able to merge the library functions to `develop` without exposing the command itself to the users.

For further description and testing instructions, please refer to the original PR https://github.com/Automattic/vip-cli/pull/1264

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -52,10 +52,8 @@ export function makeTempDir( prefix = 'vip-cli' ) {
 	debug( `Created a directory to hold temporary files: ${ tempDir }` );
 
 	process.on( 'exit', () => {
-		if ( fs.existsSync( tempDir ) ) {
-			fs.rmSync( tempDir, { recursive: true } );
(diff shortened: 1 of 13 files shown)


src/lib/utils.js
```diff
@@ -1,3 +1,22 @@
...
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+import debugLib from 'debug';
+
+/**
+ * Internal dependencies
+ */
+
+const debug = debugLib( '@automattic/vip:lib:search-and-replace' );
```
saroshaga:
```suggestion
const debug = debugLib( '@automattic/vip:lib:utils' );
```

saroshaga (approved):
One small correction with the debug logger. Otherwise, I think it should be good to get it out in the wild and test it (with feature flags in place of course)

PR #1274, opened 2023-02-03
Run validation on filename & not on full filepath

saroshaga:
## Description

Follow up on #1217 

This is to fix a bug caused in a scenario in which the user provided the relative or absolute path to the file to import instead of just the filename

## Steps to Test

1. Place your file in an invalid directory named `invalid+directory`
1. Check out this PR
1. Run `npm run build`
1. Run `./dist/bin/vip-import-sql.js @your_app.your_env /invalid+directory/<file-path>`
1. This should pass with `Status: Success ✓ imported data should be visible on your site .....`


aswasif007 (approved):
Thanks for working on this Sarosh 💯

PR #1282, opened 2023-02-23
Prefix urls with // to fix the search-replace problem

aswasif007:
## Description

Imagine the following search-replace operations will run as a step of the vip-dev-env-sync command:
```
foo.com -> test.lndo.site
sub.foo.com -> test.lndo.site
```

Because of the ordering, we will end up with a domain `sub.test.lndo.site`.

In this PR, we are aiming to fix that by appending `//` to the domains.
We are also fixing two minor issues:
- Do not replace a non-url entry of home/siteUrl (i.e. if the entry is `inherit`, it shouldn't be replaced)
- Do not display the same occurrence multiple times.

## Steps to Test

1. If you are an automattician and using proxy, run `export VIP_PROXY=socks://localhost:8080`.
2. You need to have a local development instance of a site running in your computer. Run the following command if you don't have one:
`vip dev-env create --slug=srtest @<app>.<env> && vip dev-env start --slug=srtest`
3. Run the following to test this PR:
`rm -rf dist && npm link && node ./dist/bin/vip dev-env sync sql --slug=srtest @<app>.<env>`

### DIFF
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -63,7 +63,7 @@ async function extractSiteUrls( sqlFile ) {
 		} );
 
 		readInterface.on( 'close', () => {
-			resolve( [ ...domains ] );
+			resolve( Array.from( domains ) );
 		} );
 
 		readInterface.on( 'error', reject );
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -146,11 +146,10 @@ export class DevEnvSyncSQLCommand {
 	 */
 	async runImport() {
 		const importOptions = {
-			slug: this.slug,
 			inPlace: true,
 			skipValidate: true,
 		};
-		const importCommand = new DevEnvImportSQLCommand( this.sqlFile, importOptions );
+		const importCommand = new DevEnvImportSQLCommand( this.sqlFile, importOptions, this.slug );
 		await importCommand.run( true );
 	}
 
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -11,6 +11,7 @@
 
 import fs from 'fs';
 import chalk from 'chalk';
+import urlLib from 'url';
 import { replace } from '@automattic/vip-search-replace';
 
 /**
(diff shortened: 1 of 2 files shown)


saroshaga:
> Do not replace a non-url entry of home/siteUrl (i.e. if the entry is inherit, it shouldn't be replaced)

What is a non-url entry? (I didn't quite understand by the example)

saroshaga (approved):
Haven't tested it but the changes make sense.

src/commands/dev-env-sync-sql.js
```diff
@@ -145,11 +146,10 @@ export class DevEnvSyncSQLCommand {
 	 */
 	async runImport() {
 		const importOptions = {
-			slug: this.slug,
 			inPlace: true,
 			skipValidate: true,
 		};
-		const importCommand = new DevEnvImportSQLCommand( this.sqlFile, importOptions );
+		const importCommand = new DevEnvImportSQLCommand( this.sqlFile, importOptions, this.slug );
```
saroshaga, line 152: Why are we making this change?

src/commands/dev-env-sync-sql.js
```diff
@@ -53,16 +54,16 @@ async function extractSiteUrls( sqlFile ) {
...
 			const domain = findSiteHomeUrl( line );
 			if ( domain ) {
-				domains.push( domain );
+				domains.add( '//' + domain );
 			}
 		} );
 
 		readInterface.on( 'close', () => {
-			resolve( domains );
+			resolve( [ ...domains ] );
```
```suggestion
			resolve( Array.from( domains ) );
```

plez

src/commands/dev-env-sync-sql.js
```diff
@@ -35,8 +36,8 @@ function findSiteHomeUrl( sql ) {
 	const results = sql.match( regex );
 
 	if ( results ) {
-		const domain = results[ 2 ].replace( /https?:\/\//, '' );
-		return domain;
+		const url = results[ 2 ];
+		return urlLib.parse( url ).hostname;
```
saroshaga, line 40: How does this change relate ?

aswasif007: I've noticed urls like this: `www.test.com` `www.test.com/` `inherit`

You see, from these urls, our take is only `www.test.com`. This line helps us do just that.

src/commands/dev-env-sync-sql.js
aswasif007: Adjusting this line for a change that was done in another PR: https://github.com/Automattic/vip-cli/commit/20ee77d454f14d7c2a0be474f65c81b9d2326e6b

src/commands/dev-env-sync-sql.js
aswasif007: `inherit` is the "non-url entry" I was talking about.

PR #1285, opened 2023-02-23
refactor(dev-env): Refactor Lando template to avoid running unnecessary scripts

sjinks:
This PR refactors Lando template by:
  * asking Lando not to run its script when they are not needed (results in a faster startup);
  * asking Lando not to mess with permissions for MySQL and Elasticpress;
  * asking Lando to use `exec` instead of `spawn` wherever appropriate.

Steps to test:
1. Apply the PR
2. `npm link`
3. Create a new environment and make sure it works.


rinatkhaziev:
Seems to have worked!

rinatkhaziev (approved):

PR #1287, opened 2023-02-27
fix(dev-env): Pull fresh images for new environments

sjinks:
## Description

  * Fix comparison of `number` with `undefined` (`undefined < Date.now() === false`);
  * Fix Flow typings for `InstanceData`.

## Steps to Test

1. Apply the patch
2. Create a new environment
3. Provided that `ghcr.io` resolves, ensure that Lando will try to pull Docker images.

### DIFF
--- a/src/lib/dev-environment/dev-environment-lando.js
+++ b/src/lib/dev-environment/dev-environment-lando.js
@@ -254,13 +254,14 @@ async function addHooks( app: App, lando: Lando ): Promise<void> {
 			registryResolvable = false;
 		}
 
-		data.opts.pull = registryResolvable && ( instanceData.pullAfter || 0 ) < Date.now();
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -244,7 +244,7 @@ async function addHooks( app: App, lando: Lando ) {
 			registryResolvable = false;
 		}
 
-		data.opts.pull = registryResolvable && instanceData.pullAfter < Date.now();
+		data.opts.pull = registryResolvable && ( instanceData.pullAfter || 0 ) < Date.now();
```
sjinks: * `undefined < Date.now() === false`
  * `undefined > Date.now() === false`
  * `undefined <= Date.now() === false`
  * `undefined >= Date.now() === false`

`undefined` gets coerced to `NaN`, and `NaN` is a special beast because it is not even equal to itself.

rinatkhaziev: thanks!

rinatkhaziev (approved):

PR #1290, opened 2023-03-01
Fail with a msg if an export job with a different backup is running

aswasif007:
## Description

Presently, VIP does not allow running more than one export job for each site. In this PR, we are handling the case where a user initiates the `vip dev-env sync sql` command but discovers that another export job is already running on an older backup. In that scenario, we inform the user that there is an export job already running, and we cannot run another export until that finishes.

## Steps to Test

1. Check out PR.
2. Run `export VIP_PROXY=socks://localhost:8080` if you are using a8c proxy.
3. Go to https://dashboard.wpvip.com/apps/6721/production/data/database/backups and initiate an export job on any backup other than the latest one.
1. Run `rm -rf dist && npm link && node ./dist/bin/vip dev-env sync sql @6721.production --slug=wpvip`
1. Verify that it has failed with this message:
```
Error:  There is an export job already running for this site: https://dashboard.wpvip.com/apps/6721/wasif-test/data/database/backups
Currently, we allow only one export job per site. Please try again later.
```

### DIFF
--- a/src/commands/export-sql.js
+++ b/src/commands/export-sql.js
@@ -186,6 +186,7 @@ export class ExportSQLCommand {
 		DOWNLOAD_LINK: 'downloadLink',
 		DOWNLOAD: 'download',
 	};
+	runningJobs;
 
 	/**
 	 * Creates an instance of SQLExportCommand
@@ -213,6 +214,7 @@ export class ExportSQLCommand {
 	 */
 	async getExportJob() {
 		const { latestBackup, jobs } = await fetchLatestBackupAndJobStatus( this.app.id, this.env.id );
+		this.runningJobs = jobs;
 
 		// Find the job that generates the export for the latest backup
 		return jobs.find( job => {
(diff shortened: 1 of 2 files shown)


src/commands/export-sql.js
```diff
@@ -299,25 +301,31 @@ export class ExportSQLCommand {
 	 */
 	async run() {
 		console.log( `Fetching the latest backup for ${ this.app.name }` );
-		const { latestBackup } = await fetchLatestBackupAndJobStatus( this.app.id, this.env.id );
+		const { latestBackup, jobs } = await fetchLatestBackupAndJobStatus( this.app.id, this.env.id );
```
abdullah-kasim, line 304: We're calling this function again at line 216, through the `getExportJob` call at line 313, and I don't see a good reason to call this API twice.

Maybe we could try refactor this? We could check if `this.runningJobs` is already populated (and if anyone wants to cache invalidate it, they could nullify `this.runningJobs`), or they could put a parameter whether or not to get a fresh or cached value. I can see that we're calling `getExportJob` for `pollUntil` functionality as well.

What do you think?

abdullah-kasim (commented):
I'm also wondering if we should have a timeout for the `pollUntil` function.

saroshaga (approved):

src/commands/export-sql.js
aswasif007: Actually this is intentional. getExportJob gives us the job and the job-status. We are interested in refreshing the job status on every steps.

PR #1291, opened 2023-03-01
Dev-env: Fix silent failing when passing invalid subcommand

rebeccahum:
## Description
Running `vip dev-env nonexistentcmd` just fails silently. This PR fixes it to ensure that the correct error fires.

## Steps to Test

1. Run `npm run build`
1. Run `./dist/bin/vip-dev-env.js nonexistentcmd`
1. Expect to see:
```
Error:  `nonexistentsubcommand` is not a valid subcommand. See `vip --help`
```
1. Run `./dist/bin/vip-dev-env.js nonexistent cmd args`
1. Expect to see same error

### DIFF
--- a/src/bin/vip-dev-env.js
+++ b/src/bin/vip-dev-env.js
@@ -15,7 +15,7 @@
 import command from '../lib/cli/command';
 
 command( {
-	requiredArgs: 1,
+	requiredArgs: 0,
 } )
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
awesome! Maybe this'll enable us to drop `--slug` part too in the future?

sjinks (approved):

PR #1292, opened 2023-03-01
Fix requiredArgs for other vip commands

rebeccahum:
## Description

Similar to #1291, let's fix it for `cache`, `config` and `validate` when invalid subcommands are passed in, so it doesn't fail silently.

## Steps to Test

1. Run `npm run build`
2. `./dist/bin/vip-cache.js nonexistentcmd`
3. `./dist/bin/vip-config.js nonexistent cmd`
4. `./dist/bin/vip-config.js non existent cmd`

### DIFF
--- a/src/bin/vip-cache.js
+++ b/src/bin/vip-cache.js
@@ -15,7 +15,7 @@
 import command from '../lib/cli/command';
 
 command( {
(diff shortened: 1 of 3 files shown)


sjinks (approved):
LGTM

PR #1296, opened 2023-03-06
Add trackers to capture execute, success and error events

aswasif007:
## Description

In this PR, we are adding trackers to capture execution, success and error cases of the `vip dev-env sync sql` command. 

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-cookies.js nom`
1. Verify cookies are delicious.

### DIFF
--- a/src/bin/vip-dev-env-sync.js
+++ b/src/bin/vip-dev-env-sync.js
@@ -13,9 +13,12 @@
  * Internal dependencies
  */
 import command from '../lib/cli/command';
+import { trackEvent } from '../lib/tracker';
 
 command( {
 	requiredArgs: 1,
 } )
 	.command( 'sql', 'Sync local database with a production environment' )
-	.argv( process.argv );
+	.argv( process.argv, async () => {
+		await trackEvent( 'dev_env_sync_command_execute' );
+	} );
--- a/src/bin/vip-dev-env-sync.js
+++ b/src/bin/vip-dev-env-sync.js
@@ -13,12 +13,9 @@
  * Internal dependencies
  */
 import command from '../lib/cli/command';
-import { trackEvent } from '../lib/tracker';
 
 command( {
 	requiredArgs: 1,
 } )
 	.command( 'sql', 'Sync local database with a production environment' )
-	.argv( process.argv, async () => {
-		await trackEvent( 'dev_env_sync_command_execute' );
-	} );
+	.argv( process.argv );
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -174,7 +174,7 @@ export class DevEnvSyncSQLCommand {
 			await unzipFile( this.gzFile, this.sqlFile );
 			console.log( `${ chalk.green( '✓' ) } Extracted to ${ this.sqlFile }` );
 		} catch ( err ) {
-			await this.track( 'error', { errorMessage: 'Could not extract the exported SQL file' } );
+			await this.track( 'archive_extraction_error', { errorMessage: err?.message } );
 			exit.withError( `Error extracting the SQL export: ${ err?.message }` );
 		}
 
@@ -193,7 +193,7 @@ export class DevEnvSyncSQLCommand {
 			await this.runSearchReplace();
 			console.log( `${ chalk.green( '✓' ) } Search-replace operation is complete` );
 		} catch ( err ) {
-			await this.track( 'error', { errorMessage: 'Could not run the search-replace operation' } );
+			await this.track( 'search_replace_error', { errorMessage: err?.message } );
 			exit.withError( `Error replacing domains: ${ err?.message }` );
 		}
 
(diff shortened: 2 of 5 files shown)


src/commands/export-sql.js
```diff
@@ -298,6 +301,7 @@ export class ExportSQLCommand {
 		const { latestBackup } = await fetchLatestBackupAndJobStatus( this.app.id, this.env.id );
 
 		if ( ! latestBackup ) {
+			await this.track( 'error', { errorMessage: 'No backup found for the site' } );
```
saroshaga:
```suggestion
			await this.track( 'no_backup_found_error', { errorMessage: 'No backup found for the site' } );
```

I think we need to descriptive in the event name itself since we'll track all errors separately

src/lib/tracker.js
```diff
@@ -75,3 +75,11 @@ export async function aliasUser( vipUserId ): Promise<Response> {
 export async function trackEventWithEnv( appId, envId, eventName, eventProps = {} ): Promise<Response> {
 	return trackEvent( eventName, { ...eventProps, app_id: appId, env_id: envId } );
 }
+
+export function makeCommandTracker( command, trackingInfo = {} ) {
```
saroshaga, line 79: Have you done this so that you can tag the trackers at different (seemingly unrelated) files with the same command?

src/bin/vip-dev-env-sync.js
```diff
@@ -13,9 +13,12 @@
...
 import command from '../lib/cli/command';
+import { trackEvent } from '../lib/tracker';
 
 command( {
 	requiredArgs: 1,
 } )
 	.command( 'sql', 'Sync local database with a production environment' )
-	.argv( process.argv );
+	.argv( process.argv, async () => {
+		await trackEvent( 'dev_env_sync_command_execute' );
```
saroshaga: If the `dev env sync` command cannot run without the subcommand, i.e. `sql` or something else, then there's no point to tracking this . Let's track at the sub command level instead

saroshaga (changes requested):
I think we'll need to tag each error flow with a different key and not just `error`

src/lib/tracker.js
aswasif007: We need to pass the tracker function to [command classes](https://github.com/Automattic/vip-cli/tree/develop/src/commands) because the same class can be used in multiple commands (for example, [this](https://github.com/Automattic/vip-cli/blob/develop/src/commands/dev-env-import-sql.js) one is used in both dev-env-import-sql and dev-env-sync-sql commands. 

I thought creating a command-tracker will help us keep things streamlined a bit for the future commands.

saroshaga (changes requested):
Apologies for what not making it clearer earlier but we should follow the format already set in Pendo/TRacks for events such as these

https://app.pendo.io/s/5636252415164416/trackevents/Zton7B1h7Y-IqSl7rcrH5t8cDtU
Ex: `vip_cli_dev_command_childcontext_param_error`

src/commands/dev-env-sync-sql.js
```diff
@@ -171,6 +174,7 @@ export class DevEnvSyncSQLCommand {
 			await unzipFile( this.gzFile, this.sqlFile );
 			console.log( `${ chalk.green( '✓' ) } Extracted to ${ this.sqlFile }` );
 		} catch ( err ) {
+			await this.track( 'error:failed_to_extract_gz', { errorMessage: err?.message } );
```
```suggestion
			await this.track( 'archive_extract_error', { errorMessage: err?.message } );
```

src/commands/dev-env-sync-sql.js
```diff
@@ -189,6 +193,7 @@ export class DevEnvSyncSQLCommand {
 			await this.runSearchReplace();
 			console.log( `${ chalk.green( '✓' ) } Search-replace operation is complete` );
 		} catch ( err ) {
+			await this.track( 'error:search_replace_failed', { errorMessage: err?.message } );
```
```suggestion
			await this.track( 'search_replace_error', { errorMessage: err?.message } );
```

src/commands/export-sql.js
```diff
@@ -298,6 +301,7 @@ export class ExportSQLCommand {
 		const { latestBackup } = await fetchLatestBackupAndJobStatus( this.app.id, this.env.id );
 
 		if ( ! latestBackup ) {
+			await this.track( 'error:no_backup_found', { errorMessage: 'No backup found for the site' } );
```
saroshaga:
```suggestion
			await this.track( 'no_backup_found_error', { errorMessage: 'No backup found for the site' } );
```

src/commands/export-sql.js
```diff
@@ -344,6 +349,7 @@ export class ExportSQLCommand {
 		} catch ( err ) {
 			this.progressTracker.stepFailed( this.steps.DOWNLOAD );
 			this.stopProgressTracker();
+			await this.track( 'error:download_failed', { errorMessage: err?.message } );
```
saroshaga:
```suggestion
			await this.track( 'download_failed_error', { errorMessage: err?.message } );
```

saroshaga (commented):
Just suggestions 👇 What do you think?

src/commands/dev-env-sync-sql.js
```diff
@@ -197,6 +202,7 @@ export class DevEnvSyncSQLCommand {
 			await this.runImport();
 			console.log( `${ chalk.green( '✓' ) } SQL file imported` );
 		} catch ( err ) {
+			await this.track( 'sql_import_error', { errorMessage: err?.message } );
```
```suggestion
			await this.track( 'import_error', { errorMessage: err?.message } );
```

otherwise it'll become `dev_env_sync_sql_sql_import_error`

saroshaga (approved):
Left one more correction suggestion but otherwise looks good! Let's be sure to test how these show up on Pendo

src/commands/dev-env-sync-sql.js
aswasif007: Huh, good catch!

PR #1315, opened 2023-03-28
Fix preselected options on update

pschoffer:
## Description

Introduction of config file broke the enforced multisite and title on update.
https://github.com/Automattic/vip-cli/pull/1173

The code used to set `selectedOptions` which were then ignored. This change simplifies the code and makes it that the title and multisite is in fact preselected.

## Steps to Test

1) Create a site as multisite
2) Run update
3) You shoul NOT be prompted for multisite or title


sjinks (approved):
LGTM, but ESLint complains.

PR #1317, opened 2023-04-03
Auto-login support

pschoffer:
## Description

We make the link on the info table for dev-env to show admin link that is autologin. No need to enter username/password.
This changes the Login URL output on new or newly updated environments from:

```                                                 
 LOGIN URL         http://test.vipdev.lndo.site/wp-admin/
```

to:

```                                                                                       
 LOGIN URL         http://test.vipdev.lndo.site/wp-admin/?vip-dev-autologin=1111bbf8-284b-42b1-9a98-aa726d994f1d 
```

Also does all the configuration on update/create to set up the autologin token.
* Generate uuid token on every update/create
* Sets the token to environment json
* in lando file template uses this token and injects it as environment variable
* exposes the token in infor table login link

This is a sibling PR to https://github.com/Automattic/vip-container-images/pull/454

## Steps to Test

1) create/update environment
2) Start it and click the login URL
3) Should be logged in without username/password


sjinks:
I am afraid this did not work for me.

I had an environment created earlier, I ran `vip dev-env update`, `vip dev-env start`.

Got this:
Screenshot_20230418_202538

I followed the login link and…
Screenshot_20230418_202711

All images are of their latest versions:
Screenshot_20230418_202812

sjinks (changes requested):
Please see the comment above. We definitely need a way to make old environments work.

pschoffer:
This is because the volume for `devools` is not overwritten with the new image. So even if the devtools is at latest devtools volume mapped to mu-plugins doesnt contain the latest plugin. I'll try to think about how we can rewrite the volume from within the CLI.

If you do `dev-env delete --soft` it should work.

I don't think this should be a blocker for the feature though, or?

pschoffer:
I've added an auto-removal of dev-tools during startup. Please test @sjinks

sjinks:
I am afraid it still does not work.

Screenshot_20230421_005645

Screenshot_20230421_005748

sjinks:
I have the latest version of ghcr.io/automattic/vip-container-images/dev-tools:0.9.

pschoffer:
Alright! It turns out there are several volumes, one per container. The latest commit iterates over all of them and removes them. Please try again @sjinks

sjinks:
It works now.

sjinks:
@rinatkhaziev could you please take a look as well?

pschoffer:
Note form @rinatkhaziev: Make the redirect to dashboard drop the autologin param.

pschoffer:
I took another look. The dev-tools volume matching actually didn't work if the name had `-` in it. `app.project` is what we want I think.

Also checked the note from @rinatkhaziev about it keeping the autologin param after redirect. This scenario happens when user is already logged in. I created a PR to address that in the container repo where the corresponding code lives
https://github.com/Automattic/vip-container-images/pull/493

rinatkhaziev:
I approved the container PR and tested this one - works as epxected. @sjinks do you have any additional feedback on this or is it good to go?

sjinks:
I think it's good to go. Although I would prefer that we move dev-env-plugin into mu-plugins and eventually bury dev-tools.

sjinks (approved):

PR #1329, opened 2023-04-10
feat(dev-env)!: Replace MailHog with Mailpit

sjinks:
## Description

[Mailhog is not maintained anymore](https://github.com/mailhog/MailHog/issues/442#issuecomment-1493415258). Moreover, it has serious bugs that make it inconvenient to use:

mailhog-1
mailhog-2

[Mailpit](https://github.com/axllent/mailpit) is a faster alternative to MailHog; it is recommended, for example, by [Laravel](https://laravel.com/docs/10.x/sail#previewing-emails).

This PR replaces MailHog with Mailpit, retaining backward compatibility. As a side effect, there is now one more Lando plugin that we can drop: `@lando/mailhog`.

### Backward Compatibility

1. `.lando.yml` is rebuilt automatically (because we removed `@lando/mailhog`).
2. In `instance_data.json`, `mailhog` is treated as a synonym to `mailpit`.
3. `--mailhog` command line switch is aliased to `--mailpit` (and triggers a deprecation warning).

## Steps to Test

E2E tests should pass.

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -9,6 +9,8 @@ proxy:
 <% } %>
   phpmyadmin:
     - <%= siteSlug %>-pma.vipdev.lndo.site
+  mailpit:
+    - <%= siteSlug %>-mailpit.vipdev.lndo.site:8025
 
 services:
   devtools:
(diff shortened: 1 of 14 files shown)


sjinks:
Another improvement is that Mailpit's image is much smaller than MailHog's:

mailhog/mailhog:v1.0.1: 392MB
axllent/mailpit:latest: 24.5MB

sjinks:
~There is one semi-blocking issue: https://github.com/axllent/mailpit/issues/87 (I have a workaround).~

assets/dev-env.lando.template.yml.ejs
```diff
@@ -197,9 +199,19 @@ services:
       clientcode_vipconfig: {}
 <% } %>
 
-<% if ( mailhog ) { %>
-  mailhog:
-    type: mailhog
+<% if ( mailpit ) { %>
+  mailpit:
```
pschoffer, line 203: Nice I like that we have our custom service for this. Less magic is good.

pschoffer (approved):

sjinks:
As per [this comment](https://github.com/axllent/mailpit/issues/87#issuecomment-1502720004), now [using `msmtp` to talk to `mailpit`](https://github.com/Automattic/vip-container-images/pull/464/commits/c634ec36988c7e4ffb40b670803d29f378e7099f). Works flawlessly.

sjinks:
This is how it looks after Automattic/vip-container-images#464 and Automattic/vip-container-images#465:

Screenshot_20230411_144651
Screenshot_20230411_144726

PR #1341, opened 2023-04-28
feat(dev-env): Add a concept of init-only containers

sjinks:
## Description

This PR adds a concept of "init-only" containers: they run some initialization code and the exit. This allows us to save some RAM and CPU cycles.

## Steps to Test

a. The CI should pass.
b. Apply the patch, create a new dev environment from scratch, play with it, stop it, restart it, destroy it - everything should work.

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -17,7 +17,7 @@ services:
     type: compose
     services:
       image: ghcr.io/automattic/vip-container-images/dev-tools:0.9
-      command: sleep infinity
+      command: exit 0
       volumes:
         - devtools:/dev-tools
         - scripts:/scripts
@@ -28,6 +28,7 @@ services:
     volumes:
       devtools: {}
       scripts:
+    initOnly: true
 
   nginx:
     type: compose
@@ -154,6 +155,7 @@ services:
     type: compose
     services:
       image: ghcr.io/automattic/vip-container-images/wordpress:<%= wordpress.tag %>
+#      command: /usr/local/bin/entrypoint.sh
       command: sh -c "rsync -a --chown=www-data:www-data /wp/ /shared/; sleep infinity"
       volumes:
         - ./wordpress:/shared
@@ -162,6 +164,10 @@ services:
           target: /scripts
           volume:
             nocopy: true
+#      environment:
+#        LANDO_NO_SCRIPTS: 1
+#        LANDO_NEEDS_EXEC: 1
+#    initOnly: true
 
 <% if ( muPlugins.mode == 'image' ) { %>
   vip-mu-plugins:
@@ -181,6 +187,7 @@ services:
         LANDO_NEEDS_EXEC: 1
     volumes:
       mu-plugins: {}
+    initOnly: true
 <% } %>
 
 <% if ( appCode.mode == 'image' ) { %>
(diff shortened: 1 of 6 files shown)


assets/dev-env.lando.template.yml.ejs
```diff
@@ -154,6 +156,10 @@ services:
           target: /scripts
           volume:
             nocopy: true
+#      environment:
+#        LANDO_NO_SCRIPTS: 1
+#        LANDO_NEEDS_EXEC: 1
+#    initOnly: true
```
sjinks, line 170: The commented-out lines will need to be uncommented in 2.29.

pschoffer: Is this still the case. Could you add more context to this please?

src/lib/dev-environment/dev-environment-lando.js
```diff
@@ -98,6 +98,25 @@ async function getLandoConfig() {
 
 const appMap: Map< string, App > = new Map();
 
+async function getApplication( lando: Lando, instancePath: string ): Promise< App > {
```
pschoffer: Could we somehow make the difference between getApplication and getLandoApplication more clear? What does the `getLandoApplication` actually do?

pschoffer (changes requested):
This is awesome, sorry it took me that long to get to this @sjinks

sjinks:
Let's postpone this one a bit because I want to finish the conversion of dev-env to TypeScript first, for I am sick and tired of resolving conflicts here and there :-)

assets/dev-env.lando.template.yml.ejs
sjinks: `vip dev-env start` re-pulls images from the repo once a week. I needed to give some time for the required changes to propagate to the existing environments.

src/lib/dev-environment/dev-environment-lando.js
sjinks: I have renamed one of them.

Initially, there used to be only `getLandoApplication()`. It ensured that we initialize every application only once. Then we added a feature to recover broken landofiles. The common part was extracted into `getApplication()` (which is now `initLandoApplication()`).

assets/dev-env.lando.template.yml.ejs
sjinks: I failed to sort out the issue with permissions, will leave this for another PR.

pschoffer:
I get following when starting existing environment:

```
Scanning to determine which services are ready... Please stand by...
Waiting for core files to be copied
Already existing wp-config.php file
Waiting for MySQL to come online
Checking for database connectivity...
Copying dev-env-plugin.php to mu-plugins
Waiting for Elasticsearch to come online........
Checking for WordPress installation...
WordPress already installed
Error: Cannot read properties of null (reading 'length')

Please re-run the command with "--debug @automattic/vip:bin:dev-environment" appended to it and provide the stack trace on the support ticket.

Example:

vip dev-env <command> <arguments> --debug @automattic/vip:bin:dev-environment 

```

I'll try to debug some more later today.

EDIT: This was unrelated bug related to elasticsearch but I added oneliner to fix it here. I hope thats ok @sjinks

pschoffer (approved):

sjinks:
I have applied some TypeScript magic to get rid of ESLint suppression.

PR #1354, opened 2023-05-15
fix(dev-env): Properly handle non-string arguments in resolvePhpVersion()

sjinks:
## Description

Due to the way `args` parses command-line arguments, `resolvePhpVersion()` could get a number as its argument instead of a string. This caused the strict comparison to fail; as a result, PHP 8.2 was chosen in those cases.

You can use

```
vip dev-env update -P 8.1 --debug @automattic/vip:bin:dev-environment
```

to confirm.

This PR ensures that the argument is always a string. It also adds some tests.

## Steps to Test

See above + CI should pass (we have test cases now).

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.js
+++ b/src/lib/dev-environment/dev-environment-cli.js
@@ -494,11 +494,7 @@ export function promptForBoolean( message: string, initial: boolean ): Promise<b
 	return Promise.resolve( initial );
 }
 
-export function resolvePhpVersion( version: string | number ): string {
-	if ( typeof version !== 'string' ) {
-		version = version?.toString() || '';
-	}
-
+export function resolvePhpVersion( version: string ): string {
 	debug( `Resolving PHP version %j`, version );
 
 	if ( version.startsWith( 'image:' ) ) {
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -494,23 +494,28 @@ export function promptForBoolean( message: string, initial: boolean ): Promise<b
 	return Promise.resolve( initial );
 }
 
-function resolvePhpVersion( version: string ): string {
-	debug( `Resolving PHP version '${ version }'` );
+export function resolvePhpVersion( version: string | number ): string {
+	if ( typeof version !== 'string' ) {
+		version = version?.toString() || '';
```
sjinks: Our tests pass `undefined` to the function :-(

src/lib/dev-environment/dev-environment-cli.js
```diff
@@ -494,23 +494,38 @@ export function promptForBoolean( message: string, initial: boolean ): Promise<b
...
+	}
+
+	debug( `Resolving PHP version %j`, version );
 
-	if ( typeof version === 'string' && version.startsWith( 'image:' ) ) {
+	if ( version.startsWith( 'image:' ) ) {
 		return version;
 	}
```
sjinks, line 502: We cannot drop these lines because of backward compatibility. However, I doubt that anyone used this undocumented functionality.

pschoffer:
What do you think about ensuring that verson Options are string always:
```
export function processVersionOption( value: string ): string {
	if ( ! isNaN( value ) && value % 1 === 0 ) {
		// If it's an Integer passed in, let's ensure that it has a decimal in it to match the version tags e.g. 6 => 6.0
		return parseFloat( value ).toFixed( 1 ) + '';
	}

	return value + '';
}
```
(I added the `+ ''` bits there)

From the type definition it looks like that was the intention as well. It solves the `-P 8.1` from my testing and isn't breaking `--wordpress 6.0` or similar

sjinks:
Done.

rinatkhaziev (approved):

PR #1366, opened 2023-05-25
chore(deps): Remove rollbar

sjinks:
## Description

This PR removes Rollbar; as a bonus, it also fixes several Promise-related bugs.

## Steps to Test

CI should pass.


src/bin/vip-config-envvar-get-all.js
```diff
@@ -30,7 +29,7 @@ const examples = [
 	},
 ];
 
-export async function getAllEnvVarsCommand( arg: string[], opt ): void {
+export async function getAllEnvVarsCommand( arg: string[], opt ): Promise<void> {
```
sjinks, line 32: `async` functions always return a Promise.

src/bin/vip.js
```diff
@@ -151,13 +149,12 @@ const rootCmd = async function() {
 		}
 
 		try {
-			Token.set( token.raw );
+			await Token.set( token.raw );
 		} catch ( err ) {
```
sjinks, line 154: If we don't `await` a Promise, no errors it might throw will be caught by `catch`.

The alternative is `Token.set( token.raw ).catch( ( err ) => { /* ... */ } );`

pschoffer (approved):
Nice work. There is conflict in deps now, but this looks great.

PR #1368, opened 2023-05-25
fix(dev-env): Refactor Export SQL command tracks event names

saroshaga:
## Description

Aligning the Export SQL command track events for command errors with the Import SQL command's

Also fixes some formatting issues and incompatible event properties


aswasif007 (approved):
Thanks for fixing these issues @saroshaga :tada:

PR #1370, opened 2023-05-27
fix: Fix issues found by SonarCloud

sjinks:
## Description

This PR fixes some bugs found by SonarCloud.

Ref: https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=Automattic_vip-cli

## Steps to Test

CI should pass.

### DIFF
--- a/src/bin/vip-wp.js
+++ b/src/bin/vip-wp.js
@@ -199,7 +199,7 @@ const bindReconnectEvents = ( { cliCommand, inputToken, subShellRl, commonTracki
 		// Close old streams
 		unpipeStreamsFromProcess( { stdin: currentJob.stdinStream, stdout: currentJob.stdoutStream } );
 
-		trackEvent( 'wpcli_command_reconnect', commonTrackingParams );
+		trackEvent( 'wpcli_command_reconnect', commonTrackingParams ).catch( () => {});
 
 		currentJob = await launchCommandAndGetStreams( {
 			socket: currentJob.socket,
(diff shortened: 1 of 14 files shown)


src/bin/vip-wp.js
```diff
@@ -199,7 +199,7 @@ const bindReconnectEvents = ( { cliCommand, inputToken, subShellRl, commonTracki
 		// Close old streams
 		unpipeStreamsFromProcess( { stdin: currentJob.stdinStream, stdout: currentJob.stdoutStream } );
 
-		trackEvent( 'wpcli_command_reconnect', commonTrackingParams );
+		trackEvent( 'wpcli_command_reconnect', commonTrackingParams ).catch( () => {});
```
pschoffer, line 202: Hmm this is a bit suboptimal. The current implementation of trackEvent captures all the exceptions and log them for us, however if that would ever change with the old code it would possibly be shown somewhere, the new code swallows it.

Maybe we could add one more debug line here? Although my preference would be not to change these lines at all.

That being said, I don't know how better to satisfy the SonarCloud

pschoffer: Maybe `// NOSONAR`?

pschoffer (approved):
Left a small nitpick. Up to you if you want to address it or not.

src/bin/vip-wp.js
sjinks: @pschoffer My concern here is that if trackEvent() fails, this will be an unhandled rejection, and it will terminate the process. Chances are that the user will be unhappy about it because this would kill the otherwise working application.

What do you think?

sjinks: OK, I will merge it now. We will revisit this when converting the codebase to TypeScript. There are much more issues there.

PR #1373, opened 2023-05-29
Reverting accidental push to vip-cli

abdullah-kasim:
Howdy folks. I'll quickly revert this and I'll enable branch protection.


abdullah-kasim:
I've diff'ed with dbd637dcae7336b6bfbb7869e355b80c70e491ee and this is a 1 to 1 copy of that version.

abdullah-kasim:
I'll skip the review, it's exactly the same code and I should learn to trust our tools.

PR #1374, opened 2023-05-29
Add TypeScript support

abdullah-kasim:
## Description

It's about time!

This PR adds TypeScript support to this repo. As this repo uses Flow, we're following back the settings in Parker as that's the only other repo that has both Flow and TypeScript working together to save ourselves a lot of time.

Effort is made to ensure that the compile output is almost exactly the same so that introducing TypeScript is as seamless as possible.

## Steps to Test

1. Check out the `develop` branch.
1. Run `npm install`.
1. Rename the `dist` folder to `dist-develop`.
2. Check out the PR branch
3. Run `npm install`.
4. Rename the `dist` folder to `dist-ts`
5. Using a diffing tool such as `meld`, run `meld dist-ts dist-develop`
6. Only one file `app.js` should be affected, and they should be just whitespace differences.

To test out the type-checking:

1. Check out this PR.
2. Introduce a type error by going to `src/lib/app.ts`, go to line 17, and replace the `return` with `return 123`
3. `npm run check-types` should now fail.
4. Revert your change in step 1.
5. `npm run check-types` should now pass.


sjinks (approved):

chriszarate (approved):
Awesome!

chriszarate:
@sjinks Did we ever solve the issue where `devDependencies` get installed on customer machines (because of `npm-shrinkwrap`)? Just wondering if we'll be making the install payload even heavier

sjinks:
This bug was fixed in recent enough versions of npm.

PR #1376, opened 2023-05-30
refactor: Convert keychain to TypeScript

sjinks:
## Description

Related: #1374

Converts src/lib/keychain and src/lib/token to TypeScript.

## Steps to Test

CI should pass; `npm run type-check` should not find any errors.


src/lib/keychain/browser.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { Keychain } from './keychain';
+
+export default class Secure implements Keychain {
+	getPassword( service: string ): Promise<string | null> {
+		return Promise.resolve( window.localStorage.getItem( service ) );
+	}
+
+	setPassword( service: string, password: string ): Promise<boolean> {
+		try {
+			window.localStorage.setItem( service, password );
```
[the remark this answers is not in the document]
sjinks: I hope we don't use this in a browser context because this solution is far from ideal and is really insecure.

__tests__/lib/keychain.js
```diff
@@ -7,110 +7,57 @@ import { randomBytes } from 'crypto';
...
-
-	it( 'should correctly delete token', () => {
-		return keychain.setPassword( account, password ).then( () => {
-			return keychain.deletePassword( account ).then( async () => {
-				const passwd = keychain.getPassword( account );
-				await expect( passwd ).resolves.toBeNull();
-			} );
-		} );
-	} );
-} );
```
sjinks, line 34: These tests are deceiving: they do not test the Secure keychain.

Jest hoists the `keychain = new Insecure( account );` statement; as a result, `keychain` is `Insecure` in these tests.

I removed these tests because the `token tests (insecure)` test suite checks these assertions.

__tests__/lib/keychain.js
```diff
@@ -7,110 +7,57 @@ import { randomBytes } from 'crypto';
...
-					await expect( passwd2 ).resolves.toBe( 'password2' );
-				} );
-			} );
-		} );
+		const passwd = keychain.getPassword( account );
+		return expect( passwd ).resolves.toBeNull();
 	} );
-} );
 
-describe( 'token tests (browser)', () => {
```
sjinks, line 82: We cannot run in a browser.

src/lib/keychain/browser.js
```diff
@@ -1,30 +0,0 @@
-/* global window */
-// @flow
-
-/**
- * Internal dependencies
- */
-import type { Keychain } from './keychain';
-
-export default class Secure implements Keychain {
```
sjinks, line 9: There is no sense in supporting the browser keychain because we cannot run in a browser.

src/lib/keychain.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+import Insecure from './keychain/insecure';
+import type { Keychain } from './keychain/keychain';
+
+let exportValue: Keychain;
+const debug = debugLib( '@automattic/vip:keychain' );
+
+try {
+	// Try using Secure keychain ("keytar") first
+	const Secure = require( './keychain/secure' ) as Keychain & ( new() => Keychain );
```
pschoffer: This line could use more love :) 

Shall we break it down to multiple lines. `& ( new() => Keychain );` looks pretty alien to me.

src/lib/keychain/insecure.ts
```diff
@@ -0,0 +1,48 @@
...
+		this.file = file;
+
+		this.configstore = new Configstore( this.file );
+	}
+
+	getPassword( service: string ): Promise<string | null> {
+		try {
+			const value: unknown = this.configstore.get( service );
+			// eslint-disable-next-line @typescript-eslint/no-base-to-string
+			return Promise.resolve( null === value || undefined === value ? null : value.toString() ); // NOSONAR
```
pschoffer: Could we keep this one closer to the original please? More readable but less compact.

pschoffer (changes requested):
I left few nitpicks. I like that we are doing this.

pschoffer (changes requested):
I left few nitpicks. I like that we are doing this.

src/lib/keychain.ts
sjinks: Done.

src/lib/keychain/insecure.ts
sjinks: I have changed it, but I don't think it is more readable now.

The problem is that `get()` returns `any`, and we need a `string` or `null`. `undefined` and `null` don't have a `toString()` method, we have to filter them out. For the rest, we call `toString()` to get a string value. We cannot use template literals there because of "Invalid type "{}" of template literal expression."

pschoffer (approved):

PR #1377, opened 2023-05-30
chore: Configure linting for TS files

sjinks:
## Description

This PR properly configures linting for TS files so that we can check both JS, Flow, and TypeScript files.

## Steps to Test

CI should pass :-)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,12 +1,8 @@
 {
-  "parser": "@babel/eslint-parser",
-  "plugins": [
-    "flowtype"
-  ],
   "extends": [
     "plugin:@automattic/wpvip/base",
     "plugin:@automattic/wpvip/testing",
-    "plugin:flowtype/recommended",
+    "plugin:@automattic/wpvip/typescript",
     "plugin:jest/recommended"
   ],
   "rules": {
@@ -18,7 +14,26 @@
     "security/detect-non-literal-regexp": 0,
     "security/detect-non-literal-require": 0
   },
-  "globals": {
-    "jest/globals": true
-  }
+  "overrides": [
+    {
+      "files": ["*.js"],
+      "parser": "@babel/eslint-parser",
+      "plugins": [
+        "flowtype"
+      ],
+      "extends": [
+        "plugin:flowtype/recommended"
+      ],
+      "globals": {
+        "jest/globals": true
+      }
+    },
+    {
+      "files": ["*.ts"],
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "project": ["./tsconfig.json"]
+      }
+    }
+  ]
 }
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -11,6 +11,8 @@
     "checkJs": false,
     // Disallow features that require cross-file information for emit as we're using Babel
     "isolatedModules": true,
+    "strict": true,
```
sjinks, line 14: We'd better do it right from the very beginning :-)

tsconfig.json
```diff
@@ -11,6 +11,8 @@
...
+    "resolveJsonModule": true,
```
sjinks, line 15: Some of the files import `package.json` and various JSON configs. This helps us avoid writing type definitions for them.

rebeccahum (approved):

tsconfig.json
abdullah-kasim: The `"extends": "@tsconfig/node16/tsconfig.json",` above [does this](https://github.com/tsconfig/bases/blob/c02bced38602484550551a56dd7190229079460f/bases/node16.json#L10) for you :)

But I'm okay if you just want to make it clearer.

tsconfig.json
abdullah-kasim: Awesome! Wasn't too sure about adding this option before.

.eslintrc.json
```diff
@@ -18,7 +14,26 @@
...
+      }
+    },
+    {
+      "files": ["*.ts"],
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "project": ["./tsconfig.json"]
+      }
+    }
+  ]
```
abdullah-kasim, line 38: I actually forgot to check the linting in my previous PR :sweat_smile: 

I'm wondering if we actually need the TypeScript part of the overrides. Current version of `@automattic/eslint-plugin-wpvip` [has already done it for us](https://github.com/Automattic/eslint-config-wpvip/blob/84ecb5fa43eece9b11c8ea3be05ac7e74d08b25a/configs/typescript.js#L17).

Maybe also worth updating our current `@automattic/eslint-plugin-wpvip` to the latest so that we don't rely on older version's possible quirks.

abdullah-kasim (approved):

.eslintrc.json
sjinks: > Maybe also worth updating our current @automattic/eslint-plugin-wpvip to the latest 

I believe we have the latest version :-)

> I'm wondering if we actually need the TypeScript part of the overrides.

I couldn't make it work without overrides. In VSCode, the Flow and TS ESLint plugins clashed and generated a lot of noisy messages.

tsconfig.json
sjinks: Oh :-) I totally missed the `extends` clause :-) I only looked at `compilerOptions` 😁

PR #1381, opened 2023-05-31
refactor: Convert lib/cli and dependencies to TypeScript

sjinks:
## Description

Related: #1374

Converts `src/lib/cli` (except `command.js`) and its dependencies (so that the configuration from #1377 does not find any issues) to TypeScript.

## Steps to Test

CI should pass.


__tests__/lib/cli/envAlias.js
```diff
@@ -193,11 +193,6 @@ describe( 'utils/cli/envAlias', () => {
 			const result = envAlias.parseEnvAliasFromArgv( input.argv );
 
 			expect( result ).toEqual( input.expected );
-
-			// If there was an alias, should have parsed it
-			if ( input.argv.includes( '@app.env' ) ) {
-				expect( parseEnvAliasSpy ).toHaveBeenCalledWith( '@app.env' );
-			}
```
sjinks: We cannot test this reliably :-( The old code relied upon `module.exports`, which is gone now. However, tests for `parseEnvAlias` already cover this case.

__tests__/lib/cli/envAlias.js
```diff
@@ -1,14 +1,14 @@
-/* eslint-disable jest/no-conditional-expect */
 /**
- * Internal dependencies
+ * External dependencies
  */
-import envAlias from '../../../src/lib/cli/envAlias';
+import { describe, it, expect } from '@jest/globals';
 
-const parseEnvAliasSpy = jest.spyOn( envAlias, 'parseEnvAlias' );
```
sjinks, line 7: Cannot use spyOn on a primitive value; undefined given

We used to have a weird mixture of ~Bergman and Fellini~ CJS and ESM which was ruled out by Babel. However, those hacks don't work with TypeScript.

abdullah-kasim: I wouldn't say it's Babel's fault, because Jest uses Babel for TypeScript parsing unless we use [ts-jest](https://jestjs.io/docs/getting-started#via-ts-jest). This is more due to ESNext's modules missing references and Jest not wanting to workaround, fix this, and support partial mocking properly.

Currently I think, three choices for fixing this test:

1. Use back the old `module.exports` style.
2. Use classes instead :)
3. Use [this plugin](https://github.com/Xunnamius/babel-plugin-explicit-exports-references) to practically switch back to `module.exports` just for Jest. Just make sure to enable it only for the test environment and Jest.

I'm recommending number 3, and it works really well for https://github.com/Automattic/vip-actions/pull/18 .

abdullah-kasim (commented):
Partially reviewing this for now. **If you don't mind, I've also pushed some commits** for migrating `__tests__/.../envAlias.js` to TypeScript, and added a workaround for the `undefined` issue you're facing with TypeScript exports. Feel free to revert it if you wish - I pushed it to your branch for visibility.

Will review the other parts soon.

__tests__/lib/cli/envAlias.ts
```diff
@@ -1,13 +1,20 @@
...
+/**
+ * External dependencies
+ */
+import { afterEach, describe, expect, it, jest } from '@jest/globals';
+
 /**
  * Internal dependencies
  */
-import envAlias from '../../../src/lib/cli/envAlias';
+import * as envAlias from '../../../src/lib/cli/envAlias';
```
abdullah-kasim, line 10: No default export so this needs to be `* as envAlias`.

__tests__/lib/cli/envAlias.ts
```diff
@@ -1,13 +1,20 @@
...
 
 const parseEnvAliasSpy = jest.spyOn( envAlias, 'parseEnvAlias' );
 
 describe( 'utils/cli/envAlias', () => {
-	afterEach( parseEnvAliasSpy.mockClear );
+	afterEach( () => {
+		parseEnvAliasSpy.mockClear();
+	} );
```
abdullah-kasim, line 17: Jest's types are strictly allowing only void here so we also have to be explicit to return void-ish.

babel.config.js
```diff
@@ -30,6 +30,13 @@ module.exports = {
 			},
 		],
 	],
+	env: {
+		test: {
+			// see https://github.com/facebook/jest/issues/936#issuecomment-821944391
+			// Jest has no good way of partially mocking a module
+			plugins: ['explicit-exports-references']
```
abdullah-kasim, line 37: This is the magic needed to make the undefined issue that we were facing work.

__tests__/lib/cli/envAlias.ts
sjinks: I think we have `esModuleInterop` set to `true`, which implies [`allowSyntheticDefaultImports`](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports).

abdullah-kasim (approved):
Looks good to me, though I don't know the codebase too well.

__tests__/lib/cli/envAlias.ts
abdullah-kasim: That's what I thought so I tried `import envAlias` at first. I got an error instead. I think it only works with the `module.exports` system.

PR #1385, opened 2023-05-31
refactor: Convert lib/api to TypeScript

sjinks:
## Description

Related: #1374

Converts src/lib/api to TypeScript.

## Steps to Test

CI should pass.


src/lib/api/app.ts
```diff
@@ -65,7 +78,7 @@ export default async function( app: string | number, fields: ?any, fragments: ?s
 			},
 		} );
 
-	if ( ! res || ! res.data || ! res.data.app ) {
+	if ( ! res.data.app ) {
```
pschoffer, line 74:
```suggestion
	if ( ! res?.data?.app ) {
```
Do we need to still check individual parts?

pschoffer (changes requested):
One nitpick

src/lib/api/app.ts
sjinks:
```ts
query<T = any, TVariables = OperationVariables>(options: QueryOptions<TVariables, T>): Promise<ApolloQueryResult<T>>;

export declare type ApolloQueryResult<T> = {
    data: T;
    errors?: ReadonlyArray<GraphQLError>;
    error?: ApolloError;
    loading: boolean;
    networkStatus: NetworkStatus;
    partial?: boolean;
};
```

`res` is truthy because `ApolloQueryResult` is truthy;
`res.data` is also truthy because `data` are always there.

pschoffer (approved):

pschoffer (approved):

PR #1388, opened 2023-05-31
Typescript: refactor `vip-whoami`

rebeccahum:
## Description

A baby step to the project 😄 

## Steps to Test

`npm run check-types` should not trigger an error for the affected file


src/bin/vip-whoami.ts
```diff
@@ -13,31 +12,33 @@
...
 
 	await trackEvent( 'whoami_command_execute', trackingParams );
 
-	let currentUser = {};
+	let currentUser: User;
 	try {
 		currentUser = await getCurrentUserInfo();
-	} catch ( err ) {
-		await trackEvent( 'whoami_command_error', { ...trackingParams, error: err.message } );
+	} catch ( err: unknown ) { // eslint-disable-line
```
sjinks:
```suggestion
	} catch ( err: unknown ) {
```

I think it is safe to remove now :-)

src/bin/vip-whoami.ts
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 
 /**
- * @flow
  * @format
  */
```
sjinks: I think we can drop the `@format` pragma as well.

In Prettier, `requirePragma` is `false` [by default](https://prettier.io/docs/en/options.html#require-pragma), and our [.prettierrc](https://github.com/Automattic/vip-cli/blob/4abd8cef3e807ceeb051d1bf49b080f75561c519/.prettierrc) does not speicify it either.

sjinks (approved):
Awesome! 👍🏽

PR #1391, opened 2023-06-01
refactor: Convert lib to TypeScript

sjinks:
## Description

Convert files from `src/lib` to TypeScript.

## Steps to Test

CI should pass.


src/bin/vip-import-sql.js
```diff
@@ -573,7 +573,7 @@ Processing the SQL import for your environment...
 		} catch ( uploadError ) {
 			await track( 'import_sql_command_error', {
 				error_type: 'upload_failed',
-				upload_error: uploadError,
+				upload_error: uploadError.message,
```
sjinks, line 576: We used to throw `string` from the code (OMG). I refactored the code to throw `Error` instead.

src/lib/api/http.ts
```diff
@@ -14,6 +14,10 @@ import { API_HOST } from '../../lib/api';
 
 const debug = debugLib( '@automattic/vip:http' );
 
+type FetchOptions = Omit<RequestInit, 'body'> & {
+	body?: BodyInit | Record<string, unknown>;
+}
```
sjinks, line 19: Sometimes, we pass objects here. They will be encoded into strings, but we need a proper type definition because `RequestInit` does not like `body` as an `object`.

src/lib/search-and-replace.ts
```diff
@@ -23,24 +22,18 @@ import { makeTempDir } from './utils';
 
 const debug = debugLib( '@automattic/vip:lib:search-and-replace' );
 
-const flatten = arr => {
-	return arr.reduce( ( flat, toFlatten ) => {
-		return flat.concat( Array.isArray( toFlatten ) ? flatten( toFlatten ) : toFlatten );
-	}, [] );
-};
```
sjinks, line 30: No longer needed - we can use `.flatMap()` instead.

src/lib/search-and-replace.ts
```diff
@@ -23,24 +22,18 @@ import { makeTempDir } from './utils';
...
+	fileName: string;
+	inPlace: boolean;
+	output: boolean | string | Writable;
+}
 
-export type GetReadAndWriteStreamsOutput = {
-	outputFileName?: string,
-	readStream: stream$Readable | Buffer,
-	usingStdOut: boolean,
-	writeStream: stream$Writable | Buffer,
```
sjinks, line 42: There is no way these can be `Buffer` in the code.

src/lib/search-and-replace.ts
```diff
@@ -130,7 +123,7 @@ export const searchAndReplace = async (
 	const fileSize = getFileSize( fileName );
 
 	// if we don't have any pairs to replace with, return the input file
-	if ( ! pairs || ! pairs.length ) {
+	if ( ! pairs.length ) {
```
sjinks, line 126: `pairs: string | string[]`. Luckily, both `String` and `Array` have `.length` :-)

src/lib/client-file-uploader.ts
```diff
@@ -53,51 +56,45 @@ export interface GetSignedUploadRequestDataArgs {
...
-export const getWorkingTempDir = async () =>
-	new Promise( ( resolve, reject ) => {
-		fs.mkdtemp( path.join( os.tmpdir(), 'vip-client-file-uploader' ), ( err, dir ) => {
-			if ( err ) {
-				return reject( err );
-			}
-			resolve( dir );
-		} );
-	} );
+const getWorkingTempDir = (): Promise<string> => mkdtemp( path.join( os.tmpdir(), 'vip-client-file-uploader' ) );
```
sjinks, line 67: Promise API rules!

src/lib/client-file-uploader.ts
```diff
@@ -290,19 +319,15 @@ export async function uploadUsingMultipart( {
...
 	}
 
-	if (
-		! parsedResponse &&
-		parsedResponse.InitiateMultipartUploadResult &&
-		parsedResponse.InitiateMultipartUploadResult.UploadId
-	) {
-		throw `Unable to get Upload ID from cloud storage. Error: ${ multipartUploadResult }`;
+	if ( ! ('InitiateMultipartUploadResult' in parsedResponse) || ! parsedResponse.InitiateMultipartUploadResult.UploadId ) {
+		throw new Error( `Unable to get Upload ID from cloud storage. Error: ${ multipartUploadResult }` );
```
sjinks, line 330: The original condition was wrong: `parsedResponse` is always truthy.

src/lib/search-and-replace.ts
```diff
@@ -170,34 +162,25 @@ export const searchAndReplace = async (
...
-
-				reject();
-			} );
-	} );
+	try {
+		await pipeline( replacedStream, writeStream );
+	} catch ( error ) {
+		console.log( red( "Oh no! We couldn't write to the output file.  Please check your available disk space and file/folder permissions." ) );
+		throw error;
+	}
```
sjinks, line 174: I love the `pipeline()` function :-) It was introduced in Node 15; we require at least Node 16.

src/commands/dev-env-sync-sql.js
```diff
@@ -174,8 +174,8 @@ export class DevEnvSyncSQLCommand {
 			await unzipFile( this.gzFile, this.sqlFile );
 			console.log( `${ chalk.green( '✓' ) } Extracted to ${ this.sqlFile }` );
 		} catch ( err ) {
-			await this.track( 'archive_extraction_error', { errorMessage: err?.message } );
```
pschoffer, line 177: Why this change? Is there a lint warning or something? We had a bug in upgrade wp site recently where library promised to never return null via types and it did.

pschoffer (changes requested):

src/commands/dev-env-sync-sql.js
sjinks: Because `err` will be an instance of `Error`, it cannot be `null` or `undefined`. See https://github.com/Automattic/vip-cli/pull/1391/files#diff-3e887d690f0e63a71d15440119be0462cb0d8b436c200a7425ea9f891d395dc7R106-R109

If `pipeline()` fails, it will reject with `Error`.

sjinks: > where library promised to never return null via types and it did

It is likely that type definitions for the library were compiled with [`strictNullChecks: false`](https://www.typescriptlang.org/tsconfig#strictNullChecks).

pschoffer (approved):

PR #1394, opened 2023-06-02
FORNO-1609: Combine all error events into one error event

abdullah-kasim:
## Description

This PR combines all error events into one error event.

## Steps to Test

Example:

1. Check out PR.
2. Go to [this line](https://github.com/Automattic/vip-cli/blob/forno-1609%2Ffix%2Fpendo-error-reporting/src/commands/dev-env-sync-sql.js#L183-L183) and add `throw new Error('test!')` above that line
1. Run `npm run build`
2. Run `node dist/bin/vip-dev-env-sync-sql.js --slug=<your-local-single-site-env> @6675.production`
3. Output should look like this:

```shell
Error:  Error extracting site URLs: test!
Debug:  VIP-CLI v2.30.0, Node v16.19.0, linux 5.15.0-73-generic

```

### DIFF
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -166,6 +166,8 @@ export class DevEnvSyncSQLCommand {
 		try {
 			await this.generateExport();
 		} catch ( err ) {
+			// this.generateExport probably catches all exceptions and runs exit.withError() already but just in case, let's log it here
+			await this.track('error', { error_type: 'export_sql_backup', errorMessage: err?.message, stack: err?.stack })
 			exit.withError( `Error exporting SQL backup: ${ err?.message }` );
 		}
 
@@ -174,7 +176,7 @@ export class DevEnvSyncSQLCommand {
 			await unzipFile( this.gzFile, this.sqlFile );
 			console.log( `${ chalk.green( '✓' ) } Extracted to ${ this.sqlFile }` );
 		} catch ( err ) {
-			await this.track( 'error', { error_type: 'archive_extraction', errorMessage: err.message, stack: err?.stack } );
+			await this.track( 'error', { error_type: 'archive_extraction', errorMessage: err?.message, stack: err?.stack } );
 			exit.withError( `Error extracting the SQL export: ${ err.message }` );
 		}
 
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -171,7 +171,7 @@ export class DevEnvSyncSQLCommand {
 			// the following will not get executed if this.generateExport() calls exit.withError() on all exception
 			await this.track( 'error', {
 				error_type: 'export_sql_backup',
-				errorMessage: err?.message,
+				error_message: err?.message,
 				stack: err?.stack
 			} );
 			exit.withError( `Error exporting SQL backup: ${ err?.message }` );
@@ -182,7 +182,11 @@ export class DevEnvSyncSQLCommand {
 			await unzipFile( this.gzFile, this.sqlFile );
 			console.log( `${ chalk.green( '✓' ) } Extracted to ${ this.sqlFile }` );
 		} catch ( err ) {
-			await this.track( 'error', { error_type: 'archive_extraction', errorMessage: err?.message, stack: err?.stack } );
+			await this.track( 'error', {
+				error_type: 'archive_extraction',
+				error_message: err?.message,
+				stack: err?.stack
+			} );
 			exit.withError( `Error extracting the SQL export: ${ err.message }` );
 		}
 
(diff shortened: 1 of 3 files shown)


aswasif007 (changes requested):
Please follow the approach taken here: https://github.com/Automattic/vip-cli/pull/1368

abdullah-kasim:
I need more context @aswasif007 . This is the approach that all the other CLI command uses, so I need to understand why the approach would be different here. Plus, the approach taken in the PR linked doesn't seem to have a general failure mode.

aswasif007:
Apologies for the rather vague comment.

The reason your solution won't work is because in the command workflow, [we do `exit.withError` when we catch an exception.](https://github.com/Automattic/vip-cli/blob/develop/src/commands/dev-env-sync-sql.js#L178) So, the errors won't bubble up to be caught by the `try-catch` block around the `async cmd.run()`. Instead, the process will exit.

Another issue is, we are already calling `track` in the command flow: `await this.track( 'archive_extraction_error',...)`, and we don't want to send two pings to pendo for one incident. Instead, we can update the said track call as: `await this.track( 'error', { error_type: 'archive_extraction_failure', ... } )`

aswasif007 (approved):

src/commands/dev-env-sync-sql.js
```diff
@@ -166,6 +166,8 @@ export class DevEnvSyncSQLCommand {
 		try {
 			await this.generateExport();
 		} catch ( err ) {
+			// this.generateExport probably catches all exceptions and runs exit.withError() already but just in case, let's log it here
+			await this.track('error', { error_type: 'export_sql_backup', errorMessage: err?.message, stack: err?.stack })
```
```suggestion
			await this.track( 'error', { error_type: 'export_sql_backup', errorMessage: err?.message, stack: err?.stack } )
```

src/commands/dev-env-sync-sql.js
```diff
@@ -166,6 +166,8 @@ export class DevEnvSyncSQLCommand {
 		try {
 			await this.generateExport();
 		} catch ( err ) {
+			// this.generateExport probably catches all exceptions and runs exit.withError() already but just in case, let's log it here
```
saroshaga: Is there a chance that we'd be duplicating the event? That would create bad data?

src/commands/dev-env-sync-sql.js
```diff
@@ -202,7 +209,7 @@ export class DevEnvSyncSQLCommand {
 			await this.runImport();
 			console.log( `${ chalk.green( '✓' ) } SQL file imported` );
 		} catch ( err ) {
-			await this.track( 'import_error', { errorMessage: err?.message } );
+			await this.track( 'error', { error_type: 'import', errorMessage: err?.message, stack: err?.stack } );
```
```suggestion
			await this.track( 'error', { error_type: 'import_sql_file', errorMessage: err?.message, stack: err?.stack } );
```

or something...

src/commands/dev-env-sync-sql.js
abdullah-kasim: There's a small chance because we're relying on developer discipline to `try catch` at each place and making sure it exits rather than, say, catching everything in one place and then exit.

From reading the code, the developer discipline that we have is good right now so this is safe.

More than that, we'd actually want it to trigger twice, because that'd mean that something went really wrong and someone made it trigger twice i.e. by throwing twice, or if `exit.withError` failed badly, or if `trackEvent` somehow threw an exception too! Then we could figure out why it happened, and fix it.

In all normal cases, this line shouldn't get executed because again, `this.generateExport()` probably catches and exit.withError() properly, already.

src/commands/dev-env-sync-sql.js
abdullah-kasim: Is ESlint broken in this project?

src/commands/dev-env-sync-sql.js
saroshaga: I meant that we should not trigger the `export_sql_backup` event itself twice in the same flow . And I don't think we're doing that so we should be good.

When we trigger it twice (or duplicate it essentially) we create bad data for analysers who look at the number of errors we received for a tool and gauge whether the tool is performing well or not.

src/commands/dev-env-sync-sql.js
```diff
@@ -174,14 +176,19 @@ export class DevEnvSyncSQLCommand {
...
 			exit.withError( `Error extracting the SQL export: ${ err.message }` );
 		}
 
 		try {
 			console.log( 'Extracting site urls from the SQL file...' );
 			this.siteUrls = await extractSiteUrls( this.sqlFile );
 		} catch ( err ) {
+			await this.track( 'error', {
+				error_type: 'extract_site_urls',
+				errorMessage: err?.message,
```
```suggestion
				error_message: err?.message,
```

Tracks will reject this (although Pendo will accept it)  because it was in camel case . Tracks only accepts snake_case

saroshaga (changes requested):

src/commands/dev-env-sync-sql.js
saroshaga: This will need to be changed everywhere in this file

abdullah-kasim: TIL

src/commands/dev-env-sync-sql.js
abdullah-kasim: Oh, all right. Yes, we're not triggering that particular event twice.

saroshaga (approved):

PR #1395, opened 2023-06-02
Add types for the other ways of using enquirer

abdullah-kasim:
## Description

Add types for enquirer. I copy-pasted this off https://github.com/enquirer/enquirer/pull/307 and sprinkled some salt in it so that `prompt<T>` works properly.

I agree with the previous PR, just move to `inquirer` is the better way. Apparently we were on `inquirer` but moved to `enquirer` to have working autocomplete. Which is weird - `inquirer` already has autocomplete through a plugin

## Steps to Test

1. Ensure that tests pass. This is just a type change.


sjinks (approved):
🔥🔥🔥

abdullah-kasim:
I need to fix something.

PR #1397, opened 2023-06-02
feature(dev-env): Add ability to sync multisites

aswasif007:
## Description

Presently, the command `vip dev-env sync sql` works only for single sites. For the multisites, it shows the user a message to basically divert them towards using `vip dev-env export sql` and `vip dev-env import sql` commands.

In this PR, we are adding multisite support for the `vip dev-env sync sql` command to seamlessly sync database backups to the dev-env multisites. The search-replace operation will take happen as follows:

```
root-site -> [slug].[lando-domain]
subsite-1 -> [slug].[lando-domain]
subsite-2 -> 2.[slug].[lando-domain]
subsite-3 -> 3.[slug].[lando-domain]
.
.
.
```

## Steps to Test

1. Pull the PR.
2. Create a dev-env environment for a multisite: `vip dev-env create @0000.production --slug testsite`. When creating the environment, do not use the skeleton repository. Instead, clone the actual repository from `wpcomvip`.
3. Build VIP-CLI: `rm -rf dist && npm link`
4. Run the sync command: `./dist/bin/vip dev-env sync sql @0000.production --slug testsite`

### DIFF
--- a/src/bin/vip-dev-env-sync-sql.js
+++ b/src/bin/vip-dev-env-sync-sql.js
@@ -40,8 +40,9 @@ const appQuery = `
 		primaryDomain { name }
 		uniqueLabel
 		isMultisite
-		wpSites {
+		wpSitesSDS {
 			nodes {
+				id
 				blogId
 				homeUrl
 			}
--- a/src/bin/vip-dev-env-sync-sql.js
+++ b/src/bin/vip-dev-env-sync-sql.js
@@ -60,7 +60,7 @@ command( {
 	.option( 'slug', 'Custom name of the dev environment' )
 	.examples( examples )
 	.argv( process.argv, async ( arg: string[], { app, env, slug } ) => {
-		const trackerFn = makeCommandTracker( 'dev_env_sync_sql', { app: app.id, env: env.uniqueLabel, slug } );
+		const trackerFn = makeCommandTracker( 'dev_env_sync_sql', { app: app.id, env: env.uniqueLabel, slug, multisite: env.isMultisite } );
 		await trackerFn( 'execute' );
 
 		const lando = await bootstrapLando();
--- a/src/bin/vip-dev-env-sync-sql.js
+++ b/src/bin/vip-dev-env-sync-sql.js
@@ -40,7 +40,7 @@ const appQuery = `
 		primaryDomain { name }
 		uniqueLabel
 		isMultisite
-		wpSitesSDS {
+		wpSitesSDS(first:500) {
 			nodes {
 				id
 				blogId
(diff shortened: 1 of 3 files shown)


src/bin/vip-dev-env-sync-sql.js
```diff
@@ -41,6 +40,12 @@ const appQuery = `
 		primaryDomain { name }
 		uniqueLabel
 		isMultisite
+		wpSites {
```
aswasif007: This has pagination. So by default it won't pull all the sites.

Todo: find a way around.

aswasif007: Also:
```suggestion
		wpSitesSDS {
```
Because `wpSites` will be deprecated.

src/commands/dev-env-sync-sql.js
```diff
@@ -58,12 +53,13 @@ async function extractSiteUrls( sqlFile ) {
...
 			const domain = findSiteHomeUrl( line );
 			if ( domain ) {
-				domains.add( '//' + domain );
+				domains.add( domain );
 			}
 		} );
 
 		readInterface.on( 'close', () => {
-			resolve( Array.from( domains ) );
+			// Soring by length so that longest domains are replaced first
```
aswasif007, line 61: We want the longest domains to be replaced first to handle issues like this:

Say, we do the following:
```
wpvip.com -> foo.com
static.wpvip.com -> bar.com
```

If we replace `wpvip.com` before the other one, we end up getting an unintended `static.foo.com`, and the `static.wpvip.com -> bar.com` never happens.

src/bin/vip-dev-env-sync-sql.js
saroshaga: yes

src/commands/dev-env-sync-sql.js
```diff
@@ -141,6 +138,26 @@ export class DevEnvSyncSQLCommand {
...
+		}
+
+		const networkSites = this.env.wpSitesSDS.nodes
+		if ( ! networkSites ) return;
+
+		for ( const site of networkSites ) {
+			if ( ! site.blogId || site.blogId === 1 ) continue;
+
+			const url = site.homeUrl.replace( /https?:\/\//, '' );
+			if ( ! this.searchReplaceMap[ url ] ) return;
```
abdullah-kasim: nit: I'm wondering why we're calling a `return` here. Should we have a small comment here to explain it?

abdullah-kasim (approved):
Looks good to me. I have a small nit where I couldn't understand what's going on.

src/commands/dev-env-sync-sql.js
aswasif007: Ahh this actually is a mistake. It should be `continue` 
```suggestion
			if ( ! this.searchReplaceMap[ url ] ) continue;
```

PR #1400, opened 2023-06-09
FORNO-1512: Add prompt for selecting a backup

abdullah-kasim:
## Description

This PR prompts the user to select a backup for them to `sync` (will work for `export` soon!)

Also adds the skeleton for a future feature

## TODO

- [ ] Unit tests for functions that are in use.
- [ ] Nice to have: A separate PR for fixing all the typescript issues (this PR solves some but not all)

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Run `node dist/bin/vip-dev-env-sync-sql.js --slug=0000-production @0000-production`
1. You should get a prompt to import some backups and not just the latest :)
1. Everything should just work.


src/lib/backup-copies/types.d.ts
```diff
@@ -0,0 +1,61 @@
...
+import { GetAppBackupsV2Query } from './backup-copy-manager.generated';
+import { BackupConfiguration, BackupLabel } from './backup-copy-manager';
+
+declare const appType: GetAppBackupsV2Query['app'];
+const nodeType = appType?.environments?.[ 0 ]?.backups?.nodes?.[ 0 ];
+
+export type RemoteBackupRaw = Exclude<typeof nodeType, null | undefined>
+
+export interface RemoteBackupDetails {
+	processed: RemoteBackup,
```
abdullah-kasim, line 12: I'm going to rename this to `transformed`. I originally don't want to provide `raw` but I'm trying to transplant this to existing code so having `raw` will simplify things.

src/lib/cli/format.ts
```diff
@@ -181,7 +181,16 @@ export class RunningSprite {
 	}
 }
 
-export function getGlyphForStatus( status: string, runningSprite: RunningSprite ): string {
+type Status =
```
abdullah-kasim, line 184: TypeScript starting to error out on my end so here's something to fix it quickly (or for good!)

types/enquirer/index.d.ts
```diff
@@ -326,7 +326,8 @@ declare namespace Enquirer {
 
 		export type SelectQuestionOptions = {
 			type?: 'select';
-		} & SelectQuestionOptionsBase;
+			result?: ( this: Select, value: string ) => any;
+		} & Omit<SelectQuestionOptionsBase, 'result'>;
```
abdullah-kasim, line 330: The types we copy-pasted has some issue where `this.focused` is missing - so this ensures that it's added properly.

src/lib/backup-copies/backup-copy-manager.ts
```diff
@@ -0,0 +1,615 @@
+import xdgBasedir from 'xdg-basedir';
```
abdullah-kasim, line 1: I'll rearrange the imports. Don't worry, I didn't forget.

It's just that I usually do this last because auto-import speeds things up so fast.

PR #1414, opened 2023-06-21
[dev-env] Tweak error handling dev-env sync

pschoffer:
## Description

* Unifies boolean parsing logic in exec
* Adds `--force` to dev-env sync sql to avoid check if environment is up or not
* Adds check for slug presence to replace "Contact customer support" - type of error into "Use '--slug' ..." - type

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-dev-env-sync-sql.js` 
1. Should instruct you to provide env name.


src/bin/vip-dev-env-sync-sql.js
```diff
@@ -58,15 +62,18 @@ command( {
 	module: 'dev-env-sync-sql',
 } )
 	.option( 'slug', 'Custom name of the dev environment' )
+	.option( 'force', 'Disabling validations before task execution', undefined, processBooleanOption )
```
rinatkhaziev:
```suggestion
	.option( 'force', 'Disable validations before task execution', undefined, processBooleanOption )
```

rinatkhaziev (changes requested):
I think using `present simple` would be more appropriate that `ing`,`Suppress output` and `Disable validation before task execution` sound more "natural" in the context of CLI commands.

pschoffer:
Thanks @rinatkhaziev

sjinks (approved):

PR #1421, opened 2023-06-27
refactor: Convert dev-environment to TypeScript

sjinks:
## Description

Work in progress.

## Steps to Test

CI should pass.

### DIFF
--- a/__tests__/lib/dev-environment/dev-environment-cli.js
+++ b/__tests__/lib/dev-environment/dev-environment-cli.js
@@ -22,6 +22,7 @@ import {
 	setIsTTY,
 	processVersionOption,
 	resolvePhpVersion,
+	promptForWordPress,
 } from '../../../src/lib/dev-environment/dev-environment-cli';
 import * as devEnvCore from '../../../src/lib/dev-environment/dev-environment-core';
 import * as devEnvConfiguration from '../../../src/lib/dev-environment/dev-environment-configuration-file';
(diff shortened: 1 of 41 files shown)


__tests__/lib/dev-environment/dev-environment-cli.js
```diff
@@ -22,6 +22,7 @@ import {
 	setIsTTY,
```
abdullah-kasim, line 22: nit: You could migrate the tests to TypeScript too if you want.

src/lib/dev-environment/dev-environment-configuration-file.ts
```diff
@@ -26,45 +21,44 @@ export async function getConfigurationFileOptions(): Promise< ConfigurationFileO
...
-	);
-
-	debug( 'Sanitized configuration from file:', configuration );
-	return configuration;
+	try {
+		const configuration = sanitizeConfiguration( configurationFromFile );
+		debug( 'Sanitized configuration from file:', configuration );
+		return configuration;
+	} catch ( err ) {
+		exit.withError( err instanceof Error ? err : new Error( 'Unknown error' ) );
```
abdullah-kasim, line 55: nit: Since we're already making changes to the implementation here, best to log what is inside `err` at this point of time if it's not an `Error` instance?

Feels like it's worth the effort to me.

abdullah-kasim (approved):
Having changes to the implementation scares me a little but I guess you've added tests with that in mind.

Looks good to me (admittedly I mostly glance through it).

src/lib/dev-environment/dev-environment-configuration-file.ts
sjinks: We actually do not expect anything but `Error` here. `err` is `unknown` in TS, and this entire ternary operator is just to keep TS happy.

PR #1425, opened 2023-06-27
Add Slowlogs command

luiztiago:
## Description

Customers could use the CLI to list Slow Query Logs as a similar feature used on the VIP Dashboard.

## Steps to Test

1. Check out PR.
1. Run `npm run build` and `npm link`
1. Run `node ./dist/bin/vip @SITE_ID.production slowlogs`
1. Verify list of slowlogs or "No logs found" if don't have any slowlogs to be listed


src/bin/vip-slowlogs.js
```diff
@@ -0,0 +1,218 @@
...
+		org_id: opt.app.organization.id,
+		app_id: opt.app.id,
+		env_id: opt.env.id,
+		limit: opt.limit,
+		follow: opt.follow || false,
+		format: opt.format,
+	};
+}
+
+function printSlowlogs( slowlogs, format ) {
```
pschoffer: There is a build-in formatting feature to vip-cli already. Could we use that instead to stay consistent? Or is your format too different?

Example of usage:
https://github.com/Automattic/vip-cli/blob/update/sync_error_handling/src/bin/vip-config-envvar-get-all.js#L77

pschoffer (dismissed):
This looks great! The only thing that comes to mind is resuing existing formatting logic

src/bin/vip-slowlogs.js
luiztiago: @pschoffer Good catch, I'll test it with the build-in format. It should works well :)

luiztiago: @pschoffer done, thanks :)

src/bin/vip-slowlogs.js
```diff
@@ -0,0 +1,219 @@
...
+	// Strip out __typename
+	slowlogs = slowlogs.map( log => {
+		const { timestamp, rowsSent, rowsExamined, queryTime, requestUri, query } = log;
+
+		return { timestamp, rowsSent, rowsExamined, queryTime, requestUri, query };
+	} );
+
+	// let output = '';
```
rinatkhaziev: Should this block be removed?

src/bin/vip-slowlogs.js
rinatkhaziev, line 1: Probably should implement in TS from the start? as there's ongoing work to convert everything to TS and it's likely we'll migrate to something like oclif in the near future.

rinatkhaziev (changes requested):
Thanks! left a bit of inline feedback

luiztiago:
@rinatkhaziev Thanks, updates are done. Good catches! :)

rinatkhaziev (approved):

PR #1428, opened 2023-07-05
refactor: convert `validations` to TS

sjinks:
## Description

This PR converts all files from `src/validations` to TypeScript.

## Steps to Test

CI should pass.

### DIFF
--- a/src/lib/validations/is-multi-site.ts
+++ b/src/lib/validations/is-multi-site.ts
@@ -58,7 +58,7 @@ export async function isMultiSiteInSiteMeta( appId: number, envId: number ): Pro
 			gql_err: GraphQlError,
 		} );
 		// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-		exit.withError( `StartImport call failed: ${ GraphQlError as Error }` );
+		exit.withError( `StartImport call failed: ${ GraphQlError }` );
 	}
 
 	if ( Array.isArray( res.data.app?.environments ) ) {
--- a/src/lib/validations/is-multi-site.ts
+++ b/src/lib/validations/is-multi-site.ts
@@ -1,9 +1,4 @@
 /**
- * @flow
- * @format
- */
-
-/**
  * External dependencies
  */
 import gql from 'graphql-tag';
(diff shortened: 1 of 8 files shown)


src/lib/validations/is-multi-site.ts
```diff
@@ -55,19 +57,20 @@ export async function isMultiSiteInSiteMeta( appId: number, envId: number ): Pro
...
+	if ( Array.isArray( res.data.app?.environments ) ) {
+		const environments = res.data.app!.environments;
 		if ( ! environments.length ) {
-			isMultiSite.set( arguments, false );
+			isMultiSite.set( args, false );
 			return false;
 		}
 		// we asked for one result with one appId and one envId, so...
-		const thisEnv = environments[ 0 ];
-		if ( thisEnv.isMultiSite || thisEnv.isSubdirectoryMultisite ) {
```
sjinks, line 69: This one confuses me a bit. According to the schema, the field is named `isMultisite`, not `isMultiSite`.

src/lib/validations/is-multisite-domain-mapped.ts
```diff
@@ -115,18 +114,18 @@ export async function isMultisitePrimaryDomainMapped(
...
+		// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
 		exit.withError( `StartImport call failed: ${ GraphQlError }` );
 	}
-	if ( ! Array.isArray( res?.data?.app?.environments ) ) {
+	if ( ! Array.isArray( res.data.app?.environments ) ) {
 		return false;
 	}
 
-	const environments = res.data.app.environments;
+	const environments = res.data.app!.environments;
```
abdullah-kasim, line 124: nit: I'd usually do `res.data.app?.environments || []` rather than `!` to fix the type error here, though I guess this is more of a parker bug - because the `app` query never returns null and should've been of type `App!` in the schema.

I'm just paranoid 😁 but I guess I'd rather have zero chance of the undefined access from happening?

src/lib/validations/is-multisite-domain-mapped.ts
```diff
@@ -115,18 +114,18 @@ export async function isMultisitePrimaryDomainMapped(
...
 	if ( ! environments.length ) {
 		return false;
 	}
 
-	const mappedDomains = environments[ 0 ]?.domains?.nodes?.map( domain => domain.name );
-
+	const mappedDomains = environments[ 0 ]?.domains?.nodes?.map( domain => domain?.name ) ?? [];
```
abdullah-kasim, line 129: You did it right here though!

abdullah-kasim (approved):

src/lib/validations/is-multisite-domain-mapped.ts
sjinks: We have a check a couple lines above:

```ts

if ( ! Array.isArray( res.data.app?.environments ) ) {
    return false;
}
```

If `res.data.app` is undefined, we don't reach this line. That's why `!`.

PR #1429, opened 2023-07-06
Add codegen for GraphQL queries

abdullah-kasim:
## Description

Adds the GraphQL Code generator.

Best to review it by ignoring all commits that says 'Regenerate graphql types'.

## Steps to Test

Example:

1. Check out PR.
1. Get a hold of `schema.gql` and paste it in project root - this is the schema of the endpoint that
   we communicate with.
2. Run `npm run typescript:codegen:install-dependencies` - this will install the codegen
   dependencies without updating `package.json`
3. Run `npm run typescript:codegen:generate` - this will regenerate the types


tsconfig.json
```diff
@@ -28,7 +32,8 @@
     "config",
     "__tests__",
     "__fixtures__",
-    "noImplictAnyImportBypass.d.ts"
+    "noImplictAnyImportBypass.d.ts",
+    "codegen.ts"
```
abdullah-kasim: Originally we added `"./node_modules/@types/node/globals.d.ts"` for globals, but I think line 22 does it better.

I've tested it by typing `console` in a file and the typescript language server doesn't try to import it. The code completion seems to be working, too.

src/lib/api/http.ts
```diff
@@ -52,6 +52,7 @@ export default async ( path: string, options: FetchOptions = {} ): Promise< Resp
 			'Content-Type': 'application/json',
 			...( options.headers ?? {} ),
 		},
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
abdullah-kasim, line 55: Needed to quickly fix this lint error as it wasn't really related to my commit. Might be my overzealous git hook for catching this though.

tsconfig.json
```diff
@@ -25,9 +32,10 @@
     "config",
     "__tests__",
     "__fixtures__",
-    "noImplictAnyImportBypass.d.ts",
-    // built-in globals like `console` doesn't work properly with the TypeScript language server without the following line
-    "./node_modules/@types/node/globals.d.ts"
```
abdullah-kasim, line 30: Originally we added `"./node_modules/@types/node/globals.d.ts"` for globals, but I think line 22 does it better.

I've tested it by typing `console` in a file and the typescript language server doesn't try to import it. The code completion seems to be working, too.

package.json
```diff
@@ -54,6 +54,8 @@
     "vip-wp": "dist/bin/vip-wp.js"
   },
   "scripts": {
+    "typescript:codegen:install-dependencies": "npm install --no-save @graphql-codegen/cli @graphql-codegen/typescript @graphql-codegen/near-operation-file-preset @graphql-codegen/typescript-operations",
```
abdullah-kasim, line 57: What do you think of this? I do prefer adding the dev dependency to the project itself, but perhaps we don't need this too often.

We could also create a mini sub-project just to generate the types if that's preferred.

codegen.ts
```diff
@@ -0,0 +1,24 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { CodegenConfig } from '@graphql-codegen/cli';
+
+const config: CodegenConfig = {
+	schema: './schema.gql',
+	ignoreNoDocuments: true, // for better experience with the watcher
+	hooks: { afterAllFileWrite: [ 'prettier --write' ] },
```
abdullah-kasim, line 7: For better interoperability with prettier!

package.json
sjinks: Whatever works. I am OK with either way.

sjinks (approved):

sjinks:
Awesome, thank you!

PR #1433, opened 2023-07-06
Redeploy codegen for GraphQL queries.

abdullah-kasim:
# Description

Let's wait a day before redeploying this as today is a release day.


abdullah-kasim:
I'll re-merge this as it does look like our release is okay.

If there's any new release that we need to do, it needs to be a new version.

PR #1434, opened 2023-07-06
Add more clarifying steps for making a release

abdullah-kasim:
## Description

Updates the docs for publishing a little to hopefully streamline the process a bit better for the next release.

I've also removed parts of the process where the commands doesn't exist anymore i.e. I ran `npm prepack` or `npm run prepack` but it doesn't look like those commands are there.

Not too sure if it'll help though, and shouldn't be needed once we've finalized the GitHub publish process.

Or, we could use the GitHub publish process now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,25 +97,22 @@ We use a custom pre-publish [script](https://github.com/Automattic/vip/blob/trun
 
 Further checks can be added to this flow as needed.
 
-### Pre-publish Tasks
-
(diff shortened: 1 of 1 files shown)


abdullah-kasim:
The publish button that uses `vip-actions` would require us to stick with the `develop` branch and get rid of `trunk` (or vice versa) so not too sure what to do there 🤔 .

It's already in this repo, but we haven't used it yet due to that reason.

sjinks (approved):

PR #1443, opened 2023-07-12
On-demand Database Backup: Use a less error-prone way of detecting if we're rate limited

abdullah-kasim:
## Description

A slight refactor on how we detect if we're rate limited. Makes things look nicer, and if the error message changes later, no change is needed on our side.

~~This is still awaiting changes from the server side though.~~

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-export-sql.js --generate-backup` **twice**. The second one should be an error.


sjinks:
Could you please take a look at ESLint warnings? Thank you!

abdullah-kasim:
Sure!

sjinks (approved):

PR #1456, opened 2023-07-21
BB8-10453: Fix DB engine validation regex

hanifn:
Change regex to take into account spaces before and after the `=`

## Description

This PR fixes an issue with the SQL validation logic where the `ENGINE` type not being `InnoDB` is not caught by the validator due to spaces in the target SQL file. E.g.:
```
ENGINE = MyISAM // missed
ENGINE=MyISAM // caught
```
The regex has been updated to take the optional spaces into account.

## Steps to Test

1. Get a SQL dump from an existing site.
1. Edit one of the `CREATE TABLE` lines to use `MyISAM` engine with spaces: `ENGINE = MyISAM` 
1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip-import-validate-sql @{target test site} test-dump.sql`
1. Verify that the error was caught. E.g.:
```
Error:  SQL Error: ENGINE != InnoDB on line(s) 40.
Recommendation: Ensure your application works with InnoDB and update your SQL dump to include only 'ENGINE=InnoDB' engine definitions in 'CREATE TABLE' statements. We suggest you search for all 'ENGINE=X' entries and replace them with 'ENGINE=InnoDB'!

SQL validation failed due to 1 error(s)
```

### DIFF
--- a/src/lib/validations/sql.ts
+++ b/src/lib/validations/sql.ts
@@ -337,7 +337,7 @@ const checks: Checks = {
 		recommendation: "Use search-replace to change environment's domain",
 	},
 	engineInnoDB: {
-		matcher: / ENGINE=(?!(InnoDB))/i,
+		matcher: / ENGINE\s?=(?!(\s?InnoDB))/i,
 		matchHandler: lineNumber => ( { lineNumber } ),
 		outputFormatter: lineNumberCheckFormatter,
 		results: [],
--- a/src/lib/validations/sql.ts
+++ b/src/lib/validations/sql.ts
@@ -337,7 +337,7 @@ const checks: Checks = {
 		recommendation: "Use search-replace to change environment's domain",
 	},
 	engineInnoDB: {
-		matcher: / ENGINE\s?=(?!(\s?InnoDB))/i,
+		matcher: /\sENGINE\s?=(?!(\s?InnoDB))/i,
 		matchHandler: lineNumber => ( { lineNumber } ),
 		outputFormatter: lineNumberCheckFormatter,
 		results: [],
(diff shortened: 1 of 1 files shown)


src/lib/validations/sql.ts
```diff
@@ -337,7 +337,7 @@ const checks: Checks = {
 		recommendation: "Use search-replace to change environment's domain",
 	},
 	engineInnoDB: {
-		matcher: / ENGINE=(?!(InnoDB))/i,
+		matcher: / ENGINE\s?=(?!(\s?InnoDB))/i,
```
saroshaga: Wonder why there's a mandatory whitespace in the beginning?

saroshaga: Do we need to make that optional too?

hanifn: I don't think so. If there's no space in the beginning then `ENGINE` would turn into a totally different word wouldn't it?

saroshaga: oh I thought it always expects it on a newline.

What if the `ENGINE=MyISAM` clause is on a completely newline?

hanifn: Hmm.. that's a good point. I'll test it out

hanifn: Changed in a1c7999379dc4eb8c4b24d6e7a25bd8138c56945
I've added a requirement for there to be at least 1 whitespace character before `ENGINE`. This includes space, tab and newline characters

saroshaga: @hanifn can't see the commit against this PR for some reason, but it looks good

hanifn: > can't see the commit against this PR for some reason, but it looks good

Yeah same here. Weird 🤔

saroshaga: possibly this: https://www.githubstatus.com/incidents/6503rcn8s34s

saroshaga (approved):
nice fix!

ingeniumed:
Mentioning this for posterity: Updated the runbook to mention this as a PagerDuty alert happened that would not happen once this goes out the door - https://github.com/Automattic/vip-docs/pull/230

PR #1458, opened 2023-07-24
Add warning confirmation when there's not enough space on disk.

abdullah-kasim:
## Description

Hello! This adds a warning confirmation when there's not enough space on disk for the `vip dev-env sync sql` and the `vip export sql` commands.

A lot of manual testing is involved because it's very machine dependent, and some unit tests for things that aren't, of course.

There's a small edge case in linux where the Docker machine's storage size is the same as the linux machine's storage size, because in linux, docker runs natively without going through a VM. We'll tackle this on a separate PR.

## Steps to Test

There's a bunch of things that can't be unit tested such as how much space is in the current machine and how much space is in Docker Desktop's machine - so please keep that in mind.

### Testing to see if the machine and Docker machine has the correct amount of space

1. Run `npm run build:watch` in a terminal.
2. Create a script named `backup-storage-test.js` with the following content in the project root
```js
const { BackupStorageAvailability } = require('./dist/lib/backup-storage-availability/backup-storage-availability')

const test = async () => {
  const storageAvailability = new BackupStorageAvailability(1234)
  console.log("getDockerStorageAvailable: ", storageAvailability.bytesToHuman(storageAvailability.getDockerStorageAvailable()))
  console.log("getStorageAvailableInVipPath: ", storageAvailability.bytesToHuman(await storageAvailability.getStorageAvailableInVipPath()))
}

test().then().catch(err => console.log("err: ", err.message, err.stack))
```
3. Run the script by writing `node backup-storage-test.js`
4. You should get the following output:
```
getDockerStorageAvailable:  32.51 GB
getStorageAvailableInVipPath:  635.83 GB
```
5. Compare with your machine's storage space available. On Mac, getStorageAvailableInVipPath value is equivalent to the free space in 'Disk Utility', and getDockerStorageAvailable value is equivalent to the space shown in Docker Desktop's bottom status bar i.e. `Disk 32.51 GB avail. of 62.67 GB`  Note: Windows uses GB/MB/KB erroneously. They're actually showing GiB instead of GB and so on.

### Testing if the prompt works as expected.

1. Run `npm run build:watch` in a terminal.
1. To simulate the scenario where we're low on docker storage available and machine storage available, apply the following patch. Protip: copy the text and run `pbpaste | git apply` at project root to apply the patch
```patch
Subject: [PATCH] Fix change in mock return value
---
Index: src/lib/backup-storage-availability/backup-storage-availability.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/lib/backup-storage-availability/backup-storage-availability.ts b/src/lib/backup-storage-availability/backup-storage-availability.ts
--- a/src/lib/backup-storage-availability/backup-storage-availability.ts	(revision 1f9753cbe67be0357d2a8e179228a580b5f7d2af)
+++ b/src/lib/backup-storage-availability/backup-storage-availability.ts	(date 1690853508936)
@@ -35,13 +35,7 @@
 	}
 
 	getDockerStorageAvailable(): number {
-		const kiBLeft = this.getDockerStorageKiBRaw();
-
-		if ( ! kiBLeft || Number.isNaN( Number( kiBLeft ) ) ) {
-			throw new DockerMachineNotFoundError();
-		}
-
-		return Number( kiBLeft ) * 1024;
+		return 0;
 	}
 
 	bytesToHuman( bytes: number ) {
@@ -49,10 +43,7 @@
 	}
 
 	async getStorageAvailableInVipPath() {
-		const vipDir = path.join( xdgBasedir.data ?? os.tmpdir(), 'vip' );
-
-		const diskSpace = await checkDiskSpace( vipDir );
-		return diskSpace.free;
+		return 0;
 	}
 
 	getReserveSpace(): number {
```
3. Then, run `node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production` and see the prompt comes up. This is equivalent to running the `vip dev-env sync sql` command
4. You could also run `node ./dist/bin/vip-export.sql.js @6675.production`. This is equivalent to running the `vip export sql` command.

The prompt looks like the following:
```shell
node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-07-26T04:20:31.000Z
Attaching to an existing export for the backup with timestamp 2023-07-26T04:20:31.000Z
✓ Preparing for backup download 
✓ Creating backup copy 
⠴ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
? We recommend that you have at least 1.07 GB of free space in your machine to import this database backup. Do you still want to continue with importing the database backup? (y/N) › false
```


src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+	archiveSize: number;
+
+	constructor( archiveSize: number ) {
+		this.archiveSize = archiveSize;
+	}
+
+	static createFromDbCopyJob( job: Job ): BackupStorageAvailability {
+		const bytesWrittenMeta = job.metadata?.find( meta => meta?.name === 'bytesWritten' );
+		if ( ! bytesWrittenMeta?.value ) {
+			throw new Error( 'Meta not found' );
```
aswasif007, line 23: Any chance `bytesWrittenMeta.value` can be `0`?

If so, it will throw exception, even though meta is found.

src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+			.grep( /\/dev\/vda1/ )
+			.head( { '-n': 1 } )
+			.replace( /\s+/g, ' ' )
+			.split( ' ' )[ 3 ];
+
+		return Number( kiBLeft ) * 1024;
+	}
+
+	bytesToHuman( bytes: number ) {
+		const sizeInGB = ( bytes / ( 1000 * 1000 * 1000 ) ).toFixed( 2 );
```
aswasif007: Curious why we are dividing with `1000 * 1000 * 1000`, but not `1024 * 1024 * 1024`.

src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+		const bytesWrittenMeta = job.metadata?.find( meta => meta?.name === 'bytesWritten' );
+		if ( ! bytesWrittenMeta?.value ) {
+			throw new Error( 'Meta not found' );
+		}
+
+		return new BackupStorageAvailability( Number( bytesWrittenMeta.value ) );
+	}
+
+	getDockerStorageAvailable(): number {
+		const kiBLeft = exec( `docker run --rm alpine df -k`, { silent: true } )
```
t-wright: This is clever. Surprising to me that this info isn't available with `docker system df`.

src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+	// eslint-disable-next-line id-length
+	async validateAndPromptDiskSpaceWarningForBackupImport(): Promise< boolean > {
+		const isStorageAvailable =
+			( await this.getStorageAvailableInVipPath() ) > this.getArchiveSize();
+		if ( ! isStorageAvailable ) {
+			const storageRequired = this.getArchiveSize();
+			const confirmPrompt = new Confirm( {
+				message: `We recommend that you have at least ${ this.bytesToHuman(
+					storageRequired
+				) } of free space in your machine to download this database backup. Do you still want to continue with downloading the database backup?`,
```
t-wright, line 96: Since we've assessed that they don't have sufficient space, should we say that in all these messages, letting them know how much space they do have?

src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+		const kiBLeft = exec( `docker run --rm alpine df -k`, { silent: true } )
+			.grep( /\/dev\/vda1/ )
+			.head( { '-n': 1 } )
+			.replace( /\s+/g, ' ' )
+			.split( ' ' )[ 3 ];
+
+		return Number( kiBLeft ) * 1024;
+	}
+
+	bytesToHuman( bytes: number ) {
```
t-wright, line 47: Should we use this util function instead? https://github.com/Automattic/vip-cli/blob/40ea4b18d3b2174db265dbe914b549dfc9a5c422/src/lib/cli/format.ts#L230

t-wright (commented):
Tested and it worked as described.

src/lib/backup-storage-availability/backup-storage-availability.ts
t-wright: Probably quite unlikely for users to encounter this, but mentioning it all the same: `bytesToHuman()` can also end up rounding to 0.00GB for small archives resulting in a message `We recommend that you have at least 0.00 GB of free space in your machine...`  It only happens in `validateAndPromptDiskSpaceWarningForBackupImport()` because we use the raw archive size to determine the storage requirement.

I encountered this running an export from my test site which produced an archive that was 2359458 bytes.

src/lib/backup-storage-availability/backup-storage-availability.ts
abdullah-kasim: GB vs GiB difference. 1024 * 1024 * 1024 is for GiB.

I did GiB at first but realized Macs mostly use GB.

src/lib/backup-storage-availability/backup-storage-availability.ts
abdullah-kasim: Haha, I was TRYING to avoid saying it, in case I got the amount wrong 😆 . Sure let's do that!

src/lib/backup-storage-availability/backup-storage-availability.ts
abdullah-kasim: Don't praise me, praise ChatGPT! 😆 

Yep that surprises me too.

src/lib/backup-storage-availability/backup-storage-availability.ts
aswasif007: > GB vs GiB difference

TIL 👍

src/lib/backup-storage-availability/backup-storage-availability.ts
t-wright: I have faith in your methods 💪 You could say "We estimate you have..." or "According to our calculations you have..." to provide a bit of wiggle room 😅

src/lib/backup-storage-availability/backup-storage-availability.ts
abdullah-kasim: Sure, let's do that!

However, I'll need a small reimplementation if we want to stick with what Mac and Docker says about GB or it'll introduce confusion. Shouldn't be too hard...

src/lib/backup-storage-availability/backup-storage-availability.ts
```diff
@@ -0,0 +1,122 @@
...
+		if ( ! bytesWrittenMeta?.value ) {
+			throw new Error( 'Meta not found' );
+		}
+
+		return new BackupStorageAvailability( Number( bytesWrittenMeta.value ) );
+	}
+
+	getDockerStorageAvailable(): number {
+		const kiBLeft = exec( `docker run --rm alpine df -k`, { silent: true } )
+			.grep( /\/dev\/vda1/ )
```
sjinks, line 31:
```
$ ls -1 /dev | grep vda1
$
```

:-(

abdullah-kasim: On linux I presume? Running `docker run -rm alpine /bin/sh -c 'ls -1 /dev | grep vda1'` would expectedly repeat your scenario as Docker tries to expose as little as possible of the underlying system into the kernel. When I ran this inside the Docker machine instead `docker run -it --rm --privileged --pid=host justincormack/nsenter1`, it returns a result.

So I'm guessing on Linux, it wouldn't be `/dev/vda1`.

That does tell me that I should pretty much skip the validation for now, if I couldn't detect a number. We should at least not crash.

src/lib/backup-storage-availability/backup-storage-availability.ts
abdullah-kasim: Just checked, not possible. `bytesWrittenMeta.value` is a string

src/commands/export-sql.js
```diff
@@ -410,6 +424,20 @@ export class ExportSQLCommand {
...
+		const storageConfirmed = await this.confirmEnoughStorage( await this.getExportJob() );
+
+		this.progressTracker.startPrinting();
+
+		if ( storageConfirmed ) {
+			this.progressTracker.stepSuccess( this.steps.CONFIRM_ENOUGH_STORAGE );
+		} else {
+			this.progressTracker.stepFailed( this.steps.CONFIRM_ENOUGH_STORAGE );
+			this.stopProgressTracker();
+			return false;
```
aswasif007: Let's follow suit with other error cases, and do `exit.withError` ?

src/bin/vip-dev-env-sync-sql.js
```diff
@@ -86,6 +86,9 @@ command( {
 		const cmd = new DevEnvSyncSQLCommand( app, env, slug, trackerFn );
 		// TODO: There's a function called handleCLIException for dev-env that handles exceptions but DevEnvSyncSQLCommand has its own implementation.
 		// We should probably use handleCLIException instead?
-		await cmd.run();
+		const didCommandRun = await cmd.run();
+		if ( ! didCommandRun ) {
+			console.log( 'Command canceled by user.' );
```
aswasif007, line 91: I think this should be treated as an error case and should be prefixed with `Error:`.
I'd recommend not bubbling it up to the caller function and instead proceed with `exit.withError` right away.

src/bin/vip-export-sql.js
```diff
@@ -70,6 +70,9 @@ command( {
 			{ outputFile: output, generateBackup },
 			trackerFn
 		);
-		await exportCommand.run();
+		const didCommandRun = await exportCommand.run();
+		if ( ! didCommandRun ) {
+			console.log( 'Command cancelled by user.' );
```
aswasif007: Same as above.

aswasif007 (approved):
Reviewed and tested, and works as expected.

Got one more petty stuff in mind:
```
$ node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-08-02T12:43:09.000Z
Attaching to an existing export for the backup with timestamp 2023-08-02T12:43:09.000Z
✓ Preparing for backup download
✓ Creating backup copy
⠦ Checking if there's enough storage
○ Requesting download link
○ Downloading file
✓ Preparing for backup download
✓ Creating backup copy
✕ Checking if there's enough storage
○ Requesting download link
○ Downloading file
Command canceled by user.
```

Progress steps are printed twice when the user cancels. Can we fix that?

Not a blocker, so approving 🚀

abdullah-kasim:
@aswasif007 

> Progress steps are printed twice when the user cancels. Can we fix that?
> 
> Not a blocker, so approving 🚀

It's possible though it'll require us to modify the `ProgressTracker` library a little as manipulating `stopPrinting` and `startPrinting` alone is not enough . I think it's worth fixing as I can see this being used over and over again.

saroshaga:
> @aswasif007
> 
> > Progress steps are printed twice when the user cancels. Can we fix that?
> > Not a blocker, so approving 🚀
> 
> It's possible though it'll require us to modify the `ProgressTracker` library a little as manipulating `stopPrinting` and `startPrinting` alone is not enough . I think it's worth fixing as I can see this being used over and over again.

Feel free to disagree but I think it kinda is a blocker actually @abdullah-kasim 

From a UX perspective, it's not critical but it does not seem very VIP :) 
Also, we'd need to check if this has A11Y issues. 

Since this isn't a very critical bug fix, I think we should spend some time in removing the duplication

saroshaga (changes requested):
thanks for working on this ❤️ please see my comment above

abdullah-kasim:
> > @aswasif007
> > > Progress steps are printed twice when the user cancels. Can we fix that?
> > > Not a blocker, so approving 🚀
> > 
> > 
> > It's possible though it'll require us to modify the `ProgressTracker` library a little as manipulating `stopPrinting` and `startPrinting` alone is not enough . I think it's worth fixing as I can see this being used over and over again.
> 
> Feel free to disagree but I think it kinda is a blocker actually @abdullah-kasim
> 
> From a UX perspective, it's not critical but it does not seem very VIP :) Also, we'd need to check if this has A11Y issues.
> 
> Since this isn't a very critical bug fix, I think we should spend some time in removing the duplication

@saroshaga 

I've fixed it, but I'm overall not too happy with what's happening.

The progressTracker isn't really designed to have a prompt right in the middle of a process.

For example, if we have 10 progress steps, the progress tracker will keep refreshing the last 10 lines whenever you see the spinner. So if we put a prompt, we're interrupting that step, and because the prompt causes an offset of a few lines i.e. it's now 13 lines but the progress tracker is still refreshing the last 10 lines, you start seeing that duplication effect that we see.

Now, all this is fixable, but the fix also deletes the prompt's output - for example, the user wants to know what did they answer their previous prompt with.

I've thought of having the prompt sit at the bottom but that's going to make it lose context (as the current scenario would show that it occurred while we're at the confirmation step), and if we were to delete the last few lines so that to resolve the duplicate issue, it will still cause a UX issue where we're showing the prompt after the command has finished in the wrong location. 

I'd suggest we stick with having the duplicate, and rather, fix the part where the prompt was deleted. This also helps us with debugging as we could see what the user has answered us with - particularly helpful when we have multiple prompts (which we do, right now).

What do you think @saroshaga ?

abdullah-kasim:
Also pushed the changes where I retain the duplicate behavior but ensures that the prompt question and answer did not get deleted.

aswasif007:
Unfortunately, the success case is borked now 😞 

```
$ node ./dist/bin/vip export sql @wasif-test.production
✓ Latest backup found with timestamp 2023-08-09T11:16:06.000Z
Attaching to an existing export for the backup with timestamp 2023-08-09T11:16:06.000Z
✓ Preparing for backup download
✓ Creating backup copy
⠧ Checking if there's enough storage
○ Requesting download link
○ Downloading file
✓ Preparing for backup download
✓ Creating backup copy
✓ Checking if there's enough storage
✓ Requesting download link
✓ Downloading file - 100.00% (39.55 KB/39.55 KB)
File saved to /Users/aswasif007/repo/vip-cli/2023-08-09T1116UTC-wasif-test.go-vip.net-715f47c583ec5c18-db-backup.sql.gz
```

aswasif007:
And the error case hasn't improved much either:

```
node ./dist/bin/vip export sql @wasif-test.production
✓ Latest backup found with timestamp 2023-08-09T11:16:06.000Z
Attaching to an existing export for the backup with timestamp 2023-08-09T11:16:06.000Z
✓ Preparing for backup download
✓ Creating backup copy
⠇ Checking if there's enough storage
○ Requesting download link
○ Downloading file
✔ We recommend that you have at least 40.5 KB of free space in your machine to download this database backup. Do you still want to continue with downloading the database backup? (y/N) · false

✓ Preparing for backup download
✓ Creating backup copy
✕ Checking if there's enough storage
○ Requesting download link
○ Downloading file
Error:  Command canceled by user.
Debug:  VIP-CLI v2.32.4, Node v16.15.1, darwin 22.6.0
```

aswasif007:
Perhaps we could incorporate the confirmation messaging within the progress tracker, with a generic built-in step called `userInput` ? Just thinking out loud...

abdullah-kasim:
`npm run build:watch` running? Latest commit? I'll double check - I didn't see those errors on my end.

abdullah-kasim:
Case where there's a disk space issue:

```
$ node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-08-09T12:43:07.000Z
Exporting database backup with timestamp 2023-08-09T12:43:07.000Z
✓ Preparing for backup download 
✓ Creating backup copy 
⠦ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
✔ We recommend that you have at least 1.07 GB of free space in your machine to import this database backup. We estimate that you currently have 0 Bytes of space in your machine.
Do you still want to continue with importing the database backup?
 (y/N) · false

✓ Preparing for backup download 
✓ Creating backup copy 
✕ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
```
For a 'yes' answer:

```
node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-08-09T12:43:07.000Z
Attaching to an existing export for the backup with timestamp 2023-08-09T12:43:07.000Z
✓ Preparing for backup download 
✓ Creating backup copy 
⠸ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
✔ We recommend that you have at least 1.07 GB of free space in your machine to import this database backup. We estimate that you currently have 0 Bytes of space in your machine.
Do you still want to continue with importing the database backup?
 (y/N) · true

✔ We recommend that you have at least 1.07 GB of free space in your Docker machine to import this database backup. We estimate that you currently have 0 Bytes of space in your machine.
Do you still want to continue with importing the database backup? (y/N) · true

✓ Preparing for backup download 
✓ Creating backup copy 
✓ Checking if there's enough storage 
✓ Requesting download link 
✓ Downloading file - 100.00% (38.29 KB/38.29 KB)
File saved to /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-uCo239/sql-export.sql.gz
Extracting the exported file /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-uCo239/sql-export.sql.gz...
✓ Extracted to /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-uCo239/sql-export.sql
Extracting site urls from the SQL file...
Generating search-replace configuration...
Running the following search-replace operations on the SQL file:
  abdullah-aws-test.go-vip.net -> 6675production.vipdev.lndo.site
✓ Search-replace operation is complete
Importing the SQL file...
mysql: [Warning] Using a password on the command line interface can be insecure.
Success: Created user 4.
Success: User "vipgo" created.
✓ SQL file imported
```

For the case where there's no space only in the machine:

```
node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-08-09T12:43:07.000Z
Attaching to an existing export for the backup with timestamp 2023-08-09T12:43:07.000Z
✓ Preparing for backup download 
✓ Creating backup copy 
⠸ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
✔ We recommend that you have at least 1.07 GB of free space in your machine to import this database backup. We estimate that you currently have 0 Bytes of space in your machine.
Do you still want to continue with importing the database backup?
 (y/N) · true

✓ Preparing for backup download 
✓ Creating backup copy 
✓ Checking if there's enough storage 
✓ Requesting download link 
✓ Downloading file - 100.00% (38.29 KB/38.29 KB)
File saved to /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-6bP0O6/sql-export.sql.gz
Extracting the exported file /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-6bP0O6/sql-export.sql.gz...
✓ Extracted to /var/folders/zv/h4914qz12lq5mbkl151xh8hh0000gn/T/vip-cli-6bP0O6/sql-export.sql
Extracting site urls from the SQL file...
Generating search-replace configuration...
Running the following search-replace operations on the SQL file:
  abdullah-aws-test.go-vip.net -> 6675production.vipdev.lndo.site
✓ Search-replace operation is complete
Importing the SQL file...
mysql: [Warning] Using a password on the command line interface can be insecure.
Success: Created user 4.
Success: User "vipgo" created.
✓ SQL file imported
```

abdullah-kasim:
One improvement that I think I could do is make sure that we skip the steps that has completed.

We currently display all the steps when we're utilizing the ProgressTracker so it's not doable to just push in the prompt between the current step and the next step.

EDIT: Done! See latest commit at https://github.com/Automattic/vip-cli/pull/1458/commits/a1bb62f79f2e68a49255abc65bc34fdb5d71dc06

abdullah-kasim:
Looks better!

```
$ node ./dist/bin/vip-dev-env-sync-sql.js @6675.production --slug=6675production
✓ Latest backup found with timestamp 2023-08-09T12:43:07.000Z
Attaching to an existing export for the backup with timestamp 2023-08-09T12:43:07.000Z
✓ Preparing for backup download 
✓ Creating backup copy 
⠸ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
✔ We recommend that you have at least 1.07 GB of free space in your machine to import this database backup. We estimate that you currently have 0 Bytes of space in your machine.
Do you still want to continue with importing the database backup?
 (y/N) · false

✕ Checking if there's enough storage 
○ Requesting download link 
○ Downloading file 
Error:  Command canceled by user.
Debug:  VIP-CLI v2.32.4, Node v16.15.1, darwin 22.6.0
```

abdullah-kasim:
> Perhaps we could incorporate the confirmation messaging within the progress tracker, with a generic built-in step called userInput ? Just thinking out loud...

@aswasif007 `handleContinuePrompt` does almost everything related to having such a step except for the glyph. It's a few steps away from having some sort of a `step type`. However, I don't think that'll fix the issue - the root problem is that we're displaying all the steps right now and the question arises - where do we display the prompt while ensuring things still looks chronologically accurate, and look the least awkward?

Do we.. not show all the steps so that we could show the prompt immediately the step list? Do we... delete the non-started steps above the prompt after finishing the prompt (I'm not entirely sure how doable this is)?

I decided on repeating the steps as I think it shows the chronological order best. FWIW, chronological order is important to reduce confusion and to help us debug when any issue arises.

aswasif007 (approved):
Thank you for working through all the changes, @abdullah-kasim . Looks good now! 🎉

abdullah-kasim:
Merging as @saroshaga is AFK.I think i've resolved the initial ask on repetition looking weird.

PR #1467, opened 2023-07-25
Mount vip-config directly in wp

pschoffer:
## Description

We load vip-config.php currently from `/wp/wp-content/vip-config/vip-config.php` however in live environments the vip-config folders is mapped directly to `/wp/vip-config`  (not under `wp-content`).

We should migrate the location to more closely match the prod environment. That way we can support other files too such as `client-sunrise.php` which [is loaded ](https://github.com/Automattic/vip-go-mu-plugins/blob/7294c3902d473c57c58e8505d4ad396de3470d0b/lib/sunrise/sunrise.php#L100-L106)from  `/wp/vip-config/client-sunrise.php`

This is the second part of the change where we change mount location

This should only be merged **AFTER** https://github.com/Automattic/vip-container-images/pull/516

## Steps to Test

1) Create new dev-env 
2) Attach shell to php contianer and see that vip-config is loaded under `/wp/vip-config` and everything still works.


rinatkhaziev:
@pschoffer  What would happen with the existing environments?

pschoffer:
Hmm I did test it before @rinatkhaziev but now I looked more into it and I found a scenario that breaks it a bit. I dropped the details in slack, but in essance if existing environment updates after this change and starts to mount vip-config to new location it might get ignored. Because we only [update](https://github.com/Automattic/vip-container-images/blob/master/dev-tools/setup.sh#L91) `wp-config-defaults.php`  if wp-config.php doenst exist

pschoffer:
I think we should do this - https://github.com/Automattic/vip-container-images/pull/519/files that way update won't break existing environments after this change.

rbcorrales:
@rinatkhaziev Is there anything we can do to move this PR forward? It seems that the work required on the [vip-container-images repo](https://github.com/Automattic/vip-container-images/pull/516) was already handled. Let me know if there's anything I can do to help.

sjinks:
Implemented in #2124

PR #1484, opened 2023-08-09
feat(backup): Add vip backup db command

aswasif007:
## Description

In this PR, we are adding a new command `vip backup db` to trigger a new backup for the specified environment.

This PR instantly triggers a new backup for the selected environment, just like the similar feature in the VIP Dashboard. It also comes with the same limitation as VIP Dashboard: there can be one backup in 15 minutes of rolling window for production environments. And for non-prod, the window is 60 minutes.

Below are two example runs; one for success and the other for failure.

```
$ vip backup db @vip-test.testing
Generating a new database backup...
✓ Preparing for backup generation
✓ Generating backup
New database backup created
```

```
$ vip backup db @vip-test.testing
Generating a new database backup...
✕ Preparing for backup generation
○ Generating backup
Error:  A new database backup was not generated because a recently generated backup already exists.
If you would like to run the same command, you can retry in 57 minutes
Alternatively, you can export the latest existing database backup by running: vip @app.env export sql, right away.
Learn more about limitations around generating database backups: https://docs.wpvip.com/technical-references/vip-dashboard/backups/#0-limitations
```

## Steps to Test

1. `rm -rf dist && npm run build && npm link`
2. `node ./dist/bin/vip backup db @<app>.<env>`


src/lib/cli/format.ts
```diff
@@ -241,3 +241,37 @@ export const formatBytes = ( bytes: number, decimals = 2 ): string => {
...
+ * @param {Date} to  The end date
+ * @returns {string} The duration between the two dates
+ */
+export function formatDuration( from: Date, to: Date ): string {
+	const millisecondsPerSecond = 1000;
+	const millisecondsPerMinute = 60 * millisecondsPerSecond;
+	const millisecondsPerHour = 60 * millisecondsPerMinute;
+	const millisecondsPerDay = 24 * millisecondsPerHour;
+
+	const duration = Math.abs( from.getTime() - to.getTime() );
```
abdullah-kasim: We probably want to set the duration to 0 if it's negative due to race conditions or similar. And maybe ask the user to retry in 5 seconds?

src/commands/backup-db.ts
```diff
@@ -194,7 +188,10 @@ export class BackupDBCommand {
 						stack: error.stack,
 					} );
 					const errMessage = `A new database backup was not generated because a recently generated backup already exists.
-If you would like to run the same command, you can retry on or after: ${ retryAfter }
+If you would like to run the same command, you can retry on or after ${ formatDuration(
```
abdullah-kasim: Minor:

```suggestion
If you would like to run the same command, you can retry in ${ formatDuration(
```

abdullah-kasim (changes requested):
Looks mostly good to me!

There's just one place that I think needs changing though.

src/lib/cli/format.ts
```diff
@@ -254,3 +254,38 @@ export const formatBytes = (
...
+ * @returns {string} The duration between the two dates
+ */
+export function formatDuration( from: Date, to: Date ): string {
+	const millisecondsPerSecond = 1000;
+	const millisecondsPerMinute = 60 * millisecondsPerSecond;
+	const millisecondsPerHour = 60 * millisecondsPerMinute;
+	const millisecondsPerDay = 24 * millisecondsPerHour;
+
+	let duration = Math.abs( from.getTime() - to.getTime() );
+	if ( duration < 0 ) duration = 0;
```
abdullah-kasim: When `to` is much earlier than `from`, the `duration` returned is still positive. The problem with this is that if given a duration `from` and `to` returns `30 seconds` for `formatDuration`, after 35 seconds, it will return `5 seconds` left, despite us going 5 seconds over the expected duration.

To sum it up, we need to add a test case where `from` is greater than `to` by, say, 10 seconds. Is the value returned as expected?

abdullah-kasim (changes requested):

src/lib/cli/format.ts
abdullah-kasim: This might have been missed earlier - `let duration = Math.abs( from.getTime() - to.getTime() );` will always return a positive value due to `Math.abs`. `Math.abs` does the mathematical equivalent of absolute value - any negative value passed to it will becomes positive, and positive values remain the same.

src/lib/cli/format.ts
```diff
@@ -254,3 +254,43 @@ export const formatBytes = (
...
+ * @returns {string} The duration between the two dates
+ */
+export function formatDuration( from: Date, to: Date ): string {
+	const millisecondsPerSecond = 1000;
+	const millisecondsPerMinute = 60 * millisecondsPerSecond;
+	const millisecondsPerHour = 60 * millisecondsPerMinute;
+	const millisecondsPerDay = 24 * millisecondsPerHour;
+
+	const duration = to.getTime() - from.getTime();
+	if ( duration < 0 ) return '0 second';
```
abdullah-kasim:
```suggestion
	if ( duration < 1000 ) return '0 second';
```

Edge case: If there's 999 milliseconds remaining and if I calculated correctly, it'd mean that we have 0 seconds, 0 days, 0 hours, 0 minutes, and we'd be returning a blank string!

I think we should fix this - 1 second is a long time 😅

abdullah-kasim (changes requested):

src/lib/cli/format.ts
aswasif007: Sure. Thanks for being diligent!! 🎉

abdullah-kasim (approved):

PR #1492, opened 2023-08-23
refactor: convert `site-import` directory to TypeScript

sjinks:
## Description

This PR converts files from the `src/lib/site-import` directory to TypeScript.

## Steps to Test

CI should pass.

### DIFF
--- a/src/lib/cli/format.ts
+++ b/src/lib/cli/format.ts
@@ -192,7 +192,10 @@ export class RunningSprite {
 	}
 }
 
-export function getGlyphForStatus( status: StepStatus, runningSprite: RunningSprite ): string {
+export function getGlyphForStatus(
+	status: StepStatus | string,
+	runningSprite: RunningSprite
+): string {
 	switch ( status ) {
 		case 'pending':
 			return '○';
--- a/src/lib/site-import/status.ts
+++ b/src/lib/site-import/status.ts
@@ -119,7 +119,6 @@ interface ImportFailedError {
 	commandOutput: string[] | null;
 	error: string;
 	stepName: string;
-	errorText: string;
 	launched: boolean;
 }
 
@@ -385,8 +384,6 @@ ${ maybeExitPrompt }
 						commandOutput: failedImportStep.output,
 						error: 'Import step failed',
 						stepName: failedImportStep.name,
-						// FIXME: Initially it was failedImportStep.error, but AppEnvironmentStatusProgressStep does not have this field
-						errorText: failedImportStep.result,
 						launched,
 					} as ImportFailedError );
 				}
--- a/src/lib/site-import/status.ts
+++ b/src/lib/site-import/status.ts
@@ -1,8 +1,4 @@
-/**
- * @flow
- * @format
- */
-
+/* eslint-disable complexity */
 /**
  * External dependencies
  */
@@ -15,9 +11,21 @@ import debugLib from 'debug';
  */
 import API from '../../lib/api';
 import { currentUserCanImportForApp } from '../../lib/site-import/db-file-import';
-import { ProgressTracker } from '../../lib/cli/progress';
+import { ProgressTracker, StepFromServer, StepStatus } from '../../lib/cli/progress';
 import * as exit from '../../lib/cli/exit';
 import { capitalize, formatEnvironment, getGlyphForStatus } from '../../lib/cli/format';
+import { AppQuery, AppQueryVariables } from './status.generated';
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client';
+import {
+	App,
+	AppEnvironment,
+	AppEnvironmentImportStatus,
+	AppEnvironmentStatusProgressStep,
+	Job,
+	JobProgressStep,
+	Maybe,
+	PrimaryDomainSwitchJob,
+} from '../../graphqlTypes';
 
 const debug = debugLib( 'vip:lib/site-import/status' );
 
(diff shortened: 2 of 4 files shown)


src/lib/cli/format.ts
```diff
@@ -192,7 +192,10 @@ export class RunningSprite {
 	}
 }
 
-export function getGlyphForStatus( status: StepStatus, runningSprite: RunningSprite ): string {
+export function getGlyphForStatus(
+	status: StepStatus | string,
```
sjinks, line 196: Because of this: https://github.com/Automattic/vip-cli/pull/1492/files#diff-e2db69e07a54f1d529171080f7d84e280b1dbb9b2e171ca55dc24e7e4fc571b4R207

We pass something that is not a `StepStatus` to this function.

src/lib/site-import/status.ts
```diff
@@ -64,34 +72,40 @@ const IMPORT_SQL_PROGRESS_QUERY = gql`
...
-	const [ environment ] = environments;
+	const environment = environments[ 0 ] ?? {};
 	const { importStatus, jobs, launched } = environment;
 
 	if ( ! environment.isK8sResident && ! jobs?.length ) {
 		return {};
 	}
 
-	const [ importJob ] = jobs;
+	const importJob = jobs?.[ 0 ] ?? {};
```
sjinks: TS complains that `jobs` could be `undefined`. In this case, destructuring will fail.

src/lib/site-import/status.ts
```diff
@@ -198,7 +220,9 @@ export async function importSqlCheckStatus( {
 		let statusMessage;
 		switch ( overallStatus ) {
 			case 'success':
-				statusMessage = `Success ${ sprite } imported data should be visible on your site ${ env.primaryDomain.name }.`;
+				statusMessage = `Success ${ sprite } imported data should be visible on your site ${
+					env.primaryDomain?.name ?? 'N/A'
+				}.`;
```
sjinks, line 225: These things are to keep ESLint happy: our ruleset does not alow `undefined` as a template parameter.

src/lib/site-import/status.ts
```diff
@@ -235,12 +259,12 @@ ${ maybeExitPrompt }
...
 
-	const getResults = () =>
+	const getResults = (): Promise< string | Job | PrimaryDomainSwitchJob > =>
 		new Promise( ( resolve, reject ) => {
-			const checkStatus = async () => {
+			const checkStatus = async (): Promise< unknown > => {
 				let status;
 				try {
-					status = await getStatus( api, app.id, env.id );
+					status = await getStatus( api, app.id ?? -1, env.id ?? -1 );
```
sjinks, line 267: In theory, according to GQL definitions, `id` could be `null` or `undefined`. I chose -1 to make the request fail.

src/lib/site-import/status.ts
```diff
@@ -276,9 +304,10 @@ ${ maybeExitPrompt }
 							jobStatus = 'error';
 						} else if ( statusSteps.every( ( { result } ) => result === 'success' ) ) {
 							jobStatus = 'success';
-							importJob.completedAt = new Date(
-								Math.max( ...statusSteps.map( step => step.finished_at ), 0 ) * 1000
-							).toUTCString();
+							const timestamps = statusSteps
+								.map( step => step.finished_at )
+								.filter( Boolean ) as number[];
+							importJob.completedAt = new Date( Math.max( ...timestamps, 0 ) * 1000 ).toUTCString();
```
sjinks, line 310: `Math.max()`, according to TS, does not like all kinds of `null` and `undefined`.

src/lib/site-import/status.ts
```diff
@@ -351,12 +384,11 @@ ${ maybeExitPrompt }
 						commandOutput: failedImportStep.output,
 						error: 'Import step failed',
 						stepName: failedImportStep.name,
-						errorText: failedImportStep.error,
```
sjinks, line 354: I had to drop `errorText`:
1. It is not used in `getErrorMessage()`
2. `AppEnvironmentStatusProgressStep` does not have the `error` field :-(

src/lib/site-import/status.ts
```diff
@@ -351,12 +384,11 @@ ${ maybeExitPrompt }
...
 						launched,
-					} );
+					} as ImportFailedError );
 				}
 
-				progressTracker.setStepsFromServer( jobSteps );
+				progressTracker.setStepsFromServer( jobSteps as unknown as StepFromServer[] );
```
sjinks, line 391: I am not very proud of this :-(

src/lib/site-import/status.ts
abdullah-kasim: nit: This change currently affects the logic a little. Maybe

```sh
const [ importJob ] = jobs ?? [];
```

instead?

abdullah-kasim: I was checking which line might care about this change - looks like this line: https://github.com/Automattic/vip-cli/blob/b149b4d9e770c83aac6f5a5260d1589d1bd6a359/src/lib/site-import/status.ts#L279-L279 might be affected by this change

src/lib/site-import/status.ts
```diff
@@ -100,7 +114,15 @@ async function getStatus( api, appId, envId ) {
 	};
 }
 
-function getErrorMessage( importFailed, launched = false ) {
+interface ImportFailedError {
+	inImportProgress: boolean;
+	commandOutput: string[] | null;
+	error: string;
+	stepName: string;
```
abdullah-kasim: nit: We could use `StepName | string` where `StepName` is an enum consisting of `import_preflights, importing_db, ...etc`

src/lib/site-import/status.ts
```diff
@@ -370,7 +402,9 @@ ${ maybeExitPrompt }
 
 				overallStatus = 'running';
 
-				setTimeout( checkStatus, IMPORT_SQL_PROGRESS_POLL_INTERVAL ); // NOSONAR
+				setTimeout( () => {
+					void checkStatus();
```
abdullah-kasim, line 406: TIL of the void operator.

abdullah-kasim (dismissed):

src/lib/site-import/status.ts
sjinks: Good catch, fixed!

src/lib/site-import/status.ts
sjinks: Done

abdullah-kasim (approved):

PR #1497, opened 2023-09-11
fix: remove import bypass for TypeScript

sjinks:
## Description

This PR fixes the error:

```
Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.
```

We fix this by removing `noImplicitAnyImportBypass.d.ts` and setting `module` to `Node16`.

* Update TypeScript to 5.2.2
* Update `@types/cli-table` (0.3.2), `@types/semver` (7.5.1), `@types/uuid` (9.0.4), `@types/xml2js` (0.4.12), `@types/js-yaml` (4.0.6), `@types/json2csv` (5.0.4), `@types/node-fetch` (2.6.5)
* Remove `@tsconfig/node16`

## Steps to Test

CI should pass.


abdullah-kasim (commented):
I took this opportunity to check if we still need the import bypass.

Turns out we don't anymore thanks to all of the effort in migrating this codebase to TypeScript. Only two TypeScript files were importing a javascript file. Hence, I took the liberty to migrate away from using import bypass, as well as migrating the two files.

Feel free to pull the changes from the `fix/tsconfig-abdullah` branch https://github.com/Automattic/vip-cli/compare/fix/tsconfig...fix/tsconfig-abdullah . I've also updated `eslint-plugin-wpvip` and `typescript` on that branch so that it's easier to test TypeScript 5.2.2.

Tested with `tsc` and `npm run cmd:lint -- ./src/bin`

EDIT: Might want to revert the `eslint-plugin-wpvip` update and only update the parser to the latest (not sure how, delete the shrinkwrap file and regenerate?). Looks like the latest update to `eslint-plugin-wpvip` made things stricter.

EDIT2: `npm run cmd:lint -- ./src/lib/site-import/status.ts` takes forever for some reason. I'm not sure what's up with that file.

sjinks:
> eslint-plugin-wpvip

https://github.com/Automattic/vip-cli/pull/1499

> EDIT2

Yes, I also noticed that :-( Not sure why.

abdullah-kasim:
I forgot to mention that with the changes in [this branch](https://github.com/Automattic/vip-cli/compare/fix/tsconfig...fix/tsconfig-abdullah), there doesn't seem to be any need to rewind ourselves back to Node10 module resolution and CommonJS.

What do you think? Future overrides can be added with an `index.d.ts` file in the `types` folder, and I don't see us importing anything from `src/bin/*` any time soon because they're all entrypoints.

sjinks:
> and I don't see us importing anything from src/bin/* any time soon because they're all entrypoints

That was actually on my to-do list 😄 I wanted to convert everything in `src` before we migrate from `args` to OCLIF.

> there doesn't seem to be any need to rewind ourselves back to Node10 module resolution and CommonJS

OK, I will take a closer look. We generate CommonJS modules anyway; as a consequence, the module resolution of Node16 or NodeNext will be the same as Node10 😀 We cannot migrate to ESM because of `require` statements all over the place (well, we can, but this will not be painless).

But yes, I would like to eliminate the module bypass thing.

sjinks:
> Feel free to pull the changes

Done!

abdullah-kasim (approved):
> OK, I will take a closer look. We generate CommonJS modules anyway; as a consequence, the module resolution of Node16 or NodeNext will be the same as Node10 😀 We cannot migrate to ESM because of require statements all over the place (well, we can, but this will not be painless).

Honestly speaking.. this module stuff has always confused the heck out of me. Whatever works is actually what I prefer 😅 .

Looks good, tested on my end.

PR #1513, opened 2023-09-22
refactor: use `semver` instead of `check-node-version`

sjinks:
## Description

This PR removes the `check-node-version` package and uses `semver` for version checks in `helpers/prepublishOnly.js`.

## Steps to Test

```js
const { minVersion, satisfies, valid } = require('semver');

const supported = require('./package.json').engines.node;
const current = process.version;

console.log(satisfies(current, supported));
console.log(current, supported);
console.log(valid(current), valid(minVersion(supported)));
```

:-)


abdullah-kasim (approved):
LGTM

PR #1526, opened 2023-10-09
feat(phpmyadmin): Add `vip  db phpmyadmin` command

aswasif007:
## Description

In this PR, we are adding a new command `vip db phpmyadmin` which generates a URL to `phpmyadmin`, and opens in the default browser.

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-db-phpmyadmin.js @6721.production`
1. Verify that it opened a URL with `/.wpvip/phpmyadmin/?token=...` prefix in your browser.

### DIFF
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -1,21 +1,20 @@
 /**
  * External dependencies
  */
-
+import { GraphQLFormattedError } from 'graphql';
 import gql from 'graphql-tag';
 import opn from 'opn';
 
 /**
  * Internal dependencies
  */
+import { App, AppEnvironment } from '../graphqlTypes';
 import API, {
 	disableGlobalGraphQLErrorHandling,
 	enableGlobalGraphQLErrorHandling,
 } from '../lib/api';
 import * as exit from '../lib/cli/exit';
 import { CommandTracker } from '../lib/tracker';
-import { App, AppEnvironment } from '../graphqlTypes';
-import { GraphQLFormattedError } from 'graphql';
 
 export const GENERATE_PHP_MY_ADMIN_URL_MUTATION = gql`
 	mutation GeneratePhpMyAdminAccess($input: GeneratePhpMyAdminAccessInput) {
@@ -90,6 +89,7 @@ export class PhpMyAdminCommand {
 			exit.withError( `Failed to generate PhpMyAdmin URL: ${ error.message }` );
 		}
 
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-call
 		void opn( url, { wait: false } );
 		this.log( 'PhpMyAdmin is opened in your default browser.' );
 	}
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -4,7 +4,6 @@
 import chalk from 'chalk';
 import { GraphQLFormattedError } from 'graphql';
 import gql from 'graphql-tag';
-import opn from 'opn';
 
 /**
  * Internal dependencies
@@ -66,6 +65,11 @@ export class PhpMyAdminCommand {
 		console.log( msg );
 	}
 
+	async openUrl( url: string ) {
+		const { default: open } = await import( 'open' );
+		void open( url, { wait: false } );
+	}
+
 	async run( silent = false ) {
 		this.silent = silent;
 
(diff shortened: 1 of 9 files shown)


src/commands/phpmyadmin.ts
```diff
@@ -0,0 +1,83 @@
...
+			exit.withError( 'Please specify an environment' );
+		}
+
+		this.log( 'Generating PhpMyAdmin URL...' );
+		const url = await generatePhpMyAdminAccess( this.env.id );
+
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-call
+		opn( url, { wait: false } );
+
+		this.log( 'Switch to your default browser.' );
```
saroshaga:
```suggestion
		this.log( `You can view your environment's data over at ${ url }` );
```

src/global.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'opn';
```
abdullah-kasim: See `<project-root>/types/copy-dir/index.d.ts` on how to add types for untyped node modules packages.

This is possible on this repo thanks to this line: https://github.com/Automattic/vip-cli/blob/4ccef58c0995d94c84017e403c6b16b0d4f107e9/tsconfig.json#L25-L25

abdullah-kasim:
Hope you don't mind, I made a tiny bit of change to make all our lives easier in the typescript land.

src/global.d.ts
abdullah-kasim: Whoops, it turns out the types are available at `@types/opn@5.1.0` so we'll use that instead. Don't worry about the mismatched versions - types from `@types` are generally safe to bring in as long as the major versions are the same - can't remember where I read it, but that has also been my experience so far.

However if we need to bring the types in - we could always copy paste back from there.

abdullah-kasim:
Types for graphql queries can be generated according to the instructions here: https://github.com/Automattic/vip-cli/blob/trunk/CONTRIBUTING.md#generating-the-types

src/commands/phpmyadmin.ts
aswasif007: We probably don't want to print the `url` because it contains the access token.

src/global.d.ts
saroshaga: btw, just FYI, this package has been through a bit of a rollercoaster ride : `open` was deprecated in favour of `opn` and now `opn` is also been deprecated and `open` has been brought back to life 😂 cc @aswasif007

aswasif007: 🎢 

Should we move to `open` and remove `opn` from VIP-CLI?

saroshaga: We can do that separately

aswasif007:
### 🔔 Note: Released with `@next` tag for testing: v2.36.0-dev.0 🔔

src/global.d.ts
abdullah-kasim: The new version of the package has started using ESM btw. 

And when it comes to ESM, trust me on this, but come back in 10 years before touching ESM. The tooling around it is pretty bad IMO.

aswasif007:
### 🔔 Note: Released with @next tag for testing: v2.36.2-dev.0 🔔

aswasif007:
### 🔔 Note: Released with `@next` tag for closed beta: `v2.37.0-dev.0` 🔔

aswasif007:
### 🔔 Note: Released with `@next` tag for open beta: `v2.38.0-dev.0` 🔔

aswasif007:
### 🔔 Note: Released with @next tag for open beta: v2.38.0-dev.1 🔔

saroshaga (approved):

src/commands/phpmyadmin.ts
```diff
@@ -0,0 +1,232 @@
...
+				environmentId: envId,
+			},
+		},
+	} );
+
+	// Re-enable global error handling
+	enableGlobalGraphQLErrorHandling();
+
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+	return resp?.data?.generatePHPMyAdminAccess?.url as string;
```
sjinks, line 97: Isn't this the wrong member? Shouldn't it be `resp.data?.enablePHPMyAdmin?.success` and be a `Boolean`, not `string`?

aswasif007: I think you are right 🤔

PR #1533, opened 2023-10-20
FORNO-1704: SQL Import - Improve handling of compressed files

t-wright:
## Description

This PR aims to address some inconsistencies around how compressed files are handled during SQL imports. 

1. Currently only uncompressed and `gz` files can be successfully imported to a VIP environment using `vip import sql`, but we don't validate extensions in the CLI.
1. Currently running `vip import validate-sql` on a compressed file will return normal validation errors like `Error:  SQL Error: DROP TABLE was not found.` even if the uncompressed file is valid.
2. Currently only uncompressed files can be imported using `vip dev-env import sql`; compressed files are attempted to be validated and encounter the same issue as described in 2 above.

This PR addresses these issues by:
1. Adding validation to only permit `.sql` and `.gz` extensions for both `vip import sql` and `vip dev-env import sql`
3. Adding support for `.gz` archives in `vip dev-env import sql`, extracting the archive to a temp directory before importing the resulting SQL file as normal
4. Returning a useful error on `vip import validate-sql` to explain that compressed files can't be validated

## Steps to Test

1. Download the test files below which contain this SQL:

```

  `ID` bigint(20) UNSIGNED NOT NULL,
  `timestamp` bigint(20) UNSIGNED NOT NULL,
  `args` longtext NOT NULL,
  `interval` int(10) UNSIGNED DEFAULT 0,
  `created` datetime NOT NULL,
  `last_modified` datetime NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

`.sql`: https://transfer.sh/5L7MiLwtQH/test.sql
`.sql.gz`: https://transfer.sh/1t712ht7JL/test.sql.gz
`.zip`: https://transfer.sh/KJOzR1paS9/test.zip

2. Check out PR.
3. Run `npm run build`

Validation flow:

4. Run `node ./dist/bin/vip-import-validate-sql.js test.zip`
5. Run `node ./dist/bin/vip-import-validate-sql.js test.sql.gz`
6. Verify the command exits and an error is shown
7. Run `node ./dist/bin/vip-import-validate-sql.js test.sql`
8. Verify the sql file validates

Dev Env flow:

9. Run `vip dev-env start --slug <your slug>` to start your local development environment
1. Run `node ./dist/bin/vip-dev-env-import-sql.js test.zip  --slug <your slug>`
2. Verify the command fails with an error about the extension being unsupported.
1. Run `node ./dist/bin/vip-dev-env-import-sql.js test.sql.gz --slug <your slug>`
1. Run `node ./dist/bin/vip-dev-env-import-sql.js test.sql --slug <your slug>`
1. Verify both commands succeed and the file is imported

Import flow:

15. Run `node ./dist/bin/vip-import-sql.js test.zip @<id>.production`
2. Verify the command fails with an error about the extension being unsupported.
1. Run `node ./dist/bin/vip-import-sql.js test.sql @<id>.production` (don't confirm)
3. Run `node ./dist/bin/vip-import-sql.js test.sql.gz @<id>.production` (don't confirm)


src/commands/dev-env-import-sql.js
```diff
@@ -34,6 +37,24 @@ export class DevEnvImportSQLCommand {
 		const lando = await bootstrapLando();
 		await validateDependencies( lando, this.slug, silent );
 
+		validateImportFileExtension( this.fileName );
```
saroshaga, line 40: Should we be validating the file name here too?

saroshaga (approved):
Code looks good! Haven't tested all flows yet

src/commands/dev-env-import-sql.js
t-wright: Doesn't appear to be necessary. I think the filename validation we do in `vip import sql` is a result of S3 filename requirements. We _could_ apply the same validation to keep requirements consistent between environments, though it would be an artificial limitation 🤔

src/lib/validations/sql.ts
```diff
@@ -86,6 +88,30 @@ interface ValidationResult {
...
 
+/**
+ * Check if a file has a valid extension
+ *
+ * @param {string} filename The file extension
+ * @returns {boolean} True if the extension is valid
+ */
+export const validateImportFileExtension = ( filename: string ): void => {
+	const ext = path.extname( filename ).toLowerCase();
+	if ( ! [ '.sql', '.gz' ].includes( ext ) ) {
```
saroshaga, line 99: In the public [docs](https://docs.wpvip.com/how-tos/migrate-content-databases/import-with-vip-cli/) we tell folks that we do not support compressed file imports

> A SQL database file (e.g. downloaded from [Database Backups](https://docs.wpvip.com/technical-references/vip-dashboard/backups/)) must already be located on the local machine under the home directory of the current user. The SQL database file must be decompressed (unarchived) before import.

If we are starting to support it, I think we'll need to document the extensions that we support henceforth ?

saroshaga: Also any particular reason for not supporting other formats?

t-wright: > If we are starting to support it, I think we'll need to document the extensions that we support henceforth ?

We have always supported `.gz` files in `vip import sql`. We're just aligning `vip dev-env import sql` to support them as well. Agree this should be mentioned in our docs following the release of this. cc @yolih 

The archive must be extracted for `vip import validate-sql`

> Also any particular reason for not supporting other formats?

I think it would be good to support `.zip` as well, but I feel like that's outside the scope of this PR.

saroshaga: > The archive must be extracted for vip import validate-sql

better to make this clear in the docs too!

saroshaga (approved):
thanks for working on this!

PR #1535, opened 2023-10-27
FORNO-1677: Fix dev env network site domains

hanifn:
## Description

Changes the generated domains for network sites to be more user-friendly

E.g.:
```
[subsite-id].[main-site-slug].vipdev.lndo.site
 to
[subsite-id]-[subsite-slug].[main-site-slug].vipdev.lndo.site
```

## Steps to Test

1. Check out PR.
1. Run `npm run build`
2. Create a new dev env for a multisite using `./dist/bin/vip-dev-env-create.js`
3. Run `./dist/bin/vip-dev-env-start.js`
1. Run `./dist/bin/vip-dev-env-sync-sql.js`
1. Verify that network sites' domains match new format and are accessible from the browser

### DIFF
--- a/__tests__/commands/dev-env-sync-sql.js
+++ b/__tests__/commands/dev-env-sync-sql.js
@@ -71,6 +71,18 @@ jest.spyOn( console, 'log' ).mockImplementation( () => {} );
 describe( 'commands/DevEnvSyncSQLCommand', () => {
 	const app = { id: 123, name: 'test-app' };
 	const env = { id: 456, name: 'test-env', wpSitesSDS: {} };
+	const msEnv = {
+		id: 456,
+		name: 'test-env',
+		wpSitesSDS: {
+			nodes: [
+				{
+					blogId: 2,
+					homeUrl: 'https://subsite.com',
+				},
+			],
+		},
+	};
 
 	describe( '.generateExport', () => {
 		it( 'should create an instance of ExportSQLCommand and run', async () => {
@@ -92,6 +104,18 @@ describe( 'commands/DevEnvSyncSQLCommand', () => {
 			cmd.generateSearchReplaceMap();
 
 			expect( cmd.searchReplaceMap ).toEqual( { 'test.go-vip.com': 'test-slug.vipdev.lndo.site' } );
+		} );
+
+		it( 'should return a map of search-replace values for multisite', () => {
+			const cmd = new DevEnvSyncSQLCommand( app, msEnv, 'test-slug' );
+			cmd.slug = 'test-slug';
+			cmd.siteUrls = [ 'test.go-vip.com', 'subsite.com' ];
+			cmd.generateSearchReplaceMap();
+
+			expect( cmd.searchReplaceMap ).toEqual( {
+				'test.go-vip.com': 'test-slug.vipdev.lndo.site',
+				'subsite.com': '2.subsite-com.test-slug.vipdev.lndo.site',
+			} );
 		} );
 	} );
 
--- a/__tests__/commands/dev-env-sync-sql.js
+++ b/__tests__/commands/dev-env-sync-sql.js
@@ -114,7 +114,7 @@ describe( 'commands/DevEnvSyncSQLCommand', () => {
 
 			expect( cmd.searchReplaceMap ).toEqual( {
 				'test.go-vip.com': 'test-slug.vipdev.lndo.site',
-				'subsite.com': '2.subsite-com.test-slug.vipdev.lndo.site',
+				'subsite.com': '2-subsite-com.test-slug.vipdev.lndo.site',
 			} );
 		} );
 	} );
--- a/__tests__/commands/dev-env-sync-sql.js
+++ b/__tests__/commands/dev-env-sync-sql.js
@@ -114,7 +114,7 @@ describe( 'commands/DevEnvSyncSQLCommand', () => {
 
 			expect( cmd.searchReplaceMap ).toEqual( {
 				'test.go-vip.com': 'test-slug.vipdev.lndo.site',
-				'subsite.com': '2-subsite-com.test-slug.vipdev.lndo.site',
+				'subsite.com': 'subsite-com-2.test-slug.vipdev.lndo.site',
 			} );
 		} );
 	} );
--- a/package.json
+++ b/package.json
@@ -163,6 +163,7 @@
     "semver": "7.5.4",
     "shelljs": "^0.8.5",
     "single-line-log": "1.1.2",
+    "slugify": "^1.6.6",
     "socket.io-client": "^4.5.3",
     "socket.io-stream": "npm:@wearemothership/socket.io-stream@^0.9.1",
     "socks-proxy-agent": "^5.0.1",
--- a/package.json
+++ b/package.json
@@ -163,7 +163,6 @@
     "semver": "7.5.4",
     "shelljs": "^0.8.5",
     "single-line-log": "1.1.2",
-    "slugify": "^1.6.6",
     "socket.io-client": "^4.5.3",
     "socket.io-stream": "npm:@wearemothership/socket.io-stream@^0.9.1",
     "socks-proxy-agent": "^5.0.1",
--- a/src/commands/dev-env-sync-sql.js
+++ b/src/commands/dev-env-sync-sql.js
@@ -160,9 +160,9 @@ export class DevEnvSyncSQLCommand {
 			if ( ! this.searchReplaceMap[ url ] ) continue;
 
 			slugify.extend( { '.': '-' } );
-			this.searchReplaceMap[ url ] = `${ site.blogId }-${ slugify( url, {
+			this.searchReplaceMap[ url ] = `${ slugify( url, {
 				lower: true,
-			} ) }.${ this.landoDomain }`;
+			} ) }-${ site.blogId }.${ this.landoDomain }`;
 		}
 	}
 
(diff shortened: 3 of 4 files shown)


src/commands/dev-env-sync-sql.js
```diff
@@ -158,7 +159,10 @@ export class DevEnvSyncSQLCommand {
 			const url = site.homeUrl.replace( /https?:\/\//, '' );
 			if ( ! this.searchReplaceMap[ url ] ) continue;
 
-			this.searchReplaceMap[ url ] = `${ site.blogId }.${ this.landoDomain }`;
+			slugify.extend( { '.': '-' } );
+			this.searchReplaceMap[ url ] = `${ site.blogId }-${ slugify( url, {
```
t-wright: Do we need the blog id as well? Or can we just use the slugified url?

package.json
```diff
@@ -163,6 +163,7 @@
     "semver": "7.5.4",
     "shelljs": "^0.8.5",
     "single-line-log": "1.1.2",
+    "slugify": "^1.6.6",
```
saroshaga: Do we _have_ to introduce this?

src/commands/dev-env-sync-sql.js
hanifn: Having the blog ID _may be_ useful if you have a lot of similar domains for your subsites but it is cleaner without it 🤔

package.json
hanifn: We don't _have_ to but it really does cover all the bases for slugification. I.e..: it has character replacement lists for 

Technically we can code our own slugify algorithm

saroshaga: > Technically we can code our own slugify algorithm

would definitely reduce cognitive load

saroshaga: > We are tempted to not reinvent the wheel so strong that we are ready to import large, heavy libraries to use a small function that we could easily write by ourselves. It introduces unnecessary dependencies and bloated code. Make informed decisions about when to import external libraries and when it is more appropriate to write concise, self-contained code snippets to accomplish smaller tasks.

https://github.com/zakirullin/cognitive-load#abusing-dry-principle

hanifn: I understand this reasoning but the reason I decided to use this dependency anyway is because it's [not a large or bloated library](https://github.com/simov/slugify/blob/master/slugify.js). Of course, we can be super specific in this case and make a simpler(smaller) logic specifically to slugify domain names but in that case, it can't be used elsewhere

saroshaga: > I understand this reasoning but the reason I decided to use this dependency anyway is because it's not a large or bloated library.

I think one of the dangers of still using this is that libraries do tend to get bloated or [riskier](https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/) in the future, so it becomes more about future proofing in that case. I just feel like this is a small enough task to code in ourselves - but feel free to ignore if its going to be heavy lift.

src/commands/dev-env-sync-sql.js
saroshaga: The feedback that we originally received was that the customer didn't have knowledge about subsite IDs. 

Maybe let's move the subsite Id to a later part of the slug as @abdullah-kasim suggests [here](https://vipjira.atlassian.net/browse/FORNO-1631?focusedCommentId=59501) ? It might make the slug easier to search

abdullah-kasim: The blog ID is extremely useful. Immediate, direct knowledge of which set of tables that the developer would have to deal with, versus trying to figure out this slug corresponds to which blog id. Won't matter too much if you're a solo developer and has great memory, but it works extra well in a team.

Think of it like us putting links on our PR that references back to the P2 or the JIRA ticket. It's helpful in that same manner.

But yeah, my initial suggestion is indeed to put the subsite id at the later part for reasons that @saroshaga stated. I know I said otherwise during our google call, but that's because I couldn't recall my exact solution.

hanifn: Yeah agree. Putting the blog ID at the end would make it easier to read.
Changed in 31e851087dd4b2544676f281cb56ddc365e0baa0

package.json
abdullah-kasim: > We are tempted to not reinvent the wheel so strong that we are ready to import large, heavy libraries to use a small function that we could easily write by ourselves. It introduces unnecessary dependencies and bloated code. **Make informed decisions about when to import external libraries** and when it is more appropriate to write concise, self-contained code snippets to accomplish smaller tasks.

I bolded the part where I think we should take note of: Is importing this library an informed decision? Is not importing the library also an informed decision?

> [riskier](https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/) in the future,

The problem linked is easily mitigated by having a lockfile i.e. `yarn.lock, npm-shrinkwrap.json, package-lock.json` - each package version includes an integrity hash to prevent hackers from swapping the package with something else.

> I just feel like this is a small enough task to code in ourselves

There seems to be a lot of [edge cases](https://github.com/simov/slugify/blob/master/test/slugify.js). The tests is about 5 times larger than the implementation itself. I think this is a big task. It's [MIT-licensed though](https://github.com/simov/slugify/blob/master/LICENSE) so if, even after pinning the version, we think it's too much of a risk, we could warehouse it (i.e. copy-paste the code into this codebase).

That said, since I mentioned pinning, I think we need to pin it to `1.6.6` instead of `^1.6.6` to avoid any accidental updates.

hanifn: Considering that the use case here is pretty specific, and I can't think of any other feature that would need slugs created, I think using a local function is better.
Changed in be09ed2e30fbabcfd4f54240a1aac360899a2bdc

saroshaga: > There seems to be a lot of [edge cases](https://github.com/simov/slugify/blob/master/test/slugify.js).

yeah, but do we need ALL of this functionality ?

__tests__/commands/dev-env-sync-sql.js
```diff
@@ -93,6 +105,18 @@ describe( 'commands/DevEnvSyncSQLCommand', () => {
 
 			expect( cmd.searchReplaceMap ).toEqual( { 'test.go-vip.com': 'test-slug.vipdev.lndo.site' } );
 		} );
+
+		it( 'should return a map of search-replace values for multisite', () => {
+			const cmd = new DevEnvSyncSQLCommand( app, msEnv, 'test-slug' );
+			cmd.slug = 'test-slug';
+			cmd.siteUrls = [ 'test.go-vip.com', 'subsite.com' ];
```
abdullah-kasim, line 112: Should we add tests for subsites that doesn't have a domain?

src/commands/dev-env-sync-sql.js
```diff
@@ -158,10 +158,23 @@ export class DevEnvSyncSQLCommand {
...
 		}
 	}
 
+	slugifyDomain( domain ) {
+		return String( domain )
+			.normalize( 'NFKD' ) // split accented characters into their base characters and diacritical marks
+			.replace( /[\u0300-\u036f]/g, '' ) // remove all the accents, which happen to be all in the \u03xx UNICODE block.
+			.trim() // trim leading or trailing whitespace
+			.toLowerCase() // convert to lowercase
+			.replace( /[^a-z0-9 .-]/g, '' ) // remove non-alphanumeric characters except for spaces, dots, and hyphens
```
saroshaga, line 173: is there an escape character for ` ` ?

package.json
hanifn: > > There seems to be a lot of [edge cases](https://github.com/simov/slugify/blob/master/test/slugify.js).
> 
> yeah, but do we need ALL of this functionality ?

Yeah this use case is pretty specifically to slugify an existing URL into another URL. I don't think many of these edge cases applies.

abdullah-kasim: > yeah, but do we need ALL of this functionality ?

Haha, hopefully not! Good point!

Honestly though, I frequently see people get burned out for re-inventing the wheel (i.e. [this PR](https://github.com/Automattic/vip-cli/pull/1237) since we roll our own CLI wrapper) rather than not so I tend to automatically side on not re-inventing the wheel unless we have a good reason not to.

__tests__/commands/dev-env-sync-sql.js
hanifn: Would there ever be a case for subsite to not have a domain? The `wp_blogs` [table's](https://codex.wordpress.org/Database_Description#Table:_wp_blogs) `domain` column is not nullable so I'm not sure if that will ever occur 🤔

src/commands/dev-env-sync-sql.js
hanifn: Sorry @saroshaga I'm not sure what character is that 😅

__tests__/commands/dev-env-sync-sql.js
abdullah-kasim: [Here](https://wordpress.org/documentation/article/before-you-create-a-network/#path-based)

Thinking back though, they could just go to `test-slug.vipdev.lngdo.site/subsite` to access it in that case, actually. No action needed then!

src/commands/dev-env-sync-sql.js
saroshaga: yeah I'm not sure there's one just for single space, [`\s`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape) will detect all whitespaces

saroshaga (approved):

PR #1537, opened 2023-10-30
Add vip dev-env purge command logic

Michelleeby:
## Description

Hello 👋🏼 here is an example of how we might add a `purge` command to VIP CLI `dev-env`. It leverages `getAllEnvironmentNames()` and loops through and awaits `destroyEnvironment` on each name listed.

## Steps to Test

This will need to be plugged into typical testing workflows. I did `npm link` and also `node --inspect` locally and it successfully destroyed all local `dev-env` environments.

### DIFF
--- a/src/bin/vip-dev-env-purge.js
+++ b/src/bin/vip-dev-env-purge.js
@@ -74,8 +74,7 @@ command()
 		const removeFiles = ! ( opt.soft || false );
 
 		try {
-			for ( const envName of allEnvNames ) {
-				const slug = envName;
+			for ( const slug of allEnvNames ) {
 
 				try {
 					// eslint-disable-next-line no-await-in-loop
--- a/src/bin/vip-dev-env-purge.js
+++ b/src/bin/vip-dev-env-purge.js
@@ -1,0 +1,68 @@
+#!/usr/bin/env node
+
+/**
+ * External dependencies
+ */
+import debugLib from 'debug';
+import chalk from 'chalk';
+
+/**
+ * Internal dependencies
+ */
+import { trackEvent } from '../lib/tracker';
+import command from '../lib/cli/command';
+import { destroyEnvironment, getAllEnvironmentNames } from '../lib/dev-environment/dev-environment-core';
+import { DEV_ENVIRONMENT_FULL_COMMAND } from '../lib/constants/dev-environment';
+import {
+	getEnvTrackingInfo,
+	handleCLIException,
+	validateDependencies,
+} from '../lib/dev-environment/dev-environment-cli';
+import { bootstrapLando } from '../lib/dev-environment/dev-environment-lando';
+
+const debug = debugLib( '@automattic/vip:bin:dev-environment' );
+
+const examples = [
+	{
+		usage: `${ DEV_ENVIRONMENT_FULL_COMMAND } purge`,
+		description: 'Destroys all local dev environments',
+	},
+];
+
+command()
+	.option( 'soft', 'Keep config files needed to start an environment intact' )
+	.examples( examples )
+	.argv( process.argv, async ( arg, opt ) => {
+		const allEnvNames = getAllEnvironmentNames();
+        const trackingInfo = { all: true };
+        await trackEvent( 'dev_env_purge_command_execute', trackingInfo );
+
+        try {
+            for ( const envName of allEnvNames ) {
+                const slug = envName;
+                const lando = await bootstrapLando();
+                await validateDependencies( lando, slug );
+    
+                const trackingInfo = getEnvTrackingInfo( slug );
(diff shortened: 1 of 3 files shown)


src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,96 @@
...
+			if ( ! purge ) {
+				return;
+			}
+		}
+
+		try {
+			for ( const envName of allEnvNames ) {
+				const slug = envName;
+				// eslint-disable-next-line no-await-in-loop
+				await validateDependencies( lando, slug );
```
sjinks: I think we can move this out of the loop and turn it into

```js
await validateDependencies( lando, '' );
```

We don't need to verify that Docker is available for every environment.

Because we are destroying all environments, we don't care whether the DNS works; therefore, we can skip that check.

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,96 @@
...
+				const trackingInfoChild = getEnvTrackingInfo( slug );
+				// eslint-disable-next-line no-await-in-loop
+				await trackEvent( 'dev_env_destroy_command_execute', trackingInfoChild );
```
sjinks: Are we tracking the right thing here? I would say we need to track the `dev_env_purge_command_execute` (and move it out of the loop).

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,96 @@
...
+
+				debug( 'Args: ', arg, 'Options: ', opt );
```
sjinks: This statement does not seem to depend upon the loop variables; I think we can move it out of the loop.

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,96 @@
...
+
+				try {
+					const removeFiles = ! ( opt.soft || false );
```
sjinks: Same - `removeFiles` will be the same for all environments; we can evaluate it once before entering the loop.

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env node
```
sjinks, line 1: Please make sure that this file has the execute bit set; otherwise, it will not work in Linux (GitHub does not show file attributes or maybe I cannot find them).

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,98 @@
...
+			if ( ! purge ) {
+				return;
+			}
+		}
+
+		const trackingInfoChild = getEnvTrackingInfo( slug );
+		// eslint-disable-next-line no-await-in-loop
+		await trackEvent( 'dev_env_purge_command_execute', trackingInfoChild );
+
+		// eslint-disable-next-line no-await-in-loop
```
sjinks:
```suggestion
		// eslint-disable-next-line no-await-in-loop
```

We are no longer in the loop; this suppression can be safely removed

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,98 @@
...
+			);
+
+			if ( ! purge ) {
+				return;
+			}
+		}
+
+		const trackingInfoChild = getEnvTrackingInfo( slug );
+		// eslint-disable-next-line no-await-in-loop
+		await trackEvent( 'dev_env_purge_command_execute', trackingInfoChild );
```
sjinks: I think `slug` is not available here :-(

It looks like we already track this event on line 50.

```suggestion
```

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,98 @@
...
+
+		// eslint-disable-next-line no-await-in-loop
+		await validateDependencies( lando, '' );
+		const removeFiles = ! ( opt.soft || false );
+
+		try {
+			for ( const envName of allEnvNames ) {
+				const slug = envName;
```
sjinks:
```suggestion
			for ( const slug of allEnvNames ) {
```

src/bin/vip-dev-env-purge.js
```diff
@@ -0,0 +1,97 @@
...
+
+				try {
+					// eslint-disable-next-line no-await-in-loop
+					await destroyEnvironment( lando, slug, removeFiles );
+
+					const message = chalk.green( '✓' ) + ' Environment destroyed.\n';
+					console.log( message );
+				} catch ( error ) {
+					// eslint-disable-next-line no-await-in-loop
+					await handleCLIException( error, 'dev_env_purge_command_error', trackingInfoChild );
```
sjinks, line 87: It looks like we need to add `trackingInfoChild` back to the `catch` clause

sjinks (approved):

PR #1544, opened 2023-11-06
chore(deps): update update-notifier to 7.0.0

sjinks:
## Description

This PR updates `update-notifier` to 7.0.0. Because `update-notifier` is now an ESM, this required several additional steps.

1. `update-notifier` is imported via `await import` - that's the only way to import ESM from CJS
2. Because `babel` tries to replace `await import` with `Promise.resolve().then(() => _interopRequireWildcard(require('update-notifier'))`, I had to disable the `@babel/plugin-proposal-dynamic-import` plugin.
3. This broke `proxy-agent` tests because `jest` did not like `await import` and insisted on the ESM-enabled environment. Luckily, the test works without dynamic imports.
4. I chose to disable `update-notifier` when `NODE_ENV=test` because I did not want to install `cross-env` and pass `NODE_OPTIONS=--experimental-vm-modules` to keep Jest happy. However, we may need to do this someday.
5. One day, we will need to switch to ESM. However, I don't feel adventurous enough to do that now :-)

As a bonus, we have four less medium security vulnerabilities.

## Steps to Test

1. CI should pass - my changes should not break the existing tests
2. `npm run build; ./dist/bin/vip.js` should work (it will invoke `update-notifier`; there should be no errors coming from Node.js). E2E tests will also catch an error here.

### DIFF
--- a/__tests__/lib/http/proxy-agent.js
+++ b/__tests__/lib/http/proxy-agent.js
@@ -1,9 +1,17 @@
 import { HttpsProxyAgent } from 'https-proxy-agent';
 import { SocksProxyAgent } from 'socks-proxy-agent';
 
+import { createProxyAgent } from '../../../src/lib/http/proxy-agent';
+
 // Reference for testing with env variables within a test: https://github.com/vuejs/vue-test-utils/issues/193
 
 describe( 'validate CreateProxyAgent', () => {
+	let env;
+
+	beforeAll( () => {
+		env = { ...process.env };
+	} );
+
 	beforeEach( () => {
 		// Clear all applicable environment variables before running test so each test starts "clean"
 		// using beforeEach instead of afterEach in case the client running tests has env variables set before the first test is run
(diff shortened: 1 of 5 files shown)


__tests__/lib/http/proxy-agent.js
```diff
@@ -90,11 +102,7 @@ describe( 'validate CreateProxyAgent', () => {
 			urlToHit: 'https://wpAPI.org/api',
 		},
 	] )( 'should return null with %o', async ( { envVars, urlToHit } ) => {
-		setEnvironmentVariabeles( envVars );
-		// We have to dynamically import the module so we can set environment variables above
-		// All tests must be async to support this dynamic import, otherwise the modified env variables are not picked up
-		const createProxyAgent = ( await import( '../../../src/lib/http/proxy-agent' ) )
-			.createProxyAgent;
```
sjinks, line 97: It magically works :-) I found no code in `createProxyAgent` to cache the old environment variables. And, as far as I understand, multiple `await import` (= `require()`) will load the module only once anyway.

__tests__/lib/http/proxy-agent.js
```diff
@@ -19,6 +27,10 @@ describe( 'validate CreateProxyAgent', () => {
 		}
 	} );
 
+	afterAll( () => {
+		process.env = { ...env };
+	} );
```
sjinks, line 32: Maybe this is unnecessary, but I prefer restoring things to their original state.

abdullah-kasim (approved):

PR #1550, opened 2023-11-07
chore(deps): replace cli-table with cli-table3

sjinks:
## Description

`cli-table` is abandoned and no longer maintained. `cli-table3` is the most current replacement.

`cli-table` is the last dependency `npm` complains about:

```
npm WARN skipping integrity check for git dependency ssh://git@github.com/automattic/cli-table.git
```

If you find that very commit, https://github.com/Automattic/cli-table/commit/7b14232ba779929e1859b267bf753c150d90a618, GitHub will complain that it does not belong to any branch on this repository:

Screenshot_20231107_143818

That does not look nice. `cli-table3` seems to be a safer solution.

Among other things, our dependencies use `cli-table3` as well; by switching to `cli-table3`, we remove at least one dependency.

## Steps to Test

Run something like `vip config software get` before and after and observer that the output look the same.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "@automattic/vip",
-  "version": "2.36.0",
+  "version": "2.36.1-dev.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


abdullah-kasim (approved):
Looks good to me!

PR #1557, opened 2023-11-10
fix(dev-env): Fix an issue where dev-env-import-sql command execute event was not being tracked

aswasif007:
## Description

Here, we are fixing a issue where we are not tracking the dev-env-import-sql command execution events.

### DIFF
--- a/src/bin/vip-dev-env-import-sql.js
+++ b/src/bin/vip-dev-env-import-sql.js
@@ -8,7 +8,7 @@ import {
 	handleCLIException,
 	getEnvironmentName,
 } from '../lib/dev-environment/dev-environment-cli';
-import { trackEvent } from '../lib/tracker';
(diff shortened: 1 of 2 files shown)


src/bin/vip-dev-env-import-sql.js
```diff
@@ -44,12 +44,15 @@ command( {
 		const slug = await getEnvironmentName( opt );
 		const cmd = new DevEnvImportSQLCommand( fileName, opt, slug );
 		const trackingInfo = getEnvTrackingInfo( cmd.slug );
+		const trackerFn = makeCommandTracker( 'dev_env_import_sql', trackingInfo );
+		await trackerFn( 'execute' );
```
aswasif007, line 48: This is the missing event.

saroshaga (approved):
thanks for fixing!

PR #1560, opened 2023-11-17
BYOR: Add `vip app deploy`

rebeccahum:
## Description

Adding `vip app deploy` to VIP-CLI

## Steps to Test

1) In `src/bin/vip-app-deploy.js`, need to comment out for now:
```
await track( 'deploy_app_command_execute' );
		await track( 'deploy_app_command_execute' );
		await gates( app, env, fileMeta );
```
2) `node ./dist/bin/vip-app-deploy.js -a 5190 -e production ~/Desktop/skeleton.zip`


mjangda:
@Crixu How well does `vip deploy app` fit in with the new IA for VIP-CLI? Should this be `vip app deploy` instead?

mjangda:
Similarly, if we were to add a rollback command, would that be `vip app rollback`?

Crixu:
I agree that `vip app deploy` is a better choice and makes sure we're moving in the right direction for the future

rebeccahum:
I've changed it to `app deploy` -- I like having the idea of an `app rollback` as well.

WPprodigy (approved):

PR #1615, opened 2023-12-08
feat(dev-env): add quiet mode for "import sql"

### LINKED ISSUE #1561: `dev-env import sql` requires user input
## Expected/Desired Behavior

Run `vip dev-env import sql` without user input.

## Actual Behavior

Command prompts if ES should be reindexed after import, even when environment was created with `--elasticsearch=false` flag.

1. Run `vip dev-env create` (with or without `--elasticsearch=false` flag
1. Run `vip dev-env import sql [FILE]`
1. Observe that you're prompted to reindex ES.

## Additional notes

I'm working on an automation that uses `vip dev-env` to set up an environment, import a database backup that VIP shipped to our S3 bucket, and run various WP-CLI commands against the environment. The ES-indexing prompt breaks the automation by requiring user input.

sjinks:
## Description

Related: #1561 

This PR adds:
* `--skip-reindex` flag to disable reindexing (like other boolean options, it can be negated to request reindexing: `--no-skip-reindex explicitly`). The default is to reindex if the WP CLI `vip-search` command is available;
* `--quiet` flag to suppress extra output from the script (in addition, in this mode, `--quiet` is passed to WP CLI, and `--silent` is passed to `mysql`).

These options can be combined together.

## Pull request checklist

## New release checklist

## Steps to Test

See #1561

```sh
vip dev-env create --elasticsearch true < /dev/null
vip dev-env start
vip dev-env exec -q -- wp db export - > data.sql
# This command will show more information
vip dev-env import sql data.sql
# This one won't
vip dev-env import sql data.sql -q
# This will suppress reindexing
vip dev-env import sql data.sql --skip-reindex
```

### DIFF
--- a/src/commands/dev-env-import-sql.js
+++ b/src/commands/dev-env-import-sql.js
@@ -3,7 +3,11 @@ import fs from 'fs';
 
 import * as exit from '../lib/cli/exit';
 import { getFileMeta, unzipFile } from '../lib/client-file-uploader';
-import { promptForBoolean, validateDependencies } from '../lib/dev-environment/dev-environment-cli';
+import {
+	processBooleanOption,
+	promptForBoolean,
+	validateDependencies,
+} from '../lib/dev-environment/dev-environment-cli';
 import {
 	getEnvironmentPath,
 	resolveImportPath,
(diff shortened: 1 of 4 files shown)


sjinks:
@yolih `vip dev-env import sql` will have an additional option:

```
-q, --quiet           Suppress prompts.
```

chriszarate:
@sjinks Would a flag that controls ES reindexing (`--reindex-elasticsearch=<yes|NO>`) be more useful? A `--quiet` option removes the choice of whether or not to reindex and might be another blocker to full automation.

Alternatively, we could make reindexing the default and introduce a `--skip-reindex-elasticsearch` option.

sjinks:
@chriszarate I have just added the `--[no-]skip-reindex` option.

So,
1. If `--skip-reindex` (or `--skip-reindex false|0|n|no`) is specified, the index prompt is not shown, and reindexing is not performed.
2. If `--no-skip-reindex` is specified, the index prompt is not shown, and reindexing is performed.
3. If `--quiet` is specified, no questions are asked, and reindexing is performed (the same behavior as when the stdin is not a TTY).

chriszarate:
I'm not quite seeing the value of `--quiet`, especially when it hides a default behavior from the user. In fact, I'm sort of questioning the value of the prompt entirely. Why not have a sensible default behavior (reindex when ES is enabled) and a simple `--skip-reindex` to override it?

> If `--no-skip-reindex` is specified, the index prompt is not shown, and reindexing is performed.

This appears to have been left out, which is a good choice IMO.

sjinks:
> I'm not quite seeing the value of --quiet, especially when it hides a default behavior from the user.

`--quiet` does a bit more than that. It also hides informational messages (such as "Check for Docker installation," "Check for docker-compose installation," etc) so that the script can be safely used with cron. `--quiet` assumes a non-interactive environment and assumes the default behavior. In brief, `--quiet` is the same as `--no-skip-reindex` with informational messages hidden.

chriszarate (approved):
> In brief, --quiet is the same as --no-skip-reindex with informational messages hidden.

I get that, but this behavior is hidden and not apparent from the help text. If I were writing automations using `vip`, I would certainly want a way to suppress interactivity.

But more importantly I would want clear, documented flags that let me completely control the behavior of the command. In other words, any interactive prompt should have a corresponding flag that lets me select all possible choices non-interactively.

I like this resource and its principles:

> Only use prompts or interactive elements if stdin is an interactive terminal (a TTY). This is a pretty reliable way to tell whether you’re piping data into a command or whether it’s being run in a script, in which case a prompt won’t work and you should throw an error telling the user what flag to pass.
>
> https://clig.dev/#interactivity

At any rate, I don't want to stand in the way of overall improvement, so the PR is approved and I'll let you decide the path forward.

src/commands/dev-env-import-sql.js
```diff
@@ -25,9 +28,9 @@ export class DevEnvImportSQLCommand {
 		this.slug = slug;
 	}
 
-	async run( silent = false ) {
+	async run() {
```
aswasif007, line 31: This is gonna be an issue: https://github.com/Automattic/vip-cli/blob/054cd85b5fc9a008f300342787437e7ab1bdd432/src/commands/dev-env-sync-sql.js#L177-L182

Also, I'd prefer keeping the `silent` parameter because it can be useful in re-purposing a command class in multiple places. I think all the command-classes' `run` functions have this, for this purpose.

aswasif007 (changes requested):
`run` function is being called from the dev-env-sync command with silent=true. Likely will break stuffs.

src/commands/dev-env-import-sql.js
sjinks: I have updated `dev-env-sync-sql.js` to pass `quiet: true` in the constructor's parameters.

This is because, with VIP CLI Express, we won't be able to pass arbitrary arguments to `run()`. The correct approach will be to instantiate the command with the necessary parameters.

aswasif007 (approved):
Thanks for addressing the issue!

PR #1626, opened 2023-12-28
fix: Properly propagate quotes

aswasif007:
## Description

Fix an issue with propagating quotes to wp shell. Here is the issue with current implementation:

If you execute `vip @app.env -- wp --foo='bar1 "bar2" "bar3"'` the command that gets executed in the shell is: `vip @app.env -- wp --foo="bar1 "bar2" "bar3""`. This PR fixes that quoting issue.

### DIFF
--- a/__tests__/lib/cli/format.js
+++ b/__tests__/lib/cli/format.js
@@ -36,12 +36,12 @@ describe( 'utils/cli/format', () => {
 				expected: [ '"{\\"json\\":broken json with spaces}"' ],
 			},
 			{
-				input: [ '--foo=\'bar1 "bar2" "bar3"\'' ],
-				expected: [ '--foo=\'bar1 "bar2" "bar3"\'' ],
+				input: [ '--foo=bar1 "bar2" "bar3"' ],
+				expected: [ '--foo="bar1 \\"bar2\\" \\"bar3\\""' ],
 			},
 			{
-				input: [ '--foo', '\'bar1 "bar2" "bar3"\'' ],
-				expected: [ '--foo', '\'bar1 "bar2" "bar3"\'' ],
+				input: [ '--foo', 'bar1 "bar2" "bar3"' ],
+				expected: [ '--foo', '"bar1 \\"bar2\\" \\"bar3\\""' ],
 			},
 		] )( 'should requote args when needed - %o', ( { input, expected } ) => {
 			const result = requoteArgs( input );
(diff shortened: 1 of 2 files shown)


sjinks (approved):

chriszarate:
@aswasif007 I added some failing test cases. Revert if I am misunderstanding the issue. I do wonder if we should fix this issue centrally in Parker rather than parsing and validating in the client.

saroshaga (approved):

src/lib/cli/format.ts
```diff
@@ -134,7 +134,8 @@ export function keyValue( values: Tuple[] ): string {
 export function requoteArgs( args: string[] ): string[] {
 	return args.map( arg => {
 		if ( arg.includes( '--' ) && arg.includes( '=' ) && arg.includes( ' ' ) ) {
-			return arg.replace( /^--([^=]*)=(.*)$/, '--$1="$2"' );
+			const quote = arg.includes( '"' ) ? "'" : '"';
```
aswasif007: @chriszarate Perhaps we should replace all the non-terminal `"` by `\"`? Would that be a better idea?

chriszarate: Maybe. I think I am unclear on what the "good" result is from this function. Maybe you could update the test with additional test cases that show what it is supposed to do? I mostly want to make sure that it doesn't provide a way to inject a command.

__tests__/lib/cli/format.js
```diff
@@ -35,6 +35,14 @@ describe( 'utils/cli/format', () => {
...
 				expected: [ '"{"json":broken json with spaces}"' ],
 			},
+			{
+				input: [ '--foo=\'bar1 "bar2" "bar3"\'' ],
+				expected: [ '--foo=\'bar1 "bar2" "bar3"\'' ],
+			},
+			{
+				input: [ '--foo', '\'bar1 "bar2" "bar3"\'' ],
+				expected: [ '--foo', '\'bar1 "bar2" "bar3"\'' ],
+			},
```
aswasif007: @chriszarate I don't think these two test cases are valid:

If the argument is `--foo=\'bar1 "bar2" "bar3"\'` then it actually (correctly) becomes three different arguments: `--foo='bar1` `bar2` `bar3'`

Same for the other one: `--foo \'bar1 "bar2" "bar3"\'` becomes `--foo` `'bar1` `bar2` `bar3'`

chriszarate: Ok, can you add some valid test cases for the change you're making?

aswasif007: Updated your tests instead: https://github.com/Automattic/vip-cli/pull/1626/commits/5941cca74c157a4b396e035c05b870546da9a6e8

chriszarate (approved):
Excellent, thanks!

PR #1667, opened 2024-01-26
Add event tracking for PMA enable failures

saroshaga:
## Description

Add event tracking for PMA enablement failures

Also corrects log message

### DIFF
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -206,6 +206,7 @@ export class PhpMyAdminCommand {
 				error_message: error.message,
 				stack: error.stack,
 			} );
+			this.stopProgressTracker();
 			exit.withError( 'Failed to enable PhpMyAdmin' );
 		}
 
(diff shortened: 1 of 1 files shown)


src/commands/phpmyadmin.ts
```diff
@@ -168,11 +168,23 @@ export class PhpMyAdminCommand {
...
+				const error = err as Error & {
+					graphQLErrors?: GraphQLFormattedError[];
+				};
+				void this.track( 'error', {
+					error_type: 'enable_pma',
+					error_message: error.message,
+					stack: error.stack,
+				} );
+				exit.withError( 'Failed to enable PhpMyAdmin' );
+			}
```
aswasif007: Let's do the tracking inside the `run` function to follow the same pattern with others?

https://github.com/Automattic/vip-cli/blob/d76d266d30fe17cce9b14258343431ccc47a0648/src/commands/phpmyadmin.ts#L217-L232

aswasif007 (approved):

PR #1782, opened 2024-04-16
Update the slowlogs command to follow the VIP-CLI styleguide

terriann:
## Description

Updated the voice, tense, and format of command descriptions, examples, and usage models for the `vip slowlogs` command to follow the VIP-CLI style guide.

## Pull request checklist

## New release checklist

## Steps to Test

Observe the results of the `--help` display for impacted commands:

- `vip slowlogs`

### DIFF
--- a/src/bin/vip-slowlogs.ts
+++ b/src/bin/vip-slowlogs.ts
@@ -24,6 +24,9 @@ const DEFAULT_POLLING_DELAY_IN_SECONDS = 30;
 const MIN_POLLING_DELAY_IN_SECONDS = 5;
 const MAX_POLLING_DELAY_IN_SECONDS = 300;
 
+const exampleUsage = 'vip @example-app.develop slowlogs';
+const baseUsage = 'vip slowlogs';
+
 export async function getSlowlogs( arg: string[], opt: GetSlowLogsOptions ): Promise< void > {
 	validateInputs( opt.limit, opt.format );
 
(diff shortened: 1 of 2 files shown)


src/bin/vip-slowlogs.ts
```diff
@@ -188,26 +191,30 @@ void command( {
...
-			usage: 'vip @mysite.production slowlogs',
+			description:
+				'Retrieve up to 500 of the most recent entries from the MySQL slow query logs in the default text format.',
+			usage: exampleUsage,
 		},
 		{
-			usage: 'vip @mysite.production slowlogs --limit 100',
-			description: 'Get the most recent 100 slowlog entries',
+			description:
+				'Retrieve the 100 most recent entries from the MySQL slow query logs in the default text format.',
```
yolih: I like your "up to 500" clarification in the default command setting, and I think we should continue with the "up to" clarification even when it's limited to 100. If a customer queries the slow query logs for an environment whose logs are are delightfully quiet, we don't want then to think there is something wrong if only 8 log entries are returned.

src/bin/vip-slowlogs.ts
```diff
@@ -188,26 +191,30 @@ void command( {
 	envContext: true,
 	format: true,
 	module: 'slowlogs',
+	usage: baseUsage,
 } )
 	.option( 'limit', 'The maximum number of log lines', 500 )
```
yolih: Sorry, just looking at this option description....
This uses the term "lines", but I think we should be consistent and use the term "entries" here as well. Unless of course they are not technically interchangeable?

terriann: I don't believe either is more or less accurate. I personally prefer "log entries" because the logs include more than the SQL slow query; they include additional metadata for each entry. Using "log entries" instead of "log lines" also removes any confusion about whether a multiline slow SQL query would return as 1 entry or the number of log lines it generates.

Example with the additional metadata:

```
$ vip @example-app.production slowlogs --format=json
[
	{
		"timestamp": "2024-04-12T11:11:02Z",
		"rowsSent": "1",
		"rowsExamined": "1",
		"queryTime": "0.795678",
		"requestUri": "",
		"query": "SELECT ts FROM _vt.heartbeat WHERE keyspaceShard='exampleapp1234:-';"
	}
]
```

yolih: I have the same logic for preferring "entries". I encourage you to make it so.

terriann:
Let's keep this in a holding pattern till #1785 is merged - just in case we want to adjust any language 🙂

rebeccahum:
@terriann Some merge conflicts need to be resolved

rebeccahum (approved):

PR #1809, opened 2024-05-02
fix(phpmyadmin): Add step to wait for PhpMyAdmin readiness. 

aswasif007:
## Description

After spinning up the necessary pod to run PhpMyAdmin, we have to wait for the LB to be propagated, so that the requests to the PhpMyAdmin link are served the by PhpMyAdmin pod. In this PR, we are adding a step to wait to see the LB routing is happening before we generate an access.

## Pull request checklist

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip db phpmyadmin @<app>.<env>`. This should be run on an environment for which PMA is not enabled yet.
1. Ensure that a browser page opened, and it did not end up with an error.

### DIFF
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -164,7 +164,6 @@ export class PhpMyAdminCommand {
 	async readyToServe(): Promise< boolean > {
 		const url = `https://${ this.env.primaryDomain?.name }/.wpvip/pma/auth`;
 
-		// Specify proxy
 		const resp = await fetch( url, {
 			method: 'GET',
 			redirect: 'manual',
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -189,7 +189,15 @@ export class PhpMyAdminCommand {
 			await enablePhpMyAdmin( this.env.id as number );
 			await pollUntil( this.getStatus.bind( this ), 1000, ( sts: string ) => sts === 'running' );
 		}
+
+		const timeout = setTimeout( () => {
+			this.progressTracker.updateMessage(
+				'Enabling PHPMyAdmin for this environment. This step may take a while. We appreciate your patience.'
+			);
+		}, 30000 );
+
 		await pollUntil( this.readyToServe.bind( this ), 5000 );
+		clearTimeout( timeout );
 	}
 
 	async run( silent = false ): Promise< void > {
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -164,7 +164,7 @@ export class PhpMyAdminCommand {
 	}
 
 	async readyToServe(): Promise< boolean > {
-		const url = `https://${ this.env.primaryDomain?.name }/.wpvip/pma/auth`;
+		const url = `https://${ this.env.primaryDomain?.name }/.wpvip/pma/health`;
 
 		const agent = createProxyAgent( url );
 		const resp = await fetch( url, {
@@ -173,13 +173,16 @@ export class PhpMyAdminCommand {
 			agent: agent ?? undefined,
 		} );
 
-		// If we get 401, the domain is hidden behind a login wall, so we cannot access it
-		if ( resp.status === 401 ) {
-			throw new Error( 'We cannot check if PhpMyAdmin is ready due to a login wall' );
+		if ( resp.status !== 200 ) {
+			return false;
 		}
 
-		const text = await resp.text();
-		return text.startsWith( 'Login link expired' );
+		const { status, error } = ( await resp.json() ) as { status: string; error?: string };
+		if ( status === 'Ok' ) {
+			return true;
+		}
+
+		throw new Error( `Failed to serve phpMyAdmin: ${ error }` );
 	}
 
 	async maybeEnablePhpMyAdmin(): Promise< void > {
@@ -235,7 +238,7 @@ export class PhpMyAdminCommand {
 
 		this.progressTracker.stepRunning( this.steps.PROCESSING );
 		try {
-			await pollUntil( this.readyToServe.bind( this ), 5000, undefined, 300000 );
+			await pollUntil( this.readyToServe.bind( this ), 5000 );
 			this.progressTracker.stepSuccess( this.steps.PROCESSING );
 		} catch ( err ) {
 			const error = err as Error;
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -126,7 +126,7 @@ export class PhpMyAdminCommand {
 	track: CommandTracker;
 	steps = {
 		ENABLE: 'enable',
-		PROCESSING: 'Processing',
+		PROCESSING: 'processing',
 		GENERATE: 'generate',
 	};
 	private progressTracker: ProgressTracker;
--- a/src/commands/phpmyadmin.ts
+++ b/src/commands/phpmyadmin.ts
@@ -124,6 +124,7 @@ export class PhpMyAdminCommand {
 	track: CommandTracker;
 	steps = {
 		ENABLE: 'enable',
+		LB_DELAY: 'lb_delay',
 		GENERATE: 'generate',
 	};
 	private progressTracker: ProgressTracker;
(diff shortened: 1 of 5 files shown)


src/commands/phpmyadmin.ts
```diff
@@ -213,6 +230,16 @@ export class PhpMyAdminCommand {
 			);
 		}
 
+		this.progressTracker.stepRunning( this.steps.LB_DELAY );
+		try {
+			await pollUntil( this.readyToServe.bind( this ), 5000 );
```
aswasif007: Issue: this can get stuck into infinite loop.

src/commands/phpmyadmin.ts
```diff
@@ -134,6 +135,7 @@ export class PhpMyAdminCommand {
 		this.track = trackerFn;
 		this.progressTracker = new ProgressTracker( [
 			{ id: this.steps.ENABLE, name: 'Enabling PHPMyAdmin for this environment' },
+			{ id: this.steps.LB_DELAY, name: 'Waiting for PHPMyAdmin to be ready' },
```
aswasif007:
```suggestion
			{ id: this.steps.LB_DELAY, name: 'Processing' },
```

src/lib/utils.ts
```diff
@@ -11,20 +11,20 @@ const debug = debugLib( '@automattic/vip:lib:utils' );
...
 	interval: number,
-	isDone: ( v: T ) => boolean
+	isDone?: ( v: T ) => boolean
 ): Promise< void > {
 	let done = false;
 	while ( ! done ) {
 		// eslint-disable-next-line no-await-in-loop
 		const result = await fn();
-		done = isDone( result );
+		done = isDone?.( result ) || Boolean( result );
```
saroshaga, line 31: TIL that you could call function and optional chain

src/commands/phpmyadmin.ts
```diff
@@ -134,6 +135,7 @@ export class PhpMyAdminCommand {
 		this.track = trackerFn;
 		this.progressTracker = new ProgressTracker( [
 			{ id: this.steps.ENABLE, name: 'Enabling PHPMyAdmin for this environment' },
+			{ id: this.steps.LB_DELAY, name: 'Processing' },
```
saroshaga:
```suggestion
			{ id: this.steps.PROCESSING, name: 'Processing' },
```

src/commands/phpmyadmin.ts
```diff
@@ -124,6 +124,7 @@ export class PhpMyAdminCommand {
 	track: CommandTracker;
 	steps = {
 		ENABLE: 'enable',
+		LB_DELAY: 'lb_delay',
```
saroshaga:
```suggestion
		PROCESSING: 'processing',
```

src/lib/utils.ts
```diff
@@ -11,24 +11,31 @@ const debug = debugLib( '@automattic/vip:lib:utils' );
...
+ * @param {number}   cutOff   Maximum time to poll in milliseconds
  * @return {Promise}          A promise which resolves when the polling is done
  * @throws {Error}            If the fn throws an error
  */
 export async function pollUntil< T >(
 	fn: () => Promise< T >,
 	interval: number,
-	isDone: ( v: T ) => boolean
+	isDone?: ( v: T ) => boolean,
+	cutOff?: number
```
saroshaga, line 23:
```suggestion
	cutOff: number
```

should this be optional? Also can we rename to `maxAllowedDuration` ?

src/commands/phpmyadmin.ts
```diff
@@ -134,6 +137,7 @@ export class PhpMyAdminCommand {
 		this.track = trackerFn;
 		this.progressTracker = new ProgressTracker( [
 			{ id: this.steps.ENABLE, name: 'Enabling PHPMyAdmin for this environment' },
+			{ id: this.steps.PROCESSING, name: 'Processing' },
```
saroshaga:
```suggestion
			{ id: this.steps.PREPARING, name: 'Preparing' },
```

I think this sounds better 😅

src/commands/phpmyadmin.ts
```diff
@@ -159,14 +163,30 @@ export class PhpMyAdminCommand {
...
+			agent: agent ?? undefined,
+		} );
+
+		// If we get 401, the domain is hidden behind a login wall, so we cannot access it
+		if ( resp.status === 401 ) {
+			throw new Error( 'We cannot check if PhpMyAdmin is ready due to a login wall' );
+		}
+
+		const text = await resp.text();
+		return text.startsWith( 'Login link expired' );
```
saroshaga: So are we saying that if the login link is expired , it is ready to serve?

src/commands/phpmyadmin.ts
```diff
@@ -159,14 +163,30 @@ export class PhpMyAdminCommand {
...
+		const agent = createProxyAgent( url );
+		const resp = await fetch( url, {
+			method: 'GET',
+			redirect: 'manual',
+			agent: agent ?? undefined,
+		} );
+
+		// If we get 401, the domain is hidden behind a login wall, so we cannot access it
+		if ( resp.status === 401 ) {
+			throw new Error( 'We cannot check if PhpMyAdmin is ready due to a login wall' );
```
saroshaga: Won't this bail out of the polling completely?

src/lib/utils.ts
aswasif007: I think it is fine to keep it optional, as it is already being used in other places, and we haven't faced any deadlock on them yet.

aswasif007:
Update: upon an offline discussion with Sarosh, we agreed to create a status/healthcheck endpoint in the [phpMyAdmin](https://github.com/Automattic/vip-phpmyadmin) itself, which will allow us to detect readiness in a cleaner way.

src/commands/phpmyadmin.ts
aswasif007: Yes. We don't have a choice on this if the domain itself is hidden behind a login wall.

src/commands/phpmyadmin.ts
```diff
@@ -124,6 +126,7 @@ export class PhpMyAdminCommand {
 	track: CommandTracker;
 	steps = {
 		ENABLE: 'enable',
+		PREPARING: 'preparing',
```
saroshaga: Actually I've been thinking about this and I feel like maybe we should just keep this in the Enable step, no? I think preparing is ambiguous and isn't in line with the other steps

aswasif007: I thought the same. My only discomfort is that it will be in the enable step for too long 😄

saroshaga: I think that's fine

saroshaga: @aswasif007 : I think you are correct. It is taking a _long_ time. Perhaps we should we should add some help text to the effect of "enabling PMA may take a long time in some situations"

aswasif007: Let's do one thing: if it takes longer than 30 seconds, we update the text on the step to read "it can take a while, please wait"

aswasif007: Or "we appreciate your patience" 🤷

saroshaga: yeah that sounds good.

saroshaga: "this step may take a while"

aswasif007: https://github.com/Automattic/vip-cli/pull/1809/commits/87786ae03f0970a026500bc592eab20966bb0d70

saroshaga (approved):

src/commands/phpmyadmin.ts
```diff
@@ -159,15 +161,43 @@ export class PhpMyAdminCommand {
...
 
+	async readyToServe(): Promise< boolean > {
+		const url = `https://${ this.env.primaryDomain?.name }/.wpvip/pma/health`;
+
+		const agent = createProxyAgent( url );
+		const resp = await fetch( url, {
+			method: 'GET',
+			redirect: 'manual',
+			agent: agent ?? undefined,
+		} );
```
abdullah-kasim, line 172: Should we add some cache-ignoring/cache-busting headers/cookies here? Or switch to `POST` here? It will at least eliminate the possibility that it's not the cache that's causing issues.

PR #1817, opened 2024-05-04
Do custom env validation for custom deploys

WPprodigy:
## Description

Update custom deploy validation. Rather than performing app lookup queries the generic way, we'll handle all that in the custom deployment's token validation method.

## Changelog Description

Update custom deploy validation to run fully off of the custom deploy token. There are breaking changes associated with these updates, so the VIP CLI must be updated to this latest version in order to continue using the custom deploy functionality.

### DIFF
--- a/src/bin/vip-app-deploy.ts
+++ b/src/bin/vip-app-deploy.ts
@@ -11,7 +11,7 @@ import gql from 'graphql-tag';
 /**
  * Internal dependencies
  */
-import { App, AppEnvironment, AppEnvironmentCustomDeployInput } from '../graphqlTypes';
+import { AppEnvironmentCustomDeployInput } from '../graphqlTypes';
 import API from '../lib/api';
 import command from '../lib/cli/command';
 import * as exit from '../lib/cli/exit';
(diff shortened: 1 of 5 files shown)


src/bin/vip-app-deploy.ts
```diff
@@ -108,29 +80,31 @@ export async function promptToContinue( params: PromptToContinueParams ) {
...
+	const fileInputBasename = fileMeta.basename;
 
 	debug( 'Options: ', opts );
 	debug( 'Args: ', arg );
 
-	const appId = env.appId as number;
-	const envId = env.id as number;
+	debug( 'Validating custom deploy key...' );
+	const { appId, envId, ...validatedArgs } = await validateCustomDeployKey( app, env );
+	console.log( { appId, envId, validatedArgs } );
```
rebeccahum: Let's just do `debug()` for this info, I don't think the user needs this info. Can also remove too.

src/lib/custom-deploy/custom-deploy.ts
```diff
@@ -13,47 +13,74 @@ import { validateDeployFileExt, validateFilename } from '../../lib/validations/c
...
+	customDeploysEnabled: boolean;
+};
+
+type ValidateMutationPayload = {
+	data?: {
+		validateCustomDeployAccess: CustomDeployInfo;
+	} | null;
+};
+
 export function isSupportedApp( app: App ): boolean {
```
rebeccahum: We can remove this now I think since it looks like we're only doing the validation on the Parker end?

src/lib/custom-deploy/custom-deploy.ts
```diff
@@ -13,47 +13,74 @@ import { validateDeployFileExt, validateFilename } from '../../lib/validations/c
...
 			mutation: VALIDATE_CUSTOM_DEPLOY_ACCESS_MUTATION,
 			context: {
 				headers: {
 					Authorization: `Bearer ${ WPVIP_DEPLOY_TOKEN }`,
 				},
 			},
 		} );
+
+		if ( ! result.data?.validateCustomDeployAccess ) {
+			throw new Error( 'Not found' );
```
rebeccahum, line 64: Need a better error descriptor

rebeccahum (commented):
This is a better approach, thank youu. Just a few little things:

src/bin/vip-app-deploy.ts
WPprodigy: Will remove. Debugging while I test out all the various ways an app/env can be input 🙃

src/lib/custom-deploy/custom-deploy.ts
WPprodigy: This one is just to trigger the catch block, so the message here never gets passed to the user.

rebeccahum (approved):

PR #1866, opened 2024-06-06
Updating command option descriptions to follow the VIP-CLI style guide

yolih:
Updated the voice, tense, and format for descriptions of several command options to follow the VIP-CLI style guide.

## Pull request checklist

## New release checklist

## Steps to Test

Observe the results of the `--help` display for impacted commands, such as:

- `vip dev-env logs -h`
- `vip dev-env update -h`


src/lib/dev-environment/dev-environment-cli.ts
```diff
@@ -830,43 +830,61 @@ export function processVersionOption( value: unknown ): string {
...
+			processVersionOption
+		)
+		.option(
+			[ 'u', 'mu-plugins' ],
+			'Manage the source for VIP MU plugins. Accepts "demo" (default) for a read-only image of the staging branch, or a path to a built copy of VIP MU plugins on the local machine.'
+		)
 		.option(
 			'app-code',
-			'Use the application code from a local directory or use "demo" for VIP skeleton code'
+			'Manage the source for application code. Accepts "demo" (default) for a read-only image of WordPress skeleton code, or a path to a git cloned application repo on the local machine.'
```
terriann:
```suggestion
			'Manage the source for application code. Accepts "demo" (default) for a read-only image of WordPress VIP skeleton application code, or a path to a VIP formatted application repo on the local machine.'
```
A couple of things I'd suggest on this change:

- We should be clear that it's the VIP skeleton repo since that's a format and repo specific for WordPress on VIP
- The path to the local code doesn't have to be git cloned, but it does have to be in the same/similar format as a VIP customer application repo

src/lib/dev-environment/dev-environment-cli.ts
```diff
@@ -830,43 +830,61 @@ export function processVersionOption( value: unknown ): string {
...
 		)
 		.option(
 			'phpmyadmin',
-			'Enable PHPMyAdmin component. By default it is disabled',
+			'Manage PHPMyAdmin, disabled by default. Accepts "y" (default value) to enable or "n" to disable. When enabled, refer to the value of PHPMYADMIN URLS in the information output for a local environment for the URL to access phpMyAdmin.',
```
terriann:
```suggestion
			'Manage phpMyAdmin, disabled by default. Accepts "y" (default value) to enable or "n" to disable. When enabled, refer to the value of "PHPMYADMIN URLS" in the information output for a local environment for the URL to access phpMyAdmin.',
```

We should stay consistent using the [correct capitalization of phpMyAdmin](https://www.phpmyadmin.net/) where possible, and quote the unexpected casing to highlight that's the string that a user should look for in the info output.

src/lib/dev-environment/dev-environment-cli.ts
```diff
@@ -830,43 +830,61 @@ export function processVersionOption( value: unknown ): string {
...
 			processBooleanOption
 		)
 		.option(
 			[ 'r', 'media-redirect-domain' ],
-			'Domain to redirect for missing media files. This can be used to still have images without the need to import them locally.'
+			'Configure media files to be proxied from a VIP Platform environment. Accepts a string value for the primary domain of the VIP Platform environment or "n" to disable the media proxy.'
+		)
+		.option(
+			'php',
+			'Manage the version of PHP. Accepts a string value for minor versions (8.x). Defaults to the most recent version of PHP.',
```
terriann: Is the value `8.x` accepted, or is "x", in this example, a placeholder for the number for the minor versions like 8.1 and 8.2? If it's the latter, should we be more specific instead of assuming the end user knows to replace x with a number?

Should we explore the possibility of populating this help info with the versions available in the constant `DEV_ENVIRONMENT_PHP_VERSIONS` to keep the help documentation up to date automatically?
https://github.com/Automattic/vip-cli/blob/2bce08bb71c77f37d082600dea9017c9eaa71fbd/src/lib/constants/dev-environment.ts#L36-L46

src/lib/dev-environment/dev-environment-cli.ts
yolih: I appreciate both of these suggestions and will make it so!

src/lib/dev-environment/dev-environment-cli.ts
yolih: Whoops! That's a leftover artifact that I overlooked. Thanks!

terriann (approved):
👍  looks good to me!

src/bin/vip-config-software-update.js
```diff
@@ -54,7 +54,7 @@ const cmd = command( {
 	wildcardCommand: true,
 	usage,
 } ).examples( examples );
-cmd.option( 'yes', 'Auto-confirm update' );
+cmd.option( 'yes', 'Skip the confirmation prompt and automatically submit "y".' );
```
sanmai, line 57:
```suggestion
cmd.option( 'yes', 'Skip the update confirmation prompt and automatically submit "y".' );
```
Feels more explicit and clear for me. Feel free to disregard it.

src/bin/vip-sync.js
```diff
@@ -51,7 +51,7 @@ command( {
 			let bail = false;
 
 			for ( const err of error.graphQLErrors ) {
-				if ( err.message !== 'Site is already syncing' ) {
+				if ( err.message !== 'Site is already syncing.' ) {
```
sanmai: Did we change the error in our services?

sanmai (changes requested):

src/bin/vip-sync.js
yolih: I only edited the error message for punctuation, but did not investigate the underlying mechanisms that might throw the error. To be clear, this syncing action would be from a Platform environment to a local environment in case that makes a difference.

sanmai:
```suggestion
				if ( err.message !== 'Site is already syncing' ) {
```
We are checking against the error message we received from the upstream. So, to change the error message, we need to change it where it was created. I'd like to propose that we leave this change for another PR where we can find where ends meet and fix it all over the place 🙏

yolih: Agreed! Work to change the error message should be handled in a separate PR. But I'm glad that my little PR helped you notice that the error message needed to be fixed!

src/bin/vip-config-software-update.js
yolih: Because the user will be looking at the `-h` menu for the command `vip config software update`, it seems redundant to me to add "update" to the description for the `--yes` option. Would it be acceptable to commit my version of the description as-is and we can add the clarifier later if we think it's needed?

sanmai: Sounds good!

src/lib/dev-environment/dev-environment-cli.ts
```diff
@@ -858,48 +858,68 @@ export function processVersionOption( value: unknown ): string {
...
 			[ 'r', 'media-redirect-domain' ],
-			'Domain to redirect for missing media files. This can be used to still have images without the need to import them locally.',
+			'Configure media files to be proxied from a VIP Platform environment. Accepts a string value for the primary domain of the VIP Platform environment or "n" to disable the media proxy.',
 			undefined,
 			processMediaRedirectDomainOption
 		)
-		.option( 'php', 'Explicitly choose PHP version to use', undefined, processVersionOption )
+		.option(
+			'php',
+			`Manage the version of PHP. Accepts a string value for minor versions: ${ phpVersionsSupported }`,
```
sanmai, line 910: Tested with `./dist/bin/vip-dev-env-update.js -h`

sanmai (approved):
I did another scan and ran some tests. Let's ship it 👍

PR #1962, opened 2024-07-26
Mydumper integration

abdullah-kasim:
## Description

This PR adds support for mydumper-based SQL dumps for almost all commands that interact with SQL dumps.

Those commands are:
* `vip dev-env sync sql`
* `vip dev-env import sql`
* `vip search-replace`

What's currently not supported is:
* `vip import sql` - the post-processing involved is included in this PR, but this also requires a back-end change so that  the endpoint it's communicating with can read a `mydumper` file. For now, it's best to use the [datasync feature](https://docs.wpvip.com/databases/data-sync/).

## Pull request checklist

## New release checklist

## Steps to Test

For testing `vip dev-env sync`:

1. Check out PR.
2. Run `npm run build:watch`
3. Ensure that you have a site ready that's part of the mydumper testing group. See `p6jPRI-8tH` shortlink on how to make a site be part of one.
4. Create a new dev-env by running `./dist/bin/vip-dev-env-create.js @0000.production --slug=0000-production` where `0000` is your testing site id. I also suggest that you add phpmyadmin for validation purposes
5. Then, start the dev environment - `./dist/bin/vip-dev-env-start.js @0000.production --slug=0000-production`
6. Run `./dist/bin/vip-dev-env-sync-sql.js @0000.production --slug=0000-production` and your local dev environment should now have its database ready - with all the search and replace working.

For testing `vip dev-env import sql`:

1. Do steps 1 to 5 of testing `vip dev-env sync`.
2. Run `./dist/bin/vip-dev-env-import-sql.js <path-to-mydumper-sql-file> --slug=0000-production`
3. You should see warnings about running search-replace, and the import will partially succeed with failures at creating the admin user.
4. Re-run `./dist/bin/vip-dev-env-import-sql.js <path-to-mydumper-sql-file> --slug=0000-production --search-replace="from,to"` as per its recommendation.
5. Database should import properly this time and the dev-env at the slug `0000-production` will work, just like a mysqldump-based file.

For testing `vip search-replace`:

1. Do steps 1 to 5 of testing `vip dev-env sync`
2. Run `./dist/bin/vip-search-replace.js <mydumper-file> --search-replace="some.from.domain,some.to.domain" --output="result.sql"`. The resulting file should have metadata headers i.e. `-- some_db_wp_site.00000.sql -1` with a -1 at the back. It will also properly replace the values listed in search-replace.


src/lib/database.ts
```diff
@@ -0,0 +1,130 @@
+import fs from 'fs';
+import { Readable } from 'node:stream';
+import { promisify } from 'node:util';
+import readline from 'readline';
+import yauzl, { Entry } from 'yauzl';
```
sjinks: We already have `adm-zip` in the project. To avoid an extra dependency, could we please use it here as well?

src/lib/database.ts
```diff
@@ -0,0 +1,130 @@
+import fs from 'fs';
```
sjinks:
```suggestion
import fs from 'node:fs';
```

for consistency

src/lib/database.ts
```diff
@@ -0,0 +1,130 @@
...
+import { Readable } from 'node:stream';
+import { promisify } from 'node:util';
+import readline from 'readline';
```
sjinks:
```suggestion
import readline from 'node:readline';
```

Same :-)

src/lib/database.ts
abdullah-kasim: 👍 I'm removing yauzl but not for this reason.

I realized that I don't think we'll have `.zip` files frequently - so it doesn't make sense for us to suddenly support `.zip`.

FWIW - `adm-zip` can't do stream extract so I can't read just 5 lines of the top of the file - an important need of ours if the sql files is like, any bigger than 2 GB (as we'll run out of memory). Good for our current purposes, but we might be limited by it in the future.

luismulinari:
This is awesome!

src/commands/dev-env-import-sql.ts
```diff
@@ -78,15 +88,14 @@ export class DevEnvImportSQLCommand {
...
+				skipChecks: isMyDumper ? [ 'dropTable', 'dropDB' ] : [],
 				extraCheckParams: { siteHomeUrlLando: expectedDomain },
 			} );
 		}
 
 		const fd = await fs.promises.open( resolvedPath, 'r' );
-		const importArg = [ 'db', '--disable-auto-rehash' ].concat(
-			this.options.quiet ? '--silent' : []
-		);
+		const importArg = this.getImportArgs( dumpDetails );
```
abdullah-kasim, line 97: Extracted to a function to satisfy `eslint` complexity limit - which I totally agree with.

src/commands/dev-env-import-sql.ts
```diff
@@ -111,38 +120,44 @@ export class DevEnvImportSQLCommand {
 			fs.unlinkSync( resolvedPath );
 		}
 
-		const cacheArg = [ 'wp', 'cache', 'flush', '--skip-plugins', '--skip-themes' ].concat(
-			this.options.quiet ? '--quiet' : []
-		);
-		await exec( lando, this.slug, cacheArg );
+		await flushCache( lando, this.slug, this.options.quiet );
```
abdullah-kasim: Also more refactor by extraction here so that I could read this code more easily to understand what's going on.

__tests__/commands/dev-env-sync-sql.ts
```diff
@@ -1,74 +1,24 @@
...
-import { getReadInterface } from '../../src/lib/validations/line-by-line';
-
-/**
- *
- * @param {Array<{name, data}>} eventArgs Event arguments
- * @param {timeout}             timeout
- *
- * @return {Stream} A passthrough stream
- */
-function getMockStream( eventArgs: { name: string; data?: string }[], timeout = 10 ) {
```
abdullah-kasim, line 21: I had to refactor the tests by quite a bit - it's mocking too much and adjusting the mock is starting to take too much time. Instead, I don't mock it but instead, provide a small fixture that is also speedy.

src/commands/dev-env-sync-sql.ts
```diff
@@ -87,14 +90,27 @@ export class DevEnvSyncSQLCommand {
 		return `${ this.slug }.${ this.lando.config.domain }`;
 	}
 
-	private get sqlFile(): string {
+	public get sqlFile(): string {
 		return `${ this.tmpDir }/sql-export.sql`;
 	}
 
-	private get gzFile(): string {
+	public get gzFile(): string {
```
abdullah-kasim, line 104: Made public to make tests easier.

src/lib/database.ts
```diff
@@ -0,0 +1,131 @@
...
+	MYDUMPER = 'MYDUMPER',
+	MYSQLDUMP = 'MYSQLDUMP',
+}
+
+export interface SqlDumpDetails {
+	type: SqlDumpType;
+	sourceDb: string;
+}
+
+export const getSqlDumpDetails = async ( filePath: string ): Promise< SqlDumpDetails > => {
```
abdullah-kasim, line 18: This function figures out the type of sql dump that we're dealing with (either mydumper or mysqldump), and the source DB if it's a mydumper dump.

src/lib/database.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const getSqlFileStreamFromCompressedFile = async ( filePath: string ): Promise< zlib.Gunzip > => {
+	if ( filePath.endsWith( '.gz' ) ) {
+		return await getSqlFileStreamFromGz( filePath );
+	}
+
+	throw new Error( 'Not a supported compressed file' );
+};
+
+export const fixMyDumperTransform = () => {
```
abdullah-kasim, line 109: This transform is a stream 'middleware' that modifies the metadata header in mydumper i.e. `-- some_db.tbl 12345`, and then replaces the `12345` with `-1`. This allows us to bypass the metadata bytes validation.

src/commands/dev-env-import-sql.ts
```diff
@@ -78,15 +88,14 @@ export class DevEnvImportSQLCommand {
 			const expectedDomain = `${ this.slug }.${ lando.config.domain }`;
 			await validateSQL( resolvedPath, {
 				isImport: false,
-				skipChecks: [],
+				skipChecks: isMyDumper ? [ 'dropTable', 'dropDB' ] : [],
```
abdullah-kasim, line 91: `dropTable` and `dropDB` validations doesn't make sense in the mydumper context because `isMyDumper` uses `--overwrite-table` to perform any table/db dropping - hence we're ignoring it here.

src/commands/dev-env-sync-sql.ts
```diff
@@ -23,7 +25,7 @@ import { getReadInterface } from '../lib/validations/line-by-line';
  * @return Site home url. null if not found
  */
 function findSiteHomeUrl( sql: string ): string | null {
-	const regex = "'(siteurl|home)',\\s?'(.*?)'";
+	const regex = `['"](siteurl|home)['"],\\s?['"](.*?)['"]`;
```
abdullah-kasim, line 28: Minor regex change because mydumper uses double quotes while mysqldump uses single quote.

Using two regexes may be more accurate, but this change in the regex seems to be good enough from my testing.

abdullah-kasim (commented):
Most of the changes are admittedly the refactor.

The most important logic that has changed is the addition of the `fixMyDumperTransform()` to fix/workaround a mydumper dump's stream when it's passed over to the `replace()` stream, as well as the detection logic for checking if a file is a mysqldump dump or a mydumper dump.

src/commands/dev-env-import-sql.ts
```diff
@@ -111,38 +120,44 @@ export class DevEnvImportSQLCommand {
...
+				'--skip-triggers',
+				'--skip-post',
+				'--innodb-optimize-keys',
+				'--checksum=SKIP',
+				'--metadata-refresh-interval=2000000',
+				'--stream',
+			].concat( this.options.quiet ? [ '--verbose=0' ] : [ '--verbose=3' ] );
+		}
+
+		return importArg;
```
sanmai, line 161: That's a nice refactor! I bet it would be easier to review if you package just the refactor 👍

abdullah-kasim: I'll separate it out then :)

src/lib/database.ts
```diff
@@ -0,0 +1,131 @@
...
+		type: isMyDumper ? SqlDumpType.MYDUMPER : SqlDumpType.MYSQLDUMP,
+		sourceDb: sourceDB,
+	};
+};
+
+const verifyFileExists = async ( filePath: string ) => {
+	try {
+		await fs.promises.access( filePath, fs.constants.F_OK );
+	} catch {
+		throw new Error( 'File not accessible. Does file exists?' );
```
sanmai:
```suggestion
		throw new Error( 'File not accessible. Does file exist?' );
```

src/commands/dev-env-import-sql.ts
sanmai: Please tag me there when you do 💯

abdullah-kasim: #1962

abdullah-kasim:
For future ref, here's some minimal performance tests that I've run with importing a 3 GB dump:

```sh
# vip dev-env import - mydumper - 10 SRs with wp search-replace
real    20m55.033s
user    0m31.632s
sys     0m5.237s

# vip dev-env import mydumper - 0 SRs
real    3m55.207s
user    0m30.658s
sys     0m6.472s


# 17 minutes for 10 SRs
# 1.7 minutes per SR with wp search-replace


# vip dev-env import - mysqldump - 0 SRs:
real    10m15.278s
user    0m19.742s
sys     0m6.805s

# mysqldump - no streaming by copying file into docker container.
real    10m4.920s
user    0m11.425s
sys     0m2.730s

# vip search-replace - 10 SRs with go-search-replace.
real    0m32.049s
user    0m33.223s
sys     0m12.168s


# winner: go-search-replace with mydumper.
```

src/lib/validations/sql.ts
```diff
@@ -335,7 +341,7 @@ const checks: Checks = {
 		recommendation: '',
 	},
 	siteHomeUrlLando: {
-		matcher: "'(siteurl|home)',\\s?'([^']+)'",
+		matcher: `['"](siteurl|home)['"],\\s?['"]([^'"]+)['"]`,
```
sanmai, line 344:
```suggestion
		matcher: `['"](siteurl|home)['"],\\s?(?:['"])([^'"]+)\1`,
```
I wonder if we could make use of back-references here to make the regex more reliable. 

(I haven't tested this regex. )

abdullah-kasim: It didn't work - the following test value is supposed to return `true`.

```js
const testValue = 'INSERT INTO `wp_options` VALUES(1,"siteurl","https://000.someblog.test","yes")'

// PR's regex
const regexO2 = /['"](siteurl|home)['"],\s?['"]([^'"]+)['"]/

// Suggested regex
const regexN2 = /['"](siteurl|home)['"],\s?(?:['"])([^'"]+)\1/

regexO2.test(testValue) // true

regexN2.test(testValue) // false
```

sanmai:
```
const regexN2 = /(['"])(siteurl|home)\1,\s?\1([^'"]+)\1/

regexN2.test(testValue) // true
```
Although I agree it introduces another capture group, so I'm leaving it up to you.

sanmai (approved):
Followed the testing steps, no issues 👍

src/lib/validations/sql.ts
abdullah-kasim: Ah okay, I get it. Took a good look at what it does - it utilizes the type of quote that first matches.

Though thinking hard about it - I'll need to wait until we see what the edge case looks like.

I'm thinking of a scenario where there'll be both an escaped single and double quote - which even then, both regexes will fail - so might be worth deferring until we see it hitting an edge case.

PR #1970, opened 2024-07-29
fix(validate-files): Update file-validation logic to add mimetype validation

aswasif007:
## Description

In the `vip import validate-files` command, we currently do not parse and check the mimetype of the file itself. In this PR, we are adding that logic. 

## Pull request checklist

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-import-validate-files.js <your-dir-path>`
1. Check the verification messages to see if they are correct.

### DIFF
--- a/src/lib/validations/utils.ts
+++ b/src/lib/validations/utils.ts
@@ -42,6 +42,13 @@ export function getMultilineStatement( statementRegex: RegExp ): ( line: string 
 }
 
 export function getFileType( filepath: string ): string {
+	// Check if the computer has the file command installed
+	try {
+		cp.execFileSync( '/usr/bin/file', [ '-v' ] );
+	} catch ( err ) {
+		throw new Error( '/usr/bin/file does not exist this computer' );
+	}
+
 	const regex = new RegExp( /^\w+\/[-+.\w]+/g );
 
 	let result;
--- a/src/lib/validations/utils.ts
+++ b/src/lib/validations/utils.ts
@@ -1,3 +1,5 @@
+import { execFileSync } from 'child_process';
+
 /**
  * Get SQL statements matching a supplied pattern from a file stream
  *
@@ -34,4 +36,27 @@ export function getMultilineStatement( statementRegex: RegExp ): ( line: string 
 
 		return matchingStatements;
 	};
+}
+
+export function getFileType( filepath: string ): string {
+	const regex = new RegExp( /^\w+\/[-+.\w]+/g );
+
+	let errMsg: string;
+	try {
+		const result = execFileSync( '/usr/bin/file', [ '-b', '--mime-type', filepath ] );
+
+		// execFileSync returns a Buffer so we need to convert it to a string
+		const resultStr = result.toString().trim();
+
+		// If it doesn't look like a MIME type string than it's probably an error
+		if ( ! regex.test( resultStr ) ) {
+			errMsg = `Error encountered while trying to find mime type for ${ filepath }: ${ resultStr }`;
+			throw new Error( errMsg );
+		}
+
+		return resultStr;
+	} catch ( err ) {
+		errMsg = `Failed to run command to find mime type for ${ filepath }: ${ err as string }`;
+		throw new Error( errMsg );
+	}
 }
--- a/src/lib/validations/utils.ts
+++ b/src/lib/validations/utils.ts
@@ -1,4 +1,7 @@
-import { execFileSync } from 'child_process';
+/**
+ * External dependencies
+ */
+import cp from 'child_process';
 
 /**
  * Get SQL statements matching a supplied pattern from a file stream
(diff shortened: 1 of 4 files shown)


src/lib/validations/utils.ts
```diff
@@ -35,3 +40,27 @@ export function getMultilineStatement( statementRegex: RegExp ): ( line: string
...
 	};
 }
+
+export function getFileType( filepath: string ): string {
+	const regex = new RegExp( /^\w+\/[-+.\w]+/g );
+
+	let result;
+
+	try {
+		result = cp.execFileSync( '/usr/bin/file', [ '-b', '--mime-type', filepath ] );
```
aswasif007, line 57: ⚠️ I am not sure about the windows support for this ⚠️

saroshaga: I don't see a Windows native way to do this either. The best bet is to rely on a third party program [`GnuWin32's file`](https://gnuwin32.sourceforge.net/packages/file.htm)

saroshaga: For now, could we conditionally run this if the file command exists?

saroshaga: In this way we could even support GnuWin32 file based on the OS platform.

sjinks: I am not even sure about proper Linux support :-) On a typical Linux desktop environment, `file` is not installed by default.

We need to check if `file` exists and not bail out if it does not or cannot be executed.

rinatkhaziev (changes requested):
Needs more thinking around portability/x-platform support. I.e. Windows native/Windows WSL/Linux.

There are a few npm packages that are doing exactly the logic we're trying to do here, have we evaluated them?

src/lib/validations/utils.ts
aswasif007: Added a check to see if the command exists on computer. If not, the mimetype check is bypassed.

rinatkhaziev:
So I tested the latest version under Windows native, and WSL on Ubuntu. 

On Windows it repeatedly presents the error message `"Failed to extract mimetype of file %FILEPATH% because of error: /usr/bin/file does not exist on this computer"`, I think it's a bit excessive, can we group the errors somehow? A Linux/Mac path under Windows also doesn't make sense, as a user I'd be very confused. 

Additionally, I checked the file service code and I don't think it explicitly blocks specific extensions - it checks for known extensions and sets the mime_type if the mapping found, otherwise it marks the file `application/octet-stream` , so the logic we're doing here does not entirely match VIP Files Service.

https://github.com/Automattic/wpvip-files-service/blob/a27ac4400b901bcdf48655be39ac4e1dd9cc731f/vipv2-files-web/src/class-files-service.php#L144-L162

And last but not least, we seem to be blocking fonts here, which is not great because they're allowed both in WordPress and in File Service.

src/lib/validations/utils.ts
```diff
@@ -35,3 +40,34 @@ export function getMultilineStatement( statementRegex: RegExp ): ( line: string
 		return matchingStatements;
 	};
 }
+
+export function getFileType( filepath: string ): string {
+	// Check if the computer has the file command installed
+	try {
+		cp.execFileSync( '/usr/bin/file', [ '-v' ] );
```
ariskataoka, line 47: I believe this check could be better placed somewhere else where it would be shown outside of the loop. 
Maybe before it starts, what do you think? 
https://github.com/Automattic/vip-cli/pull/1970/files#diff-84484f13e45d85ea3750110cf1f1485773c64ec3efefe7278f3a6d720e26d9a7L62

saroshaga:
> So I tested the latest version under Windows native, and WSL on Ubuntu.

🙌 thanks @rinatkhaziev !! Are you using a VM for Windows or did you have a different machine? :D 

> On Windows it repeatedly presents the error message `"Failed to extract mimetype of file %FILEPATH% because of error: /usr/bin/file does not exist on this computer"`, I think it's a bit excessive, can we group the errors somehow? A Linux/Mac path under Windows also doesn't make sense, as a user I'd be very confused.

Yeah I've changed the PR label to `In Progress` . We're trying to figure out the best approach and we will definitely have to reference the PATH based on the OS that we detect. We're also evaluating the `file-type` library but we'd like to avoid if possible given that the actual command `vip import media` uses the `file` command.

> Additionally, I checked the file service code and I don't think it explicitly blocks specific extensions - it checks for known extensions and sets the mime_type if the mapping found, otherwise it marks the file `application/octet-stream` , so the logic we're doing here does not entirely match VIP Files Service.
>
> https://github.com/Automattic/wpvip-files-service/blob/a27ac4400b901bcdf48655be39ac4e1dd9cc731f/vipv2-files-web/src/class-files-service.php#L144-L162

Taking this conversation elsewhere

> And last but not least, we seem to be blocking fonts here, which is not great because they're allowed both in WordPress and in File Service.

Are there specific font types supported in WordPress? Can you please point me to that?

rinatkhaziev:
> Are there specific font types supported in WordPress? Can you please point me to that?

woff, woff2, otf, ttf, it's a bit weird to test because they're not uploadable via Media Library, but can be uploaded via the new feature called Fonts Library.

saroshaga:
> > Are there specific font types supported in WordPress? Can you please point me to that?
> 
> woff, woff2, otf, ttf, it's a bit weird to test because they're not uploadable via Media Library, but can be uploaded via the new feature called Fonts Library.

They are also uploaded to `wp-content/fonts` and not `wp-content/uploads`, so that'll be interesting for VIP 
https://make.wordpress.org/core/2024/03/14/new-feature-font-library/#customizing-the-fonts-upload-directory

rinatkhaziev:
@saroshaga no, that was rolled back and it's `wp-content/uploads/fonts` now.

PR #2007, opened 2024-08-27
Add saveErrorLog flag to import media command

### LINKED ISSUE #2004: Allow error report save option to be set via flag (import media command)
## Expected/Desired Behavior

The prompt that asks users to `Download import errors report now?` should be able to be bypassed by setting the preference via a flag passed to the `vip import media` command. It's possible to otherwise run the command without interaction (using the `--force` flag to bypass the confirmation prompt) and it would be useful to also set this, so that several media import operations could be performed in sequence in a shell script (e.g., importing several smaller archives).

## Actual Behavior

When an import operation concludes, if there were any errors (e.g., image already exists and was skipped, or was not able to be uploaded for some reason, like a filename that is too long) the script stops and waits for user input.

1. Upload images using `vip import media` which generate errors (such as file already exists, without the overwrite option)
2. Observe that there is a prompt for what to do with the generated error log that is not able to be set via a flag to the command

Flag should allow this block of code to be bypassed: https://github.com/Automattic/vip-cli/blob/trunk/src/lib/media-import/status.ts#L358-L361

kevinfodness:
## Description

Adds a `--saveErrorLog` flag to the `import media` command to allow the user to bypass the confirmation prompt for saving an error log if it is generated.

Fixes #2004 

## Pull request checklist

- [X] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables.
- [X] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs).
- [X] [Manually test](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#manual-testing) the relevant changes.
- [X] Follow the [pull request checklist](https://github.com/Automattic/vip-cli/blob/trunk/docs/RELEASING.md#new-pull-requests)
- [X] Add/update [automated tests](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#automated-testing) as needed.

## New release checklist

## Steps to Test

Note: I was not able to get the `vip` command to run locally in my tests - it was erroring with `Error: spawn vip.js-import ENOENT` and I'm not sure how to get around this. Hopefully an Automattician can help to test this change or provide me with instructions on how to test locally.

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip.js import media` with the new `--saveErrorLog` flag set targeting an archive that contains a file that will error.
1. Confirm that the error log is saved without user input.

### DIFF
--- a/src/bin/vip-import-media-status.js
+++ b/src/bin/vip-import-media-status.js
@@ -34,11 +34,7 @@ command( {
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
-	.option(
-		'saveErrorLog',
-		'Download file-error logs without prompting',
-		false
-	)
+	.option( 'saveErrorLog', 'Download file-error logs without prompting', 'prompt' )
 	.argv( process.argv, async ( arg, { app, env, exportFileErrorsToJson, saveErrorLog } ) => {
 		const { id: envId, appId } = env;
 		const track = trackEventWithEnv.bind( null, appId, envId );
--- a/src/bin/vip-import-media-status.js
+++ b/src/bin/vip-import-media-status.js
@@ -34,7 +34,12 @@ command( {
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
-	.argv( process.argv, async ( arg, { app, env, exportFileErrorsToJson } ) => {
+	.option(
+		'saveErrorLog',
+		'If errors were encountered, save the error log file without prompting',
+		false
+	)
+	.argv( process.argv, async ( arg, { app, env, exportFileErrorsToJson, saveErrorLog } ) => {
 		const { id: envId, appId } = env;
 		const track = trackEventWithEnv.bind( null, appId, envId );
 
@@ -53,5 +58,11 @@ command( {
 Checking the Media import status for your environment...
 `;
 
-		await mediaImportCheckStatus( { app, env, progressTracker, exportFileErrorsToJson } );
+		await mediaImportCheckStatus( {
+			app,
+			env,
+			progressTracker,
+			exportFileErrorsToJson,
+			saveErrorLog,
+		} );
 	} );
--- a/src/bin/vip-import-media-status.js
+++ b/src/bin/vip-import-media-status.js
@@ -36,7 +36,7 @@ command( {
 	)
 	.option(
 		'saveErrorLog',
-		'If errors were encountered, save the error log file without prompting',
+		'Download file-error logs without prompting',
 		false
 	)
 	.argv( process.argv, async ( arg, { app, env, exportFileErrorsToJson, saveErrorLog } ) => {
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -100,7 +100,7 @@ Are you sure you want to import the contents of the url?
 	)
 	.option(
 		'saveErrorLog',
-		'If errors were encountered, save the error log file without prompting',
+		'Download file-error logs without prompting',
 		false
 	)
 	.option( 'overwriteExistingFiles', 'Overwrite any existing files', false )
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -98,7 +98,7 @@ Are you sure you want to import the contents of the url?
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
-	.option( 'saveErrorLog', 'Download file-error logs without prompting', 'prompt' )
+	.option( 'saveErrorLog', 'Download file-error logs without prompting' )
 	.option( 'overwriteExistingFiles', 'Overwrite any existing files', false )
 	.option( 'importIntermediateImages', 'Import intermediate image files', false )
 	.examples( examples )
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -98,18 +98,29 @@ Are you sure you want to import the contents of the url?
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
+	.option(
+		'saveErrorLog',
+		'If errors were encountered, save the error log file without prompting',
+		false
+	)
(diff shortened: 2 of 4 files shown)


aswasif007:
Hi @kevinfodness 👋 
Instead of adding `--saveErrorLog` flag, I think it is better to make sure that the `--force` flag bypass the prompt in question. I've raised a PR for that: https://github.com/Automattic/vip-cli/pull/2011

src/bin/vip-import-media-status.js
```diff
@@ -34,7 +34,12 @@ command( {
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
-	.argv( process.argv, async ( arg, { app, env, exportFileErrorsToJson } ) => {
+	.option(
+		'saveErrorLog',
+		'If errors were encountered, save the error log file without prompting',
```
aswasif007: Let's make this description a bit more concise?

```suggestion
		'Download file-error logs without prompting',
```

src/bin/vip-import-media.js
```diff
@@ -98,18 +98,29 @@ Are you sure you want to import the contents of the url?
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
+	.option(
+		'saveErrorLog',
+		'If errors were encountered, save the error log file without prompting',
```
aswasif007: Let's do the same here too:

```suggestion
		'Download file-error logs without prompting',
```

src/lib/media-import/status.ts
```diff
@@ -306,12 +308,16 @@ Downloading errors details from ${ fileErrorsUrl }
...
+		const failureDetails = ! saveErrorLog
+			? await prompt( {
+					type: 'confirm',
+					name: 'download',
+					message:
+						'Download import errors report now? (Report will be downloadable for up to 7 days from the completion of the import)',
+			  } )
+			: {
+					download: true,
+			  };
```
aswasif007: If `--saveErrorLog no` is specified, we still get this:
```
? Download import errors report now? (Report will be downloadable for up to 7 days from the completion of the import) (y/N)
```

I think it should not ask for user-input on whether to download the logs or not, since the user has already answered that.

aswasif007: We have to maintain three states of `saveErrorLog`: `true`, `false` and `undefined`.

kevinfodness: True - we'll need to make it not-a-boolean, then, since it would have three possible values - yes, no, or prompt.

aswasif007: Correct.

kevinfodness: I pushed up a change that should address this. I'm testing it locally now.

kevinfodness: Found a bug - if you pass the flag with no value, like `--saveErrorLog`, it doesn't treat that as "true." I'll try to address.

kevinfodness: Because this is a string type now and not a boolean, I don't think we'll be able to allow users to pass it without specifying a value and have it count as "true" unless we change the default. Therefore, it will need to be invoked with `--saveErrorLog=true` or `--saveErrorLog=yes` in order to bypass, and `prompt` will be the default, and users can pass `false` or `no` to not prompt and have it not save the log file. I'm pushing up this change now.

kevinfodness: This is working well in my tests so far.

aswasif007: This seems to solve the issue:

```diff
diff --git a/src/bin/vip-import-media.js b/src/bin/vip-import-media.js
index 39f23e1a..ceffe1b8 100755
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -98,7 +98,7 @@ Are you sure you want to import the contents of the url?
 		'Export any file errors encountered to a JSON file instead of a plain text file',
 		false
 	)
-	.option( 'saveErrorLog', 'Download file-error logs without prompting', 'prompt' )
+	.option( 'saveErrorLog', 'Download file-error logs without prompting' )
 	.option( 'overwriteExistingFiles', 'Overwrite any existing files', false )
 	.option( 'importIntermediateImages', 'Import intermediate image files', false )
 	.examples( examples )
diff --git a/src/lib/cli/command.js b/src/lib/cli/command.js
index 1863147a..68c11420 100644
--- a/src/lib/cli/command.js
+++ b/src/lib/cli/command.js
@@ -476,6 +476,8 @@ args.argv = async function ( argv, cb ) {
 					value: options.exportFileErrorsToJson ? '✅ Yes' : `${ chalk.red( 'x' ) } No`,
 				} );

+				options.saveErrorLog =
+					options.saveErrorLog === undefined ? 'prompt' : `${ options.saveErrorLog }`;
 				info.push( {
 					key: 'Download file-error logs?',
 					value: options.saveErrorLog,
```

kevinfodness: That block of code only runs if the user doesn't specify `--force`. I'd imagine that the most common use case is in combination with `--force` to allow commands to be chained without user input. We could move negotiation of that option up to above the if conditional for force, though. I'll take a pass at this.

kevinfodness: I pushed up a change that should allow for this. There are a few other places where option values are being set only if the force flag is not set which you may want to also move up to this new switch statement: [here](https://github.com/Automattic/vip-cli/blob/trunk/src/lib/cli/command.js#L392-L395), [here](https://github.com/Automattic/vip-cli/blob/trunk/src/lib/cli/command.js#L452-L455), [here](https://github.com/Automattic/vip-cli/blob/trunk/src/lib/cli/command.js#L461-L464), and [here](https://github.com/Automattic/vip-cli/blob/trunk/src/lib/cli/command.js#L470-L473).

aswasif007:
Many thanks for working on this PR, @kevinfodness 👏

aswasif007 (approved):

PR #2011, opened 2024-08-30
fix(import-media): Bypass download-error prompt if --force option is set

### LINKED ISSUE #2004: Allow error report save option to be set via flag (import media command)

aswasif007:
## Issue

https://github.com/Automattic/vip-cli/issues/2004

## Description

Generally, `--force` flag in any command is supposed to bypass the prompts so that the command can be executed via a script, without any manual interaction. But `vip import media` command has a `download-errors?` prompt that can't be bypassed with the `--force` flag. In this PR, we are fixing that issue.

## Pull request checklist

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip import media @<app>.<env> https://getsamplefiles.com/download/zip/sample-1.zip` 
2. Run the above command again. The reason we are running it twice is to ensure that some `file-already-exists` errors are generated. You should see that the `Download import errors report now?` prompt has appeared.
1. Now, run the same command again, but with the `--force` flag. This time, the above prompt should not be present, and the errors should be downloaded automatically.

### DIFF
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -102,8 +102,14 @@ Are you sure you want to import the contents of the url?
 	.option( 'importIntermediateImages', 'Import intermediate image files', false )
 	.examples( examples )
 	.argv( process.argv, async ( args, opts ) => {
-		const { app, env, exportFileErrorsToJson, overwriteExistingFiles, importIntermediateImages } =
-			opts;
+		const {
+			app,
+			env,
+			exportFileErrorsToJson,
+			overwriteExistingFiles,
+			importIntermediateImages,
+			force,
+		} = opts;
 		const [ url ] = args;
 
 		if ( ! isSupportedUrl( url ) ) {
@@ -149,7 +155,7 @@ Importing Media into your App...
 				},
 			} );
 
-			await mediaImportCheckStatus( { app, env, progressTracker, exportFileErrorsToJson } );
+			await mediaImportCheckStatus( { app, env, progressTracker, exportFileErrorsToJson, force } );
 		} catch ( error ) {
 			if ( error.graphQLErrors ) {
 				for ( const err of error.graphQLErrors ) {
--- a/src/lib/media-import/status.ts
+++ b/src/lib/media-import/status.ts
@@ -55,6 +55,7 @@ export interface MediaImportCheckStatusInput {
 	env: AppEnvironment;
 	progressTracker: MediaImportProgressTracker;
 	exportFileErrorsToJson: boolean;
+	force?: boolean;
 }
 
 async function getStatus(
@@ -159,6 +160,7 @@ export async function mediaImportCheckStatus( {
 	env,
 	progressTracker,
 	exportFileErrorsToJson,
+	force,
 }: MediaImportCheckStatusInput ) {
 	// Stop printing so we can pass our callback
 	progressTracker.stopPrinting();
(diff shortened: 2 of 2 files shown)


src/bin/vip-import-media.js
```diff
@@ -102,8 +102,14 @@ Are you sure you want to import the contents of the url?
...
 	.argv( process.argv, async ( args, opts ) => {
-		const { app, env, exportFileErrorsToJson, overwriteExistingFiles, importIntermediateImages } =
-			opts;
+		const {
+			app,
+			env,
+			exportFileErrorsToJson,
+			overwriteExistingFiles,
+			importIntermediateImages,
+			force,
```
saroshaga, line 111: Traditionally, `--force` is also used to bypass warnings in the CLI. I'm wondering if we should use a name that conveys exactly what we're trying to "force" . Perhaps `'--download-failure-report', '-D'` ? It's a mouthful but I would rather be precise and we would have the shortcode option as well.

saroshaga (commented):
Thanks for picking this up @aswasif007!

src/bin/vip-import-media.js
saroshaga: I think the global `--force` flag was meant to skip the confirmation that some commands require _before_ the command is actually run.

https://github.com/Automattic/vip-cli/blob/trunk/src/lib/cli/command.js#L582

src/lib/media-import/status.ts
```diff
@@ -306,28 +308,30 @@ Downloading errors details from ${ fileErrorsUrl }
 	}
 
 	async function promptFailureDetailsDownload( fileErrorsUrl: string ) {
-		const failureDetails = await prompt( {
-			type: 'confirm',
-			name: 'download',
-			message:
-				'Download import errors report now? (Report will be downloadable for up to 7 days from the completion of the import)',
-		} );
+		if ( ! force ) {
```
saroshaga, line 311: If the default value for this is `false` then will the user even know that they _could_ download the report?

src/bin/vip-import-media.js
aswasif007: What if the user who is running a script wants the error reports to be downloaded without manual intervention?

src/lib/media-import/status.ts
aswasif007: If the value of `force` is false, then the user would see the prompt, which tells the user this:
`Download import errors report now? (Report will be downloadable for up to 7 days from the completion of the import)`

aswasif007: I don't think I understood your question tbh

src/bin/vip-import-media.js
saroshaga: > What if the user who is running a script wants the error reports to be downloaded without manual intervention?

Same idea could be applied to that. I've edited my original comment, please check.

src/lib/media-import/status.ts
saroshaga: Ah right! thanks for explaining! I think I got confused.

src/bin/vip-import-media.js
kevinfodness: @saroshaga adding the more specific flag is what I was going for in #2007 although I personally don't mind if `--force` auto-saves the log as well.

aswasif007: @saroshaga Thanks for your input. Let's go with #2007 then. I will close this PR.

aswasif007:
Closing in favour of https://github.com/Automattic/vip-cli/pull/2007

PR #2012, opened 2024-08-30
fix(import-media): Display success/total count even if all files fail to import

### LINKED ISSUE #2005: Print total files processed even when they all error (import media command)
## Expected/Desired Behavior

When importing an archive of media files via `vip import media`, if all files in the archive error out, no status is printed. This can happen if importing an archive of files that was already imported to the platform without asking the command to overwrite files, in which case it will skip them because they've already been imported. It would be ideal to still print the number of total files, but show zero processed, like `Imported Files: 0/100 - 0%`

## Actual Behavior

The line that normally includes the number of imported files vs. the total files in the archive is missing, like this:

```
Importing Media into your App...

Status: COMPLETED ✓ : The imported files should be visible on your App
```

1. Use the `vip import media` command to import an archive of images into your application.
2. Re-run the import with the same archive and observe that the line that normally prints the number of processed files vs. total files is missing.

I believe the issue is that this line is checking to see if any files have been processed, which in the case of errors will always be zero: https://github.com/Automattic/vip-cli/blob/10419d678684d94c8738928a7bcbff2c6911366a/src/lib/media-import/progress.ts#L66 - it would be ideal to only check the total number of files so that progress can be shown immediately, updated when files are processed, and will be available after the command has finished executing so the user can verify the total number of processed files, even if they have all errored out.

aswasif007:
## Issue

https://github.com/Automattic/vip-cli/issues/2005

## Description

In the `vip import media` command, when all files in the archive fails/errors while importing, the import does not show the success/total count in the end. This is because of a bug in the null-check filter where the case of `filesProcessed = 0` is being consumed by the null check.

## Pull request checklist

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `node ./dist/bin/vip import media @<app>.<env> https://getsamplefiles.com/download/zip/sample-1.zip` twice. In the second run, all the files should fail to import as they are already imported. When that happens, make sure that the `Imported Files: 0/4 - 0% ✓` line is present in the output.

### DIFF
--- a/src/lib/media-import/progress.ts
+++ b/src/lib/media-import/progress.ts
@@ -63,7 +63,7 @@ export class MediaImportProgressTracker {
 
 		const statusIcon = getGlyphForStatus( this.status.status ?? '', this.runningSprite );
 		let logs;
-		if ( this.status.filesProcessed && this.status.filesTotal ) {
+		if ( typeof this.status.filesProcessed === 'number' && this.status.filesTotal ) {
 			const progressPercentage = Math.floor(
 				( this.status.filesProcessed / this.status.filesTotal ) * 100
 			);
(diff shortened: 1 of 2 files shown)


saroshaga (approved):
Thanks for fixing this!

PR #2054, opened 2024-10-14
Update the `import` command group to follow the VIP-CLI style guide

yolih:
## Description

Updated the voice, tense, and format of command descriptions, examples, and usage models for the `vip import` command group to follow the VIP-CLI style guide.

## Pull request checklist

- [ n/a ] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables.
- [ ] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs).
- [ ] [Manually test](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#manual-testing) the relevant changes.
- [ x ] Follow the [pull request checklist](https://github.com/Automattic/vip-cli/blob/trunk/docs/RELEASING.md#new-pull-requests)
- [ x ] Add/update [automated tests](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#automated-testing) as needed.

## New release checklist

- [ ] [Automated tests](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#automated-testing) pass.
- [ ] The [Preparing for release checklist](https://github.com/Automattic/vip-cli/blob/trunk/docs/RELEASING.md#preparing-for-release) is completed.

## Steps to Test

Observe the results of the `--help` display for impacted commands:

- `vip import`
- `vip import validate-files`
- `vip import validate-sql`
- `vip import sql`
- `vip import sql status`
- `vip import media`
- `vip import media abort`
- `vip import media status`

PR #2056, opened 2024-10-15
Fix `--output` option of `vip import sql` command

mehmoodak:
## Description

Fixed an issue with the `output` option of the `vip import sql` command, which wasn't creating an output file as expected. Alongside this fix, this PR also improves the help text for the `vip search-replace` command.

## Pull request checklist

- [N/A] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables.
- [N/A] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs).

## New release checklist

## Steps to Test

1. Check out PR.
1. Run `vip @example-app.develop export sql --output=output.sql.gz`
1. Extracts the sql file from gzip.
1. Run `node ./dist/bin/vip-import-sql.js @example-app.develop output.sql --search-replace="from,to" --output="test-output.sql"`
1. Verify `test-output.sql` file is created with the updated content.

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -397,11 +397,7 @@ void command( {
 		'Search for a given URL in the SQL file and replace it with another. The format for the value is `<from-url>,<to-url>` (e.g. --search-replace="https://from.com,https://to.com"'
 	)
 	.option( 'in-place', 'Search and Replace explicitly on the given input file' )
-	.option(
-		'output',
-		'Specify the replacement output file for Search and Replace',
-		'process.stdout'
-	)
+	.option( 'output', 'Specify the replacement output file for Search and Replace' )
 	.examples( examples )
 	.argv( process.argv, async ( arg, opts ) => {
 		const { app, env } = opts;
@@ -512,7 +508,7 @@ Processing the SQL import for your environment...
 			const { outputFileName } = await searchAndReplace( fileName, searchReplace, {
 				isImport: true,
 				inPlace: opts.inPlace,
-				output: true,
+				output: opts.output ?? true, // "true" creates a temp output file instead of printing to stdout, as we need to upload the output to S3.
 			} );
 
 			if ( typeof outputFileName !== 'string' ) {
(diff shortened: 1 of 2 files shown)


yolih (commented):
@mehmoodak Thank you so much for fixing this bug! I have another PR open to improve the command descriptions and examples: #2054 
Would it be possible for your PR to only fix the `--output` bug and for my PR to make the updates to the `--help` menu?

mehmoodak:
> Would it be possible for your PR to only fix the --output bug and for my PR to make the updates to the --help menu?

@yolih I have made some changes, please take a look again.

yolih:
Thanks, @mehmoodak ! 🙌 

A couple more questions:

- Does the `--output` option behave differently when passed with the `vip import sql` and the `vip search-replace` commands? If the behavior of  `--output` is the same for both commands then I think the description for the option should also be the same where it appears in the `--help` menu.
- Could you tell me why you removed "Defaults to STDOUT" for the `vip import sql` command? I believe that is the correct default for `--output` with that command as well.

mehmoodak:
@yolih The key difference is that `vip import sql` requires a SQL file to be uploaded to S3, and printing to `stdout` will result in an error. In contrast, `vip search-replace` can use stdout without any issues.

yolih:
Oooohhhh... interesting. Thank you for your patience with all of my questions! 

Based on your explanation, I agree with removing `Defaults to STDOUT` from the `vip import sql` version of `--output`. 

For the `vip search-replace` version of `--output`, I suggest that the description in the help menu reads something like:

> Create a local copy of the file with the completed search and replace operations. Ignored if the command includes --in-place. Accepts a local file path. Defaults to STDOUT.

What do you think?

mehmoodak:
> What do you think?

@yolih help text updated. Thanks for helping me improve it.

yolih (approved):
LGTM. Thanks for the collaboration!

PR #2079, opened 2024-10-31
 Update the `search-replace` command to follow the VIP-CLI style guide

yolih:
## Description

Updated the voice, tense, and format of command descriptions, examples, and usage models for the `vip search-replace` command to follow the VIP-CLI style guide.

## Pull request checklist

## New release checklist

## Steps to Test

Observe the results of the `--help` display for the impacted commands:

- `vip`
- `vip search-replace`

### DIFF
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -400,7 +400,7 @@ void command( {
 	)
 	.option(
 		'in-place',
-		'Perform a search and replace operation on a local SQL file, save the results to the file, then import the updated file.'
+		'Perform a search and replace operation on the local input file, and save the results to the file prior to import.'
 	)
 	.option(
 		'output',
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -404,7 +404,7 @@ void command( {
 	)
 	.option(
 		'output',
-		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command, and ignored if the command excludes a --search-replace operation.'
+		'The local file path to save a copy of the results from the search and replace operation when the --search-replace option is passed. Ignored when used with the --in-place option.'
 	)
 	.examples( examples )
 	.argv( process.argv, async ( arg, opts ) => {
--- a/src/bin/vip-import-sql.js
+++ b/src/bin/vip-import-sql.js
@@ -400,11 +400,11 @@ void command( {
 	)
 	.option(
 		'in-place',
-		'Perform a search and replace operation on the local input file, and save the results to the file prior to import.'
+		'Overwrite the local input file with the results of the search and replace operation prior to import.'
 	)
 	.option(
 		'output',
-		'Create a local copy of the imported file with the completed search and replace operations. Ignored if the command includes --in-place, or excludes a --search-replace operation. Accepts a local file path.'
+		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command, and ignored if the command excludes a --search-replace operation.'
 	)
 	.examples( examples )
 	.argv( process.argv, async ( arg, opts ) => {
--- a/src/bin/vip-search-replace.js
+++ b/src/bin/vip-search-replace.js
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
 const examples = [
 	// `search-replace` flag
 	{
-		usage: 'vip search-replace <file.sql> --search-replace="from,to"',
+		usage: 'vip search-replace file.sql --search-replace="from,to"',
 		description:
-			'Replace instances of <from> with <to> in the provided <file.sql>\n' +
-			'       * Ensure there are no spaces between your search-replace parameters',
+			'Search for every instance of the value "from" in the local input file named "file.sql" and replace it with the value "to".\n' +
+			'       * Results of the operation output to STDOUT by default.',
 	},
 	// `in-place` flag
 	{
-		usage: 'vip search-replace <file.sql> --search-replace="from,to" --in-place',
-		description: 'Perform Search and Replace explicitly on the provided input <file.sql> file',
+		usage: 'vip search-replace file.sql --search-replace="from,to" --in-place',
+		description:
+			'Perform the search and replace operation and save the results to the local input file "file.sql".',
 	},
 	// `output` flag
 	{
-		usage: 'vip search-replace <file.sql> --search-replace="from,to" --output="<output.sql>"',
+		usage: 'vip search-replace file.sql --search-replace="from,to" --output=output-file.sql',
 		description:
-			'Search and Replace to the specified output <output.sql> file\n' +
-			'       * Has no effect when the `in-place` flag is used',
+			'Perform the search and replace operation and save the results to a local clone of the input file named "output-file.sql".',
 	},
 ];
 
 command( {
 	requiredArgs: 1,
 } )
-	.option( 'search-replace', 'Specify the <from> and <to> pairs to be replaced' )
-	.option( 'in-place', 'Perform the search and replace explicitly on the input file' )
+	.option(
+		'search-replace',
+		'Pass a string value to search for and a string value to replace it with. Separate the values by a comma only; no spaces (e.g. --search-replace=“from,to”).'
+	)
+	.option(
+		'in-place',
+		'Save the results of a search and replace operation to the local input file.'
+	)
 	.option(
 		'output',
-		'Create a local copy of the file with the completed search and replace operations. Ignored if the command includes --in-place. Accepts a local file path. Defaults to STDOUT.'
+		'Save the results of the search and replace operation to a clone of the local input file. Ignored if the command includes --in-place. Accepts a local file path.'
 	)
 	.examples( examples )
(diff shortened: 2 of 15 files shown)


src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
+			'Search for every instance of the value "from" in the local input file named "file.sql" and replace it with the value "to".\n' +
+			'       * Results of the operation output to STDOUT by default.',
 	},
 	// `in-place` flag
 	{
-		usage: 'vip search-replace <file.sql> --search-replace="from,to" --in-place',
-		description: 'Perform Search and Replace explicitly on the provided input <file.sql> file',
+		usage: 'vip search-replace file.sql --search-replace="from,to" --in-place',
+		description:
+			'Perform the search and replace operation and save the results to the local input file "file.sql".',
```
sjinks: Should we mention that the S&R operation is performed on `file.sql`?

src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
 	},
 	// `output` flag
 	{
-		usage: 'vip search-replace <file.sql> --search-replace="from,to" --output="<output.sql>"',
+		usage: 'vip search-replace file.sql --search-replace="from,to" --output=output-file.sql',
 		description:
-			'Search and Replace to the specified output <output.sql> file\n' +
-			'       * Has no effect when the `in-place` flag is used',
+			'Perform the search and replace operation and save the results to a local clone of the input file named "output-file.sql".',
```
sjinks, line 30: It is unclear to me (from the previous description) where the output will go. Given the "Has no effect when the `in-place` flag is used," I would assume that the in-place operation would be performed on the source file (`file.sql`) and the result will be saved to that very file, not to `output-file.sql`.

src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
 ];
 
 command( {
 	requiredArgs: 1,
 } )
-	.option( 'search-replace', 'Specify the <from> and <to> pairs to be replaced' )
-	.option( 'in-place', 'Perform the search and replace explicitly on the input file' )
+	.option(
+		'search-replace',
+		'Pass a string value to search for and a string value to replace it with. Separate the values by a comma only; no spaces (e.g. --search-replace=“from,to”).'
```
sjinks:
```suggestion
		'Pass a string value to search for and a string value to replace it with. Separate the values by a comma only; no spaces (e.g. --search-replace="from,to").'
```

(fixed quotes in case someone copies and pastes the example)

src/bin/vip-search-replace.js
sjinks: See https://github.com/Automattic/vip-cli/pull/2079/files#diff-1d697c6fa0eb15c3a87ec97840846b8674b75621357aac3c1e65681f03f25708R47

src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
+	)
+	.option(
+		'in-place',
+		'Save the results of a search and replace operation to the local input file.'
```
aswasif007: I feel like we should use the word "overwrite" somewhere in there, because that is more catchy, and as soon as the user notices this word, the purpose of `in-place` becomes clear without reading the full sentence.

Example:
`Overwrite the local input file with the results of the search and replace operation`

src/bin/vip-import-sql.js
```diff
@@ -400,7 +400,7 @@ void command( {
 	)
 	.option(
 		'in-place',
-		'Perform a search and replace operation on a local SQL file, save the results to the file, then import the updated file.'
+		'Perform a search and replace operation on the local input file, and save the results to the file prior to import.'
```
aswasif007: Same as https://github.com/Automattic/vip-cli/pull/2079#discussion_r1826916772

src/bin/vip-search-replace.js
yolih: One of the goals of the updated descriptions is to put any limitations or requirements for a command option in the description for the option, not in the usage example. 

I agree with you that the "Has no effect" phrase is unclear. What if we updated the description for `--output` to something like this:

```
.option(
		'output',
		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command.'
	)
```

_(I edited the above description after reading the suggestion from @aswasif007)_

src/bin/vip-search-replace.js
yolih: I like that suggestion!

src/bin/vip-import-sql.js
yolih: How about this:

```
.option(
		'in-place',
		'Overwrite the local input file with the results of the search and replace operation prior to import.'
	)
	.option(
		'output',
		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command, and ignored if the command excludes a --search-replace operation.'
	)
```

src/bin/vip-search-replace.js
yolih: I'm having a hard time understanding what needs to be improved in order to address your concern.

Is this wording a bit better?
`Perform the search and replace operation on the local input file "file.sql" and overwrite the file with the results.`

If not, I welcome a version of the description from you that is clearer. 🙏

src/bin/vip-import-sql.js
aswasif007: That looks good! 👍

src/bin/vip-search-replace.js
terriann: I'd suggest instead:

```
A local file name to save the results of the search and replace operation. Has no effect if the --in-place parameter is also included in the command.
```

src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
 ];
 
 command( {
 	requiredArgs: 1,
 } )
-	.option( 'search-replace', 'Specify the <from> and <to> pairs to be replaced' )
-	.option( 'in-place', 'Perform the search and replace explicitly on the input file' )
+	.option(
+		'search-replace',
+		'Pass a string value to search for and a string value to replace it with. Separate the values by a comma only; no spaces (e.g. --search-replace="from,to").'
```
terriann: We can make this more concise.

```suggestion
		'A comma-separated pair of strings specifying the search and replacement values (e.g., --search-replace="from,to").'
```

src/bin/vip-search-replace.js
yolih: I think it's important to make it clear that the option will create the file; that the file does not need to pre-exist for the option to work. 🤔

src/bin/vip-search-replace.js
yolih: I like your proposed concise version! I am inclined to stick with the term "replace" instead of "replacement" though. What do you think of this version?
`A comma-separated pair of strings that specify the values to search for and replace (e.g., --search-replace="from,to").`

src/bin/vip-search-replace.js
```diff
@@ -12,33 +12,39 @@ const debug = debugLib( '@automattic/vip:bin:vip-search-replace' );
...
+		'A comma-separated pair of strings that specify the values to search for and replace (e.g. --search-replace="from,to").'
+	)
+	.option(
+		'in-place',
+		'Overwrite the local input file with the results of the search and replace operation.'
+	)
 	.option(
 		'output',
-		'Create a local copy of the file with the completed search and replace operations. Ignored if the command includes --in-place. Accepts a local file path. Defaults to STDOUT.'
+		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command.'
```
terriann: This can be more concise and use more straightforward language without losing clarity.

```suggestion
		'The local file path used to save a copy of the results from the search and replace operation. Ignored when used with the --in-place option.'
```

Overridden feels inaccurate because the output location isn't superseded by a value passed with the `--in-place` argument; it's ignored.

src/bin/vip-import-sql.js
```diff
@@ -400,11 +400,11 @@ void command( {
 	)
 	.option(
 		'in-place',
-		'Perform a search and replace operation on a local SQL file, save the results to the file, then import the updated file.'
+		'Overwrite the local input file with the results of the search and replace operation prior to import.'
 	)
 	.option(
 		'output',
-		'Create a local copy of the imported file with the completed search and replace operations. Ignored if the command includes --in-place, or excludes a --search-replace operation. Accepts a local file path.'
+		'Clone the local input file and overwrite the clone with the results of the search and replace operation. Accepts a local file path. Overridden by --in-place if both options are run in the same command, and ignored if the command excludes a --search-replace operation.'
```
terriann: This description is very verbose. The following feels just as accurate in describing the outcome for the user while using fewer words and simpler language.

```suggestion
		'The local file path to save a copy of the results from the search and replace operation when the --search-replace option is passed. Ignored when used with the --in-place option.'
```

terriann (approved):
👍

PR #2084, opened 2024-11-06
Provide dedicated error message for when a token is marked as disabled due to inactivity

andrea-sdl:
## Description

This PR adds support for the new unauthorized error for inactive tokens. 

## Changelog Description

Adds support for the disabled token due to inactivity error so that it shows a dedicated error message.

## Pull request checklist

- [X] [Manually test](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#manual-testing) the relevant changes.

## New release checklist

## Steps to Test

1. Check out PR.
1. Run `npm run build`
1. Run `VIP_PROXY="" API_HOST=http://localhost:4000 node ./dist/bin/vip app list` with a token that is disabled due to inactivity
1. Check the error message is 'Unauthorized: Your token has been disabled due to inactivity, please logout with `vip logout` then try again.'
1. Repeat the same test with a valid token and an expired token, to ensure previous behavior hasn't changed.

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -41,7 +41,7 @@ export default function API( {
 				'You are not authorized to perform this request; please logout with `vip logout`, then try again.';
 			if ( 'result' in networkError && networkError.result?.code === 'token-disabled-inactivity' ) {
 				message =
-					'The configured VIP-CLI token has been disabled due to inactivity, please logout with `vip logout` then try again.';
+					'Your token has been disabled due to inactivity; please log out with `vip logout`, then try again.';
 			}
 			console.error( chalk.red( 'Unauthorized:' ), message );
 			process.exit( 1 );
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -8,7 +8,6 @@ import { setContext } from '@apollo/client/link/context';
 import { ApolloLink } from '@apollo/client/link/core';
 import { ErrorResponse, onError } from '@apollo/client/link/error';
 import { ServerError } from '@apollo/client/link/utils';
-
 import chalk from 'chalk';
 
 import http from './api/http';
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -6,7 +6,9 @@ import {
 } from '@apollo/client/core';
 import { setContext } from '@apollo/client/link/context';
 import { ApolloLink } from '@apollo/client/link/core';
-import { onError } from '@apollo/client/link/error';
+import { ErrorResponse, onError } from '@apollo/client/link/error';
+import { ServerError } from '@apollo/client/link/utils';
+
 import chalk from 'chalk';
 
 import http from './api/http';
@@ -30,6 +32,15 @@ export function enableGlobalGraphQLErrorHandling(): void {
 	globalGraphQLErrorHandlingEnabled = true;
 }
 
+function isServerError(
+	networkError: ErrorResponse[ 'networkError' ]
+): networkError is ServerError {
+	if ( ! networkError ) {
+		return false;
+	}
+	return 'result' in networkError;
+}
+
 export default function API( {
 	exitOnError = true,
 }: {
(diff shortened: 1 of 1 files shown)


src/lib/api.ts
```diff
@@ -37,10 +37,13 @@ export default function API( {
 } = {} ): ApolloClient< NormalizedCacheObject > {
 	const errorLink = onError( ( { networkError, graphQLErrors } ) => {
 		if ( networkError && 'statusCode' in networkError && networkError.statusCode === 401 ) {
-			console.error(
-				chalk.red( 'Unauthorized:' ),
-				'You are unauthorized to perform this request, please logout with `vip logout` then try again.'
-			);
+			let message =
+				'You are unauthorized to perform this request, please logout with `vip logout` then try again.';
+			if ( 'result' in networkError && networkError.result?.code === 'token-disabled-inactivity' ) {
```
mjangda: Is the `in` check actually necessary since we have `?`?

```suggestion
			if ( networkError.result?.code === 'token-disabled-inactivity' ) {
```

src/lib/api.ts
```diff
@@ -37,10 +37,13 @@ export default function API( {
...
+				message =
+					'The configured VIP-CLI token has been disabled due to inactivity, please logout with `vip logout` then try again.';
```
mjangda:
```suggestion
					'Your token has been disabled due to inactivity; please log out with `vip logout`, then try again.';
```

src/lib/api.ts
```diff
@@ -37,10 +37,13 @@ export default function API( {
 } = {} ): ApolloClient< NormalizedCacheObject > {
 	const errorLink = onError( ( { networkError, graphQLErrors } ) => {
 		if ( networkError && 'statusCode' in networkError && networkError.statusCode === 401 ) {
-			console.error(
-				chalk.red( 'Unauthorized:' ),
-				'You are unauthorized to perform this request, please logout with `vip logout` then try again.'
-			);
+			let message =
+				'You are unauthorized to perform this request, please logout with `vip logout` then try again.';
```
mjangda:
```suggestion
				'You are not authorized to perform this request; please logout with `vip logout`, then try again.';
```

mjangda (approved):
Left some minor wording changes, otherwise this is great.

src/lib/api.ts
andrea-sdl: @mjangda We need the `result in` to make TS aware we're using the class `ServerError` instead of `ServerParseError` from the Apollo Client.

```
import { ServerError } from '../utils';
import { ServerParseError } from '../http';
export interface ErrorResponse {
    graphQLErrors?: ReadonlyArray<GraphQLError>;
    networkError?: Error | ServerError | ServerParseError;
    [....]
}
```

If we use your code the type checker will error out with

```
 Property 'result' does not exist on type 'ServerParseError'.

42    if ( networkError.result?.code === 'token-disabled-inactivity' ) {
```

mjangda: Interesting, thank you for explaining that.

Maybe we should make the check more obvious to indicate that?

```suggestion
			if ( networkError instanceof ServerError && networkError.result?.code === 'token-disabled-inactivity' ) {
```

andrea-sdl: This doesn't work either (I tested before landing with the `result in`) since we're checking against a type (ServerError) and not a class.

```
> tsc

src/lib/api.ts:43:33 - error TS2359: The right-hand side of an 'instanceof' expression must be either of type 'any', a class, function, or other type assignable to the 'Function' interface type, or an object type with a 'Symbol.hasInstance' method.

43    if ( networkError instanceof ServerError && networkError.result?.code === 'token-disabled-inactivity' ) {
                                   ~~~~~~~~~~~

src/lib/api.ts:43:61 - error TS2339: Property 'result' does not exist on type 'ServerError | ServerParseError'.
  Property 'result' does not exist on type 'ServerParseError'.

43    if ( networkError instanceof ServerError && networkError.result?.code === 'token-disabled-inactivity' ) {
                                                               ~~~~~~


Found 2 errors in the same file, starting at: src/lib/api.ts:43
```

What we can do, if we want to improve the readability, is to have a `isServerError` function with the same check @mjangda 

something like
```
function isServerError(networkError){
return 'result' in networkError;
}
```

mjangda: Even more interesting :)

I think the readability would be nice improvement but it's very minor.

andrea-sdl: Added in 0d8eefed9cf6d6ce5729ca0d2b0eb88b2f1ed57e . For future reference here's also the docs around using `in` operator narrowing:  https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing

PR #2139, opened 2024-12-05
refactor(dev-env): let `dev-tools` handle their volumes

sjinks:
## Description

This PR updates the dev-env code to avoid messing with the `dev-tools` volume.

1. We no longer remove `dev-tools` volumes;
2. We ensure that the `php` service starts after `devtools` completes to avoid race conditions;
3. We sync `/dev-tools-orig` with `/dev-tools`. This step is necessary because Lando sets its own entrypoint script if we don't provide ours.

Related PR: Automattic/vip-container-images#985 (must be merged before this one)
Volume removal logic was introduced in #1317.

## Pull request checklist

- [ ] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables.
- [ ] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs).
- [x] [Manually test](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#manual-testing) the relevant changes.
- [x] Follow the [pull request checklist](https://github.com/Automattic/vip-cli/blob/trunk/docs/RELEASING.md#new-pull-requests)
- [ ] Add/update [automated tests](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#automated-testing) as needed.

## New release checklist

- [ ] [Automated tests](https://github.com/Automattic/vip-cli/blob/trunk/docs/TESTING.md#automated-testing) pass.
- [ ] The [Preparing for release checklist](https://github.com/Automattic/vip-cli/blob/trunk/docs/RELEASING.md#preparing-for-release) is completed.

## Steps to Test

1. Check out Automattic/vip-container-images#985, `cd dev-tools && docker build -t dev-tools:local`
2. Check out this PR, `npm run build && npm link`
3. vip dev-env create < /dev/null
4. Create `~/.local/share/vip/dev-environment/vip-local/.lando.local.yml`:
```yml
services:
  devtools:
    services:
      image: dev-tools:local
      pull_policy: never
```
5. Run `vip dev-env start` and ensure everything works
6. Run `vip dev-env shell -- sh -c 'md5sum /dev-tools/*'` and save the output.
7. Update one of the dev-tools scripts and rebuild the image (`docker build -t dev-tools:local`)
8. Restart the environment (`vip dev-env start`).
[...]

PR #2143, opened 2024-12-08
refactor(dev-env): make `wordpress` init-only container

sjinks:
## Description

This PR updates the Dev Environment Lando template to make the `wordpress` service init-only (see #1341 for the background). Because we use `wordpress` only to provide a volume with WordPress installation, we don't need the container after the initialization stage. Letting the container die peacefully will free up some memory and CPU and reduce the number of mount points.

This PR and #2139 will let us merge the `dev-tools` and `wordpress` services into one. We will also be able to replace the dependency condition for `wordpress` from `service_started` to `service_completed_successfully`; this will simplify the WordPress setup script because Docker will start the `php` service after WordPress files are copied to where they belong.

## Pull request checklist

## New release checklist

## Steps to Test

1. Check out the PR, `npm run build && npm link`.
2. `vip dev-env create < /dev/null && vip dev-env start`.
3. `vip dev-env logs -S wordpress` must show nothing because we bypass Lando.
4. `docker ps -a --no-trunc` must show the entry point for the `wordpress` service as `"/usr/bin/rsync -a --chown=1000:1000 /wp/ /shared/"` (with the UID and GID of the current user instead of 1000:1000).
5. The environment must work.

### DIFF
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -198,10 +198,6 @@ services:
           target: /scripts
           volume:
             nocopy: true
-      environment:
-        LANDO_NO_USER_PERMS: 1
-        LANDO_NO_SCRIPTS: 1
-        LANDO_NEEDS_EXEC: 1
     initOnly: true
     entrypoint: /usr/bin/rsync -a --chown=${LANDO_HOST_USER_ID}:${LANDO_HOST_GROUP_ID} /wp/ /shared/
 
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -73,7 +73,7 @@ services:
         devtools:
           condition: service_completed_successfully
         wordpress:
-          condition: service_started
+          condition: service_completed_successfully
 <% if ( muPlugins.mode == 'image' ) { %>
         vip-mu-plugins:
           condition: service_started
--- a/assets/dev-env.lando.template.yml.ejs
+++ b/assets/dev-env.lando.template.yml.ejs
@@ -191,8 +191,6 @@ services:
     type: compose
     services:
       image: ghcr.io/automattic/vip-container-images/wordpress:<%= wordpress.tag %>
-#      command: /usr/local/bin/entrypoint.sh
-      command: sh -c "rsync -a --chown=www-data:www-data /wp/ /shared/; sleep infinity"
       volumes:
         - ./wordpress:/shared
         - type: volume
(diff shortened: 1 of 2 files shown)


src/lib/dev-environment/dev-environment-lando.ts
```diff
@@ -89,6 +89,10 @@ async function getLandoConfig(): Promise< LandoConfig > {
 		home: fakeHomeDir,
 		domain: 'vipdev.lndo.site',
 		version: 'unknown',
+		env: {
+			LANDO_HOST_USER_ID: process.platform === 'win32' ? '1000' : `${ userInfo().uid }`,
+			LANDO_HOST_GROUP_ID: process.platform === 'win32' ? '1000' : `${ userInfo().gid }`,
```
sjinks, line 94: These variables will be passed to `docker compose`; it will use them for substitutions inside `docker-compose` files.

See https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/ for details.

assets/dev-env.lando.template.yml.ejs
```diff
@@ -175,19 +175,19 @@ services:
...
-#      environment:
-#        LANDO_NO_SCRIPTS: 1
-#        LANDO_NEEDS_EXEC: 1
-#    initOnly: true
+      environment:
+        LANDO_NO_USER_PERMS: 1
+        LANDO_NO_SCRIPTS: 1
+        LANDO_NEEDS_EXEC: 1
+    initOnly: true
+    entrypoint: /usr/bin/rsync -a --chown=${LANDO_HOST_USER_ID}:${LANDO_HOST_GROUP_ID} /wp/ /shared/
```
sjinks, line 202: `${LANDO_HOST_USER_ID}` and `${LANDO_HOST_GROUP_ID}` will be taken directly from the environment; we don't need to spawn a shell to pass the values to the program.

Because the container is init-only, we don't need Lando scripts to manage permissions either; `rsync` perfectly understands numeric IDs and does not require the presence of `/etc/passwd`. This allows us to optimize WordPress images.

src/lib/dev-environment/dev-environment-lando.ts
rinatkhaziev: Does this need Windows native and WSL2 testing?

sjinks: No, this was stolen from Lando:

https://github.com/Automattic/lando-cli/blob/6ca2668273b1c4d07bc293d0c49f1a56646c03ac/lib/user.js#L20

rinatkhaziev:
Tested under all platforms (except Linux), works as expected.

rinatkhaziev (approved):

PR #2245, opened 2025-02-24
Add new parameter `allow-all-file-types` for Media import to bypass file type validations

aagam-shah:
## Description

This PR tries to add a new optional parameter `allow-all-file-types` which will bypass file type validations when performing a Media import.

It also updates the older parameters from `camelCase` to `kebab-case`. For backward compatibility, we are still allowing older `camelCase` parameters while introducing the newer ones.

## Steps to Test
1. Check out PR.
1. Run `npm run build`
2. Run `npm link` so that your locally built lib is used
3. Create a zip file that contains a file with unsupported file format
1. Run `./dist/bin/vip import media @SITE_ID ZIP_FILE_LINK --allow-all-file-types`
1. Verify that the import was successful

## Post-Release
- Update the CLI documentation with the new supported parameters

### DIFF
--- a/src/bin/vip-import-media.js
+++ b/src/bin/vip-import-media.js
@@ -106,17 +106,30 @@ Are you sure you want to import the contents of the URL?
 		'Check the status of a currently running media import or retrieve an error log of the most recent media import.'
 	)
 	.command( 'abort', 'Abort the media import currently in progress.' )
-	.option( 'exportFileErrorsToJson', 'Format the error log in JSON. Default is TXT.' )
+	.option( 'exportFileErrorsToJson', 'Alias for --export-file-errors-to-json. (deprecated)' )
+	.option( 'export-file-errors-to-json', 'Format the error log in JSON. Default is TXT.' )
+	.option( 'saveErrorLog', 'Alias for --save-error-log. (deprecated)' )
 	.option(
-		'saveErrorLog',
+		'save-error-log',
 		'Skip the confirmation prompt and download an error log for the import automatically.'
 	)
+	.option( 'overwriteExistingFiles', 'Alias for --overwrite-existing-files. (deprecated)', false )
 	.option(
-		'overwriteExistingFiles',
+		'overwrite-existing-files',
 		'Overwrite existing files with the imported files if they have the same path and file name.',
 		false
 	)
-	.option( 'importIntermediateImages', 'Include intermediate image files in the import.', false )
+	.option(
+		[ 't', 'allow-all-file-types' ],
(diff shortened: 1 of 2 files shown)


src/bin/vip-import-media.js
```diff
@@ -106,17 +106,30 @@ Are you sure you want to import the contents of the URL?
...
+	.option( 'overwriteExistingFiles', 'Alias for --overwrite-existing-files. (deprecated)', false )
 	.option(
-		'overwriteExistingFiles',
+		'overwrite-existing-files',
 		'Overwrite existing files with the imported files if they have the same path and file name.',
 		false
 	)
-	.option( 'importIntermediateImages', 'Include intermediate image files in the import.', false )
+	.option(
+		[ 't', 'allow-all-file-types' ],
```
mjangda, line 123: Why `-t`?

Also, since none of the other params have short variants, we should probably skip it for this one as well.

aagam-shah: By default each param has a short variant (_which is not documented_) and that's the first character of the param. For `allow-all-file-types` it was setting to `-A`. In order to avoid confusion with the application parameter which has `-a` as the short hand, I decided to have separate param for it which is not used for this command.

Screenshot 2025-02-24 at 8 06 47 PM

mjangda (approved):

saroshaga:
Let's await feedback on the RfC now that we've posted it there: https://veetoop2.wordpress.com/2025/02/24/rfc-option-to-allow-all-file-type-during-import/

src/bin/vip-import-media.js
```diff
@@ -106,17 +106,30 @@ Are you sure you want to import the contents of the URL?
 		'Check the status of a currently running media import or retrieve an error log of the most recent media import.'
 	)
 	.command( 'abort', 'Abort the media import currently in progress.' )
-	.option( 'exportFileErrorsToJson', 'Format the error log in JSON. Default is TXT.' )
+	.option( 'exportFileErrorsToJson', 'Alias for --export-file-errors-to-json. (deprecated)' )
+	.option( 'export-file-errors-to-json', 'Format the error log in JSON. Default is TXT.' )
```
saroshaga, line 110: Do we need to explicitly add short form options as you've done with `allow-all-file-types` for these too?

saroshaga, line 110: I think we can set default values in the command option, using the property `defaultValue` 

I wish there was one for marking an option as deprecated as well

aagam-shah:
Closing this as we don't require additional flag to bypass file validation

PR #2279, opened 2025-03-13
feat(dev-env): add initial support for `.env` files

sjinks:
## Description

This PR adds initial support for `.env` files.

It requires a custom version of Lando: Automattic/lando-cli#92 (make sure to run `npm ci before testing this PR).

A new `.env` file is created in `~/.local/share/vip/dev-environment/<slug>` when an environment is created. Docker automatically loads it.

For now, the file needs to be updated manually (and then the environment must be restarted).

**UPDATE:** `vip @app.env dev-env create` syntax is supported; in this case, environment variables are fetched from the specified environment.

## Pull request checklist

## New release checklist

## Steps to Test

1. Check out the PR, `npm ci && npm run build`
2. `vip dev-env create --slug=test-env < /dev/null`
3. `echo "FOO=bar" >> ~/.local/share/vip/dev-environment/test-env/.env`
4. `vip dev-env start --slug=test-env`
5. Verify that `vip dev-env shell --slug=test-env -- env | grep FOO` displays `BAR`.

Grab variables from the environment:

1. Check out the PR, `npm ci && npm run build`
2. `vip dev-env create @<app>.<env> < /dev/null`
4. `vip dev-env start --slug=...`
5. Verify that `vip dev-env shell --slug=test-env -- env` shows variables fetched from the env (variables from `vip config envvar get-all` should be there).

### DIFF
--- a/src/lib/dev-environment/dev-environment-core.ts
+++ b/src/lib/dev-environment/dev-environment-core.ts
@@ -599,10 +599,8 @@ async function prepareLandoEnv(
 
 	if ( envVars !== undefined ) {
 		const env: string[] = [];
-		Object.entries( envVars ?? {} ).forEach( ( [ key, value ] ) => {
-			// See https://docs.docker.com/reference/compose-file/services/#env_file-format
-			value = value.replace( /([$"\\])/g, '\\$1' );
-			env.push( `${ key }="${ value }"` );
+		Object.entries( envVars ?? {} ).forEach( ( [ key ] ) => {
+			env.push( `VIP_ENV_VAR_${ key }=` );
 		} );
 
 		await fs.promises.writeFile( envFilePath, env.join( '\n' ) );
@@ -672,10 +670,8 @@ export async function getApplicationInformation(
 			branch,
 			isMultisite,
 			environmentVariables {
-				total
 				nodes {
 					name
-					value
 				}
 			}
 			getIntegrationsDevEnvConfig {
@@ -725,8 +721,8 @@ export async function getApplicationInformation(
 		if ( envData ) {
 			const envVars: Record< string, string > = {};
 			envData.environmentVariables?.nodes?.forEach( envvar => {
-				if ( envvar?.name && envvar?.value ) {
-					envVars[ envvar.name ] = envvar.value;
+				if ( envvar?.name ) {
+					envVars[ envvar.name ] = '';
 				}
 			} );
 
--- a/src/lib/dev-environment/dev-environment-core.ts
+++ b/src/lib/dev-environment/dev-environment-core.ts
@@ -116,12 +116,17 @@ export async function startEnvironment(
 
 	const instancePath = getEnvironmentPath( slug );
 
-	debug( 'Instance path for', slug, 'is:', instancePath );
+	debug( 'Instance path for %s is %s', slug, instancePath );
 
 	const environmentExists = fs.existsSync( instancePath );
 
 	if ( ! environmentExists ) {
 		throw new Error( DEV_ENVIRONMENT_NOT_FOUND );
+	}
+
+	const envFilePath = path.join( instancePath, '.env' );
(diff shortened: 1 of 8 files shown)


src/lib/dev-environment/dev-environment-core.ts
```diff
@@ -585,6 +597,21 @@
 		fs.promises.writeFile( instanceDataTargetPath, instanceDataFile ),
 	] );
 
+	if ( envVars !== undefined ) {
+		const env: string[] = [];
+		Object.entries( envVars ?? {} ).forEach( ( [ key, value ] ) => {
+			// See https://docs.docker.com/reference/compose-file/services/#env_file-format
+			// We don't want interpolation, hence single quotes.
+			// Inner quotes must be escaped.
+			value = value.replace( /'/g, "\\'" );
```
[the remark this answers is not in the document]
sjinks: This does not work. `a\b` is treated as `a\b`, and `a\\b` will be `a\\b`.

However, `a\\'b` fails, and `a\\\'b` results in `a\\'b` instead of `a\'b`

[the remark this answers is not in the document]
sjinks: Will use a different approach: enclose in double quotes and escape `\`, `$`, and `"`.

src/lib/dev-environment/dev-environment-core.ts
```diff
@@ -646,6 +671,13 @@ export async function getApplicationInformation(
 			type,
 			branch,
 			isMultisite,
+			environmentVariables {
+				total
+				nodes {
+					name
+					value
```
mjangda: We should not fetch the value since some of the values may be secrets (ideally, the API would have a way to mark certain keys as secrets but that doesn't exist yet).

mjangda:
Will this work with our `vip_get_env_var()` function? Or do we need to write out the values as constant in the specifically named format?

src/lib/dev-environment/dev-environment-core.ts
sjinks: I've copied the logic from https://github.com/Automattic/vip-cli/blob/b87f4d7c1811f943aa9cb3f3d03e23ab3d14ab87/src/lib/envvar/api-get-all.ts#L10-L25 (`vip config envvar get-all`). It does not do any filtering, either.

rinatkhaziev: I agree with @mjangda , I don't thing setting the values on create/update will be safe. We can ask them whether they want values or no, but setting them on their behalf without asking is tricky: 
1. Purely security implications (maybe the didn't want that key to written out in the first place)
2. Getting some service API key that does some remote call and triggers some actions  - may result in an unpleasant surprise, like sending test data to production endpoints.

sjinks: Makes sense… OK, I will revert a9fe418 and company.

mjangda: I would keep the feature; it's actually helpful. Just don't fetch the `value`.

sjinks: @mjangda I would like to clarify: we fetch only variable names and set those variables to empty values, is that correct?

sjinks:
> Will this work with our vip_get_env_var() function?

I think so. This is handled in another PR, Automattic/vip-container-images#1118: All variables that begin with `VIP_ENV_VAR_`, will be added as PHP constants into `wp-config.php` and will be available for the `vip_get_env_var()` function.

src/lib/dev-environment/dev-environment-core.ts
sjinks: Fixed in 63092ca.

mjangda: > we fetch only variable names and set those variables to empty values, is that correct?

Yes, and ideally they can set the values themselves after.

rinatkhaziev (approved):

PR #2327, opened 2025-04-18
Add ability to run WP-CLI commands over SSH

luismulinari:
## Description

This PR introduces a new feature for executing WP-CLI commands via SSH instead of using websockets.

## Pull request checklist

## New release checklist

## Steps to Test

1. Check out PR.
2. Run `npm run build`
3. `node ./dist/bin/vip-wp.js <site> --yes -- user list`


src/commands/wp.ts
```diff
@@ -0,0 +1,190 @@
...
+	private track: CommandTracker;
+
+	constructor( app: App, env: AppEnvironment ) {
+		this.app = app;
+		this.env = env;
+
+		this.track = makeCommandTracker( 'wp', {
+			app: this.app.id,
+			env: this.env.id,
+			executionType: 'ssh',
```
mjangda: I think this needs to be snake case:

```
  @automattic/vip:analytics:clients:tracks Error: Invalid prop name detected: executionType -- this event will be rejected during ETL +10s
```

```suggestion
			execution_type: 'ssh',
```

src/bin/vip-wp.js
```diff
@@ -368,6 +370,12 @@ commandWrapper( {
 
 		let countSIGINT = 0;
 
+		if ( ! opts.env.isK8sResident ) {
```
mjangda: Should we do an explicit check that it's a WP Cloud site?

src/commands/wp.ts
```diff
@@ -0,0 +1,190 @@
...
+	} ) {
+		return new Promise( ( resolve, reject ) => {
+			const conn = new ssh2.Client();
+
+			conn
+				.on( 'ready', () => {
+					conn.exec(
+						`GUID=${ guid } INPUT_TOKEN=${ inputToken } VERSION=${ pkg.version }`,
+						( err, stream ) => {
+							if ( err ) throw err;
```
abdullah-kasim:
```suggestion
							if ( err ) {
							   reject( err );
							   return;
							 }
```
Would this be better? I don't think `throw err` here would cause the promise to reject but I might be wrong, and ssh2 would pass the thrown error to `on( 'error' )`

If ssh2 does pass the error to `on( 'error' )`, maybe it's worth adding a comment here as it's not obvious, at least for me. If ssh2 does pass the error to `on( 'error' )`, I wonder if all libraries that rely on event works that way - it's a nice way to ensure all errors are handled in one place for sure!

luismulinari: Good catch!

100%!

src/bin/vip-wp.js
luismulinari: Great idea!

Done: https://github.com/Automattic/vip-cli/pull/2327/commits/3ab374e0e888348265b7f96d1a0807d9974e76dc#diff-eaa8488f5301b9eb6525750493ddf8fdfbe68a5e1fe7da1866156637722920ddR374

src/bin/vip-wp.js
```diff
@@ -29,6 +30,8 @@ const appQuery = `id, name,
 	appId
 	type
 	name
+	isK8sResident
```
mjangda: Minor: Don't need this field anymore:

```suggestion
```

src/commands/wp.ts
```diff
@@ -0,0 +1,255 @@
+/**
```
mjangda: Minor: Should we give this file a more specific name, like `wp-ssh.ts`?

src/commands/wp.ts
```diff
@@ -0,0 +1,255 @@
...
+				error: 'no_app_env_id',
+				message: 'No app or env ID provided',
+			} );
+
+			throw new Error( 'No app ID or environment ID provided' );
+		}
+
+		debug( "Requesting SSH authentication for command '%s'", command );
+
+		const sshAuth = await getSSHAuthForCommand( this.app.id, this.env.id, command );
```
mjangda: Do we want this in a try-catch to handle failures more gracefully?

src/commands/wp.ts
```diff
@@ -0,0 +1,255 @@
...
+
+			console.error( `${ chalk.red( 'Error:' ) } ${ message }` );
+
+			await this.track( 'error', {
+				guid: data?.command.guid,
+				error: 'ssh_command_failed',
+				message: 'Error executing command over SSH',
+			} );
+
+			throw new Error( message );
```
mjangda: Instead of throwing, should we just be exiting?

```suggestion
			process.exit( 1 );
```

src/commands/wp.ts
```diff
@@ -0,0 +1,255 @@
...
+			env: this.env.id,
+			execution_type: 'ssh',
+		} );
+	}
+
+	public async run( command: string ): Promise< void > {
+		if ( ! this.app.id || ! this.env.id ) {
+			await this.track( 'error', {
+				error: 'no_app_env_id',
+				message: 'No app or env ID provided',
```
mjangda: In `src/bin/vip-wp.js`, we also include a stripped down version of the command. Should we do that here as well? This can be done in a separate PR.

mjangda (approved):
Noted some minor things but overall, this looks great.

PR #2398, opened 2025-06-04
feat(dev-env): add `vip dev-env envvar` commands

sjinks:
## Description

This PR adds the following commands:
- `vip dev-env envvar delete [--slug slug] <NAME>`
- `vip dev-env envvar get-all [--slug slug] [--format csv|table|json]`
- `vip dev-env envvar get [--slug slug] <NAME>`
- `vip dev-env envvar list [--slug slug] [--format csv|table|json]`
- `vip dev-env envvar set [--slug slug] [--from-file file] <NAME> [VALUE]`

They mostly mimic their `vip dev-env config envvar` counterparts (except for confirmations).

## Pull request checklist

## New release checklist

## Steps to Test

TBD.


src/bin/vip-dev-env-envvar-delete.js
```diff
@@ -0,0 +1,85 @@
...
+		description: 'Delete the environment variable "MY_VARIABLE" from the environment.',
+	},
+];
+
+async function deleteEnvVarCommand( args, opt ) {
+	debug( 'args: %o, opt: %o', args, opt );
+
+	const slug = await getEnvironmentName( opt );
+	const trackingInfo = getEnvTrackingInfo( slug );
+	const name = args[ 0 ]?.trim() ?? '';
```
rebeccahum, line 31: Do we still want this to proceed if `name` is `''` value??

src/bin/vip-dev-env-envvar-get.js
```diff
@@ -0,0 +1,72 @@
...
+	},
+];
+
+async function getEnvVarsCommand( args, opt ) {
+	debug( 'args: %o, opt: %o', args, opt );
+
+	const slug = await getEnvironmentName( opt );
+	const trackingInfo = getEnvTrackingInfo( slug );
+
+	const name = args[ 0 ]?.trim() ?? '';
```
rebeccahum, line 32: Similar to earlier, do we still want it to run if `name` is `''`?

rebeccahum (changes requested):

src/bin/vip-dev-env-envvar-delete.js
sjinks: `args[0]` is always defined because of `requiredArgs: 1`, but the TypeScript checker does not understand this and insists that it can be `undefined`. That's why we need `?? ''` part to make the analyzer understand that `name` is a `string` :-)

We can still have an empty argument (if the user explicitly passes an empty string). However, in this case, there will be no match, and the user will see a warning message.

sjinks:
```
$ vip dev-env envvar delete ""
The environment variable "" does not exist
```

src/bin/vip-dev-env-envvar-get.js
sjinks: Same as above.

```
$ vip dev-env envvar get ""
The environment variable "" does not exist
```

sjinks: We can add a call to [`validateNameWithMessage()`](https://github.com/Automattic/vip-cli/pull/2398/files#diff-175dc8dc89643b9069c2d0973e5aca1d8225fb2f8cbd4857ce5feeed062bfe64R51) if you think this is necessary :-)

I did not want to add extra checks because:
  - The operation would fail anyway;
  - Local reads are cheap.

rebeccahum (approved):

sjinks:
@yolih, how would you like me to proceed? Should I release this feature and you write the docs later, or should I wait for the docs?

yolih:
Yes please! Go ahead and release. I can test and document as soon as it is available in the next VIP-CLI version.

PR #2451, opened 2025-07-07
Update to the description of the  `--wordpress` option for `dev-env`

yolih:
## Description

Updated the description of the `--wordpress` option in the help menu for `vip dev-env create` and `vip dev-env update` to include the recently introduced `latest` value option. 

### Changed

- Updated the description of the accepted values for the `--wordpress` option in `vip dev-env create` and `vip dev-env update` to include `latest`.

## Pull request checklist

## New release checklist

## Steps to Test

Observe the results of the `--help` display for impacted commands:

- `vip dev-env create`
- `vip dev-env update`

### DIFF
--- a/src/lib/dev-environment/dev-environment-cli.ts
+++ b/src/lib/dev-environment/dev-environment-cli.ts
@@ -976,7 +976,7 @@ export function addDevEnvConfigurationOptions( command: Args ): Args {
 	return command
 		.option(
 			'wordpress',
-			'Manage the version of WordPress. Accepts a string value for major versions (6.x). Defaults to the most recent version of WordPress.',
+			'Manage the version of WordPress. Accepts a string value for major versions (6.x) or "latest" (default).',
 			undefined,
 			processVersionOption
 		)
--- a/src/lib/dev-environment/dev-environment-cli.ts
+++ b/src/lib/dev-environment/dev-environment-cli.ts
@@ -976,7 +976,7 @@ export function addDevEnvConfigurationOptions( command: Args ): Args {
 	return command
 		.option(
 			'wordpress',
-			'Manage the version of WordPress. Accepts a string value for major versions (6.x) or "latest" (default).',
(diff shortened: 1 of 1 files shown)


sjinks:
It's a bit more complicated :-(

We have a file with WordPress versions available at: https://github.com/Automattic/vip-container-images/blob/master/wordpress/versions.json. This file is updated dynamically. When you don't specify a version in `vip dev-env create/update` commands, the system picks the very first version from that file. It is *usually* the latest version, but sometimes it is a prerelease that we want users to test.

When `latest` is specified, the versions are sorted from the latest to the oldest, and we pick the first version that is not `prerelease`. This will be the most recent stable version available.

yolih:
@sjinks Of course it's more complicated. I should have known! 😆 

So it seems that the previous description for the option has been incorrect for quite a while.
> Defaults to the most recent version of WordPress.

That seemed like a reasonable and expected behavior for the `--wordpress` option.
And I understand the logic of this:
>sometimes it is a prerelease that we want users to test.

But how would a customer know that's happening?
Let's say they don't read instructions and they spin up a local env without specifying a version of WordPress. While developing, they keep noticing a weird behavior but they can't figure out what it is. I think it's more likely that they blame the weird behavior on a possible bug in the VIP Local Development Environment than the possibility that they are running a prerelease version of WordPress. 

But if this behavior has been this way for ages, then I guess I just need to work with it. 🙂 

Does this description read truth-y?
>Manage the version of WordPress. Accepts a string value for major versions (6.x) or "latest". Defaults to the recommended version of WordPress for development.

sjinks:
> Does this description read truth-y?

Yes!

> But how would a customer know that's happening?

Alas :-( They can run `vip dev-env info --extended` and see what they have.

I know that this is far from ideal, but we have what we have.

sjinks (approved):

yolih:
@sjinks Thank you! 🖤

PR #2769, opened 2026-03-11
Improve command description for `import media`

yolih:
## Description

The VIP-CLI command `vip import media` can only be run against the production environment of an application. This PR restores that clarification which was unintentionally lost about a year ago. 

### Changed

- Edited the command description for `import media` to clarify that it can only be run against a production environment.

## Steps to Test

1. Check out PR.
2.  Observe the results of the `--help` display for impacted command `vip import media`.

### DIFF
--- a/src/bin/vip-import.js
+++ b/src/bin/vip-import.js
@@ -8,7 +8,7 @@ command()
 	.command( 'validate-sql', 'Validate a local SQL database file prior to import.' )
 	.command(
 		'validate-files',
-		'Validate that the directory structure and contents of a local media file directory can be successfully importedß.'
+		'Validate that the directory structure and contents of a local media file directory can be successfully imported.'
 	)
 	.command(
 		'media',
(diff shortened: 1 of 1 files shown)


src/bin/vip-import.js
```diff
@@ -8,11 +8,11 @@ command()
 	.command( 'validate-sql', 'Validate a local SQL database file prior to import.' )
 	.command(
 		'validate-files',
-		'Validate the directory structure and contents of a local media file directory prior to archiving and uploading it to a publicly accessible URL.'
+		'Validate that the directory structure and contents of a local media file directory can be successfully importedß.'
```
sjinks:
```suggestion
		'Validate that the directory structure and contents of a local media file directory can be successfully imported.'
```

@yolih, I guess this "scharfes S" is a typo?

yolih: What the heck. 100% a typo! I wish I could blame this on a "cat on the keyboard" but I don't have a cat. 
Thank you for your sharp eyes, @sjinks !

sjinks (approved):

PR #2869, opened 2026-06-04
Add Step-up rechallenge implementation

rinatkhaziev:
## Description

This pull request implements step-up rechallenge support end-to-end for elevated-permission operations, and wires it into the shipped CLI experience.

### Shipped behavior

- Adds a rechallenge flow in the GraphQL request path so protected mutations can trigger step-up verification and then retry automatically with elevated authorization.
- Adds operation-scoped elevated-token caching with keychain-backed persistence (scoped by API host), so repeated protected operations can reuse a valid elevation window.
- Clears elevated-token cache on both login and logout so cached elevation cannot carry across user identity changes.
- Introduces the `vip defensive-mode` command group with `enable`, `disable`, and `configure` subcommands.
- Adds production safety guards for defensive-mode mutations, including explicit confirmation requirements in non-interactive contexts.
- Adds non-interactive rechallenge fail-fast behavior by default, with explicit opt-in wait support via `--rechallenge-wait` / `VIP_RECHALLENGE_WAIT=1`.
- Extracts API host constants into a dedicated module to support rechallenge integration without circular dependency regressions.

### Test coverage included

- Adds unit tests for rechallenge client, flow orchestration, Apollo link integration, token cache behavior, defensive-mode API and commands, API-chain retries, and logout/login cache clearing paths.

### Added

- Added new `vip defensive-mode` commands to enable, disable, and configure defensive mode for an environment.
- Added step-up verification support for protected actions, including automatic retry after verification succeeds.

### Changed

- Non-interactive runs now fail fast when step-up verification is required, unless you explicitly opt in to wait with `--rechallenge-wait` or `VIP_RECHALLENGE_WAIT=1`.
- Production defensive-mode mutations in non-interactive workflows now require explicit confirmation bypass via `--skip-confirmation`.

### Fixed

- Login and logout now clear cached elevated permissions to prevent reuse across account switches on the same machine.

## Pull request checklist

## New release checklist

## Steps to Test

Example:

1. Check out PR.
1. Run `npm run build`
1. Run `./dist/bin/vip-cookies.js nom`
1. Verify cookies are delicious.


src/lib/defensive-mode/api.ts
```diff
@@ -0,0 +1,119 @@
...
+		mutation: STATUS_MUTATION,
+		variables: {
+			input: { id: input.appId, environmentId: input.envId, enabled: input.enabled },
+		},
+	} );
+	return (
+		result.data as {
+			updateDefensiveModeStatus: { success: boolean; message: string };
+		}
+	 ).updateDefensiveModeStatus;
```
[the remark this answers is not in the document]
rinatkhaziev: Addressed in c9fc8dfc — added `if ( ! result.data )` guard with descriptive error before the cast in both `updateDefensiveModeStatus` (api.ts:87-91) and `updateDefensiveModeConfig` (api.ts:119-123). Apollo's errorLink still handles the typical GraphQL-errors case via `process.exit(1)`; this guard catches the edge where `data` is null without a thrown combined error.

src/lib/defensive-mode/api.ts
```diff
@@ -0,0 +1,119 @@
...
+	}
+	const result = await api.mutate( {
+		mutation: CONFIG_MUTATION,
+		variables: { input: mutationInput },
+	} );
+	return (
+		result.data as {
+			updateDefensiveModeConfig: { success: boolean; message: string };
+		}
+	 ).updateDefensiveModeConfig;
```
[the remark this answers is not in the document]
rinatkhaziev: Same fix landed in c9fc8dfc — `if ( ! result.data )` guard at api.ts:119-123 before the cast in `updateDefensiveModeConfig`.

src/lib/rechallenge/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export class RechallengeUnsupportedVersionError extends RechallengeError {
+	constructor( version: string, scope: string ) {
+		super(
+			`Server requested rechallenge version "${ version }" but this CLI only supports v2. Update vip-cli.`,
+			scope
+		);
+		this.name = 'RechallengeUnsupportedVersionError';
+	}
+}
```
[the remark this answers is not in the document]
rinatkhaziev: Addressed in c9fc8dfc — `RechallengeUnsupportedVersionError` now imports `RECHALLENGE_VERSION` from `./types` and interpolates it: `"... but this CLI only supports ${ RECHALLENGE_VERSION }. Update vip-cli."` (errors.ts:1, 17).

src/lib/rechallenge/link.ts
```diff
@@ -0,0 +1,154 @@
...
+	const errors = result.errors ?? [];
+	for ( const err of errors ) {
+		const ext = ( err.extensions ?? {} ) as Record< string, unknown >;
+		if ( ext.code !== ELEVATED_PERMISSION_ERROR_CODE ) {
+			continue;
+		}
+		const rechallenge = ext.rechallenge as RechallengeExtension | undefined;
+		if ( rechallenge && typeof rechallenge.createSessionPath === 'string' ) {
+			return { rechallenge };
+		}
```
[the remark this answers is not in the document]
rinatkhaziev: Addressed in c9fc8dfc — `extractElevatedPermission` now validates all four required fields (`createSessionPath`, `statusPathTemplate`, `exchangePathTemplate`, `elevatedHeaderName`) as strings before returning the payload (link.ts:47-53). If any are missing, the function returns `null` and the original GraphQL error propagates upstream so errorLink handles it consistently.

__tests__/lib/rechallenge/client.test.ts
```diff
@@ -0,0 +1,141 @@
+import { describe, expect, it, jest, beforeEach } from '@jest/globals';
+
+import http from '../../../src/lib/api/http';
+import * as client from '../../../src/lib/rechallenge/client';
+import { RechallengeHttpError } from '../../../src/lib/rechallenge/errors';
+
```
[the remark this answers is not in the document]
rinatkhaziev: Addressed in c9fc8dfc — added `import type { Response } from 'node-fetch';` (client.test.ts:7), matching the implementation file. `tsc --noEmit` was already clean (Node's `@types/node` provides a global `Response`), but the explicit import makes the dependency unambiguous.

src/bin/vip-defensive-mode-configure.js
```diff
@@ -0,0 +1,259 @@
...
+		input.connectionThresholdPercentage = percentage;
+	}
+
+	const currentConfig = opt.env.defensiveMode?.config?.effective ?? null;
+	if ( currentConfig ) {
+		console.log( chalk.bold( 'Current defensive-mode configuration:' ) );
+		console.log( JSON.stringify( currentConfig, null, 2 ) );
+	}
+	console.log( chalk.bold( 'Proposed defensive-mode configuration:' ) );
+	console.log( JSON.stringify( input, null, 2 ) );
```
sjinks: Maybe make the output more user-friendly?

src/lib/logout.ts
```diff
@@ -1,11 +1,13 @@
 import http from '../lib/api/http';
+import tokenCache from '../lib/rechallenge/token-cache';
 import Token from '../lib/token';
 import { trackEvent } from '../lib/tracker';
 
 export default async (): Promise< void > => {
 	await http( '/logout', { method: 'post' } );
 
 	await Token.purge();
+	await tokenCache.clearAll();
```
sjinks: Should we do the same on login?

Elevated tokens are keyed only by the API host and operation scope. Logout clears them, but direct vip login token replacement does not, so cached elevation can cross user identities on the same machine.

This is not a very likely scenario, but I think it makes sense to apply this hardening.

package.json
```diff
@@ -25,6 +25,10 @@
     "vip-config-software-update": "dist/bin/vip-config-software-update.js",
     "vip-db": "dist/bin/vip-db.js",
     "vip-db-phpmyadmin": "dist/bin/vip-db-phpmyadmin.js",
+    "vip-defensive-mode": "dist/bin/vip-defensive-mode.js",
+    "vip-defensive-mode-configure": "dist/bin/vip-defensive-mode-configure.js",
+    "vip-defensive-mode-disable": "dist/bin/vip-defensive-mode-disable.js",
+    "vip-defensive-mode-enable": "dist/bin/vip-defensive-mode-enable.js",
```
sjinks, line 33: I guess we need to update `src/lib/cli/internal-bin-loader.js` as well.

src/lib/rechallenge/flow.ts
```diff
@@ -0,0 +1,165 @@
...
+		const timer = setTimeout( () => {
+			signal?.removeEventListener( 'abort', onAbort );
+			resolve();
+		}, ms );
+		function onAbort() {
+			clearTimeout( timer );
+			reject( new Error( 'aborted' ) );
+		}
+		signal?.addEventListener( 'abort', onAbort, { once: true } );
+	} );
```
sjinks:
```typescript
import { setTimeout } from 'node:timers/promises';
```

```suggestion
    try {
        await setTimeout( ms, undefined, { signal } );
    } catch ( err ) {
        if ( err.name === 'AbortError' ) {
            throw new Error( 'aborted' );
        } else {
            throw err;
        }
    }
```

src/bin/vip-defensive-mode-configure.js
```diff
@@ -0,0 +1,259 @@
...
+		return true;
+	}
+	if ( [ 'false', 'no', '0', 'off', 'disable', 'disabled' ].includes( normalized ) ) {
+		return false;
+	}
+	return null;
+}
+
+function parsePositiveInt( raw ) {
+	if ( raw === undefined || raw === null || raw === '' ) {
```
sjinks:
```suggestion
	if ( raw === undefined || raw === null || raw === '' || typeof raw === 'boolean' ) {
```

`--connection-threshold-absolute` without parameters is treated like `--connection-threshold-absolute=true`; `Number( true )` coerces to 1. I am not sure if it is OK.

src/bin/vip-defensive-mode-configure.js
rinatkhaziev: Reworked in 5a7ca834: current/proposed values now render as a settings table (via the existing `table()` helper), and the JSON blob was dropped from the production confirm prompt.

src/lib/logout.ts
rinatkhaziev: Applied in 5a7ca834: `vip login` now calls `tokenCache.clearAll()` right after `Token.set()` replaces the stored token. That's the only `Token.set` call site, so login and logout now both clear cached elevation.

package.json
rinatkhaziev: Added in 5a7ca834. Diffed package.json bin keys against the loader — the four `vip-defensive-mode*` entries were the only missing ones.

src/lib/rechallenge/flow.ts
rinatkhaziev: Done in 5a7ca834: the hand-rolled `sleep()` is replaced with `setTimeout` from `node:timers/promises` (with `{ signal }`). The caller already converts any sleep rejection to `RechallengeAbortedError`, so the AbortError branch lives there.

src/bin/vip-defensive-mode-configure.js
rinatkhaziev: Fixed in 5a7ca834: `parsePositiveInt` now rejects booleans (bare flags) and blank/whitespace strings, so `--connection-threshold-absolute` without a value errors instead of coercing to 1. Added a regression test.

src/bin/vip-defensive-mode-configure.js
```diff
@@ -0,0 +1,299 @@
...
+			`Apply the proposed configuration to ${ formatEnvironment( opt.env.type ) } for ${
+				opt.app.name
+			}?`
+		);
+		if ( ! yes ) {
+			await trackEvent( 'defensive_mode_configure_command_cancelled', trackingParams );
+			console.log( 'Command cancelled' );
+			process.exit();
+		}
+	}
```
sjinks: AI insists that this is a bug:

---
`configure` skips the production confirmation guard in non-interactive mode

- **Artifact:** `src/bin/vip-defensive-mode-configure.js:249-262` vs `src/lib/defensive-mode/cli-helpers.ts:30-61` (`guardProductionMutation`, used by enable/disable)
- **Category:** user-workflow
- **Severity:** HIGH
- **Confidence:** high
- **Classification:** Confirmed issue
- **Trigger:** `vip @app.production defensive-mode configure --enabled=false --
- **Intended contract (reviewer clarification 2026-06-18):** 
- **Risk:** enable/disable call `guardProductionMutation`, which hard-errors on production non-interactive runs lacking `--skip-confirmation`. `configure` only prompts when `interactive && !skipConfirmation && env.type==='production'`; in non-interactive mode that condition is false, so it proceeds straight to `updateDefensiveModeConfig` with no acknowledgment required. Because `configure` can set `enabled=false`, it silently disables production defensive mode in CI/automation — the exact unattended case sibling commands refuse. The clarification reinforces this: the missing `--skip-confirmation` gate is the whole point of the unattended safety contract.
- **Evidence:** configure has no `guardProductionMutation` call and no non-interactive refusal branch; enable/disable both call it (enable:36-43, disable:32-39).
- **Suggested fix:** Route `configure` through `guardProductionMutation` (action `'configure'`) before mutating, so unattended (`--non-interactive`, no TTY) production runs **fail fast unless `--skip-confirmation` is passed** — matching enable/disable. Do not add a TTY prompt for unattended mode; require the explicit flag instead.
---

I tell it that the `--non-interactive` flag is designed for unattended/script use, where no TTY is available for prompting.

It responds that the suite's safety model — established by enabling/disabling via `guardProductionMutation` — is therefore: an unattended production mutation must be **explicitly acknowledged with `--skip-confirmation`; otherwise, fail fast**. A prompt is intentionally not the unattended safeguard; the explicit flag is. And that my clarification strengthens the finding — `configure` is the only command that skips that explicit acknowledgment and silently mutates production.

At any rate, we must document the behavior in the code, as this has been flagged twice and will be found in subsequent security reviews.

src/lib/rechallenge/link.ts
```diff
@@ -0,0 +1,160 @@
...
+							}
+						},
+					} );
+				} );
+
+			return () => {
+				cancelled = true;
+				firstSub?.unsubscribe();
+				retrySub?.unsubscribe();
+			};
```
sjinks, line 199: Apollo unsubscribe does not abort the in-flight rechallenge polling flow

- **Artifact:** `src/lib/rechallenge/link.ts:117-121` (runRechallenge called without `signal`); `153-157` (dispose sets `cancelled` only)
- **Category:** resource-lifecycle
- **Severity:** MEDIUM
- **Confidence:** high
- **Classification:** Confirmed issue
- **Trigger:** Consumer unsubscribes / Apollo tears down the operation while rechallenge is polling (process interrupt, `client.stop()`).
- **Risk:** `flow.runRechallenge` supports an `AbortSignal`, but the link never passes one. The dispose closure sets `cancelled=true`, which only suppresses post-resolution work; the polling loop (HTTP status calls), `trackEvent` emissions, and `tokenCache.set` continue in the background until terminal/expiry.
- **Evidence:** `flow.ts:31,83-91` accepts/handles `signal`; `link.ts:117` omits it; dispose cannot interrupt the loop.
- **Suggested fix:** Create an `AbortController` in the Observable, pass `signal` into `runRechallenge`, and call `abort()` in the dispose function.

src/lib/rechallenge/flow.ts
```diff
@@ -0,0 +1,150 @@
...
+			`  If your browser did not open, copy and paste the URL above. Expires at ${ expiresIso }.`
+		);
+	} else {
+		console.warn(
+			`Step-up verification required for ${ requestedOperation }. ` +
+				`Complete it at: ${ verificationUrl } (expires at ${ expiresIso }).`
+		);
+	}
+
+	const interval = Math.max( session.pollIntervalSeconds, 0 ) * 1000;
```
sjinks: Is `0` a good value? It will cause a tight status-poll loop.

Another potential issue is that if `session.pollIntervalSeconds` is `NaN`, `Math.max()` will return `NaN`, not 0.

We need smth like `Math.max( Number(session.pollIntervalSeconds) || DEFAULT, DEFAULT ) * 1000

src/lib/rechallenge/flow.ts
```diff
@@ -0,0 +1,150 @@
...
+		console.warn( `  Opened ${ chalk.cyan( verificationUrl ) }` );
+		console.warn(
+			`  If your browser did not open, copy and paste the URL above. Expires at ${ expiresIso }.`
+		);
+	} else {
+		console.warn(
+			`Step-up verification required for ${ requestedOperation }. ` +
+				`Complete it at: ${ verificationUrl } (expires at ${ expiresIso }).`
+		);
+	}
```
sjinks, line 84: When automation runs a mutation (e.g., `defensive-mode disable --skip-confirmation --non-interactive`) that the server answers with `elevated-permission-required`, with no human to complete browser verification, the flow polls until `expiresAt` (potentially many minutes) before failing — the automated command blocks instead of failing fast with an actionable error.

Maybe it makes sense to throw immediately with guidance to run interactively, rather than polling.

__tests__/bin/vip-defensive-mode-configure.js
```diff
@@ -0,0 +1,148 @@
+import { describe, expect, it, jest, beforeEach } from '@jest/globals';
```
sjinks, line 1: Test gap: `configure` production guard behavior is unverified

- **Artifact:** `__tests__/bin/vip-defensive-mode-configure.js` (no production/non-interactive guard case)
- **Category:** verification-gap
- **Severity:** MEDIUM
- **Confidence:** high
- **Classification:** Test gap
- **Risk:** enable/disable tests assert non-interactive production refusal, but configure has no equivalent test, so the divergence in Finding 1 is unguarded and could regress silently.
- **Evidence:** configure test file covers flag parsing/tracking only; no `env.type:'production'` + `nonInteractive:true` assertion.
- **Suggested fix:** Add a test asserting configure refuses (or requires `--skip-confirmation`) on production in non-interactive mode once Finding 1 is fixed.

src/lib/rechallenge/flow.ts
rinatkhaziev: Good call-out on unattended automation blocking until `expiresAt`.

One wrinkle before we change this: the non-interactive path is currently an *intentional, tested* contract — `runRechallenge` prints the verification URL and polls even when non-interactive, so a human can complete step-up on another device within the window (see `flow.test.ts` → "does not call open() when interactive=false"). Flipping it to fail-fast reverses that deliberately, so I'd rather not change it unilaterally.

Two options I'd like the team's read on:

1. **Opt-in wait (preferred):** default to fail-fast in non-interactive mode — throw immediately with guidance to re-run interactively — and gate today's print-URL-and-poll behavior behind an explicit signal (e.g. `VIP_RECHALLENGE_WAIT=1` or `--rechallenge-wait`). Safe-by-default for CI; still supports the verify-on-another-device flow when explicitly requested.
2. **Bounded wait:** keep polling in non-interactive mode but cap it with a timeout (e.g. `--rechallenge-timeout`, default ~60s) instead of waiting until `expiresAt`. Less explicit, no new opt-in concept.

Both line up with the fail-fast-in-automation stance we're applying to `configure`. Holding implementation here until we agree on a direction — leaning toward (1).

src/bin/vip-defensive-mode-configure.js
rinatkhaziev: Addressed in 4552c397. `configure` now routes through the same `guardProductionMutation( opt, 'configure', … )` as enable/disable (vip-defensive-mode-configure.js:256). On a production env in non-interactive mode without `--skip-confirmation` it hard-errors and `process.exit(1)` before any mutation (cli-helpers.ts:41-50) — no TTY prompt is added for unattended mode; the explicit `--skip-confirmation` flag is required, matching the enable/disable contract. The unattended behavior is documented inline at configure.js:253-255. Regression tests in the companion thread.

__tests__/bin/vip-defensive-mode-configure.js
rinatkhaziev: Covered in 4552c397. `__tests__/bin/vip-defensive-mode-configure.js` now asserts both sides of the guard: "refuses production mutation in non-interactive mode without --skip-confirmation" (expects the process-exit, asserts `updateDefensiveModeConfig` is not called) and "allows production mutation … with --skip-confirmation". Mirrors the enable/disable production-guard tests.

src/lib/rechallenge/link.ts
rinatkhaziev: Addressed in 4552c397. The Observable now owns an `AbortController` (link.ts:85), passes its `signal` into `runRechallenge` (link.ts:122), and calls `abortController.abort()` in the dispose closure (link.ts:159). On unsubscribe the in-flight `sleep` rejects, so the poll loop, its `trackEvent` emissions, and the `tokenCache.set` stop instead of running to expiry. `flow.ts` already honored the signal (flow.ts:93 top-of-loop check, flow.ts:98 abortable sleep); the `flow.test.ts` "aborts when the abort signal fires" case covers that path.

src/lib/rechallenge/flow.ts
rinatkhaziev: Addressed. The interval is floored: `Math.max( Number.isFinite( requestedInterval ) ? requestedInterval : MIN_POLL_INTERVAL_SECONDS, MIN_POLL_INTERVAL_SECONDS ) * 1000` with `MIN_POLL_INTERVAL_SECONDS = 2` (flow.ts:23, 76-81), so 0 / NaN / missing can't tight-loop or yield NaN — the `Math.max( Number(…) || DEFAULT, DEFAULT )` shape you suggested. Floor logic in 4552c397; test "floors an unusable poll interval instead of tight-looping or producing NaN" (flow.test.ts:85, added in 01f781ad) covers the NaN and sub-floor cases.

src/lib/rechallenge/flow.ts
rinatkhaziev: Implemented option 1 (opt-in wait) in 154a75fe. Non-interactive runs now fail fast by default: `runRechallenge` throws `RechallengeInteractionRequiredError` before creating a session, with guidance to re-run interactively. The print-URL-and-poll behavior is gated behind `--rechallenge-wait` or `VIP_RECHALLENGE_WAIT=1` (mirroring `--non-interactive` / `VIP_NON_INTERACTIVE`), so the verify-on-another-device flow still works when explicitly requested. The flag is declared on enable/disable/configure; added `shouldWaitForRechallenge` unit tests plus a fail-fast test asserting no session is created. Resolving.

src/lib/rechallenge/flow.ts
```diff
@@ -0,0 +1,181 @@
...
+	/* eslint-enable no-await-in-loop */
+}
+
+export function isInteractiveContext( argvOrFlags: string[] = process.argv ): boolean {
+	if ( process.env.VIP_NON_INTERACTIVE === '1' ) {
+		return false;
+	}
+	if ( argvOrFlags.includes( '--non-interactive' ) ) {
```
rebeccahum: Do we need to check the alias too ?

rinatkhaziev: you mean the shorthand? there ain't shorthand.

src/lib/rechallenge/errors.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export class RechallengeInteractionRequiredError extends RechallengeError {
+	constructor( scope: string ) {
+		super(
+			`Step-up verification is required for ${ scope }, but this is a non-interactive session. ` +
+				'Re-run the command interactively, or pass --rechallenge-wait (or set VIP_RECHALLENGE_WAIT=1) ' +
+				'to print the verification URL and wait for you to complete it on another device.',
+			scope
```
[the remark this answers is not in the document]
sjinks: I am not sure what to do with this one. This is the third reviewer who flags this issue.

I have verified that no other command triggers rechallenge, but I understand why reviews flag this issue.

@rinatkhaziev, what do you think?

src/lib/rechallenge/errors.ts
```diff
@@ -0,0 +1,66 @@
...
+		);
```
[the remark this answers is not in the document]
sjinks: Same as https://github.com/Automattic/vip-cli/pull/2869#discussion_r3509238699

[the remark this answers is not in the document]
sjinks: Duplicate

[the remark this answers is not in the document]
sjinks: I heard you. Enough.

src/lib/rechallenge/flow.ts
rebeccahum: ok cool 👍

rebeccahum (approved):

PR #2905, opened 2026-06-23
fix(upload): harden fetch upload streams

sjinks:
## Summary

- replaces upload progress `PassThrough`/`data` listeners with `Transform` streams that count bytes while forwarding them to `undici.fetch`
- removes the unsupported `Expect` header before dispatching requests through `undici.fetch`
- preserves useful low-level `fetch` cause details when `fetchWithRetry` exhausts retries
- adds regression coverage for:
  - single-PUT uploads streaming exactly `Content-Length` bytes
  - `TypeError: fetch failed` with an underlying `write ECONNRESET` cause
  - incoming `Expect: 100-continue` headers

## Why

Follow-up from PLTFRM-2491 / the custom deploy incident thread. After v4.0.7, the customer no longer sees the disturbed-stream or missing-`duplex` errors, but later debug traces exposed two more local undici failure modes:

```text
[cause]: NotSupportedError: expect header not supported
code: 'UND_ERR_NOT_SUPPORTED'
```

and on the single-PUT path:

```text
[cause]: RequestContentLengthMismatchError: Request body length does not match content-length header
code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
```

`undici.fetch` cannot send the `Expect` header, so leaving that header in presigned options guarantees a local failure before the request reaches S3.

For the content-length mismatch, the upload body was piped through a `PassThrough` with a `data` listener for progress. That can put the stream into flowing mode before undici attaches its reader, allowing progress accounting to consume bytes that are no longer delivered to fetch. Using a `Transform` keeps progress accounting in the same stream pipeline as the upload body and preserves backpressure/byte delivery.

If future retries still exhaust with a lower-level transport cause, the final error becomes more actionable, e.g. `fetch failed: write ECONNRESET`, without including presigned URLs or other request details.

## Testing

- `npm run jest -- __tests__/lib/client-file-uploader.js __tests__/lib/fetch-with-retry-undici.js --runInBand` (ran all non-e2e suites: 56 passed / 593 tests)
- `npm run check-types`
- `npm run cmd:lint -- src/lib/client-file-uploader.ts __tests__/lib/client-file-uploader.js`
- `npm run build`

### DIFF
--- a/__tests__/lib/client-file-uploader.js
+++ b/__tests__/lib/client-file-uploader.js
@@ -15,6 +15,7 @@ import {
 	getFileMeta,
 	getPartBoundaries,
 	parseEtagHeader,
+	uploadImportFileToS3,
 	uploadParts,
 } from '../../src/lib/client-file-uploader';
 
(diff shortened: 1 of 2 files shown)


rinatkhaziev (approved):
:intensifies-more:

PR #2918, opened 2026-06-25
Add Edge Workers commands with safe deploy and explicit enable

alexcriss:
## Description

Add a complete `vip edge-workers` command suite for safely scaffolding, validating, deploying, inspecting, and managing WASM Edge Workers.

The workflow is intentionally review-first:

- `init` scaffolds an Edge Workers project with exact supported AssemblyScript/SDK dependencies and refuses destructive initialization.
- `new <name>` creates a validated worker manifest and starter source.
- `build [name]` / `build --all` compile deterministic WASM artifacts.
- `validate <name>` / `validate --all` perform server-side dry-run validation without uploading.
- `deploy <name>` / `deploy --all` prepare and print the complete plan before any remote mutation, then create or update by name.
- `list` and `get <name>` inspect deployed workers; source is fetched only when explicitly requested.
- `enable`, `disable`, and `delete` manage deployed workers with production/destructive confirmation safeguards.

This revision also adds explicit activation to deploy through `--enable`, while keeping upload-only deployment as the default.

### Deploy/activation safety contract

- The corresponding platform API must land and deploy first; it guarantees that every create is inactive and accepts no active input.
- `deploy --enable` defaults to `false` and is not a confirmation bypass.
- Without `--enable`, creates and inactive updates remain inactive.
- With `--enable`, a create or inactive update uploads first and enables only after a successful upload.
- An already-active update becomes live on upload and skips a redundant enable request.
- Production shows the full deployment plan and uses one confirmation that explicitly says “Deploy” or “Deploy and enable” and names every target.
- `--all` prepares every worker before remote writes, then applies sequentially with accurate completed/failed/not-attempted reporting.
- If enable fails or times out, the final active state is unknown. The command reports the last confirmed upload state and requires `get`/`list` verification; it never claims the worker remained inactive.
- There is no automatic retry, rollback, disable, delete, or new non-interactive bypass.
- Success telemetry contains only low-cardinality counts/booleans; source, paths, worker names, and raw errors are excluded.

The operator guide documents the inactive-create dependency, location preserve/null-clear semantics, source storage behavior, live active updates, partial failures, `--enable`, and the required non-production lifecycle.

### Added

- Added `vip edge-workers` commands to scaffold, build, validate, deploy, inspect, enable, disable, and delete Edge Workers.
- Added `vip edge-workers deploy --enable` to explicitly enable newly created or inactive workers after a successful upload.

### Changed

- Edge Worker deploys are upload-only by default and now preview current and intended active state before mutation.
- Edge Worker deployment failures now distinguish upload failures from ambiguous enable failures and provide state-verification guidance.

## Pull request checklist

- [x] No new environmental variables.
- [x] Updated the Edge Workers operator documentation.
- [ ] Manually tested the complete lifecycle in an approved non-production environment (blocked until the platform inactive-create guarantee is deployed).
- [x] Followed the pull request checklist and documented release/deployment ordering.
- [x] Added/updated automated tests.

## New release checklist

- [x] Automated tests pass locally.
- [ ] The preparing-for-release checklist is completed.

## Verification

Run with Node 24.19.0:

```sh
npm test
npm run build
npm run smoke:release
NODE_ENV=test DO_NOT_TRACK=1 node dist/bin/vip-edge-workers-deploy.js --help
```

Local results:

- Full test gate: 89 suites passed; 929 tests passed; 1 existing todo. Lint and typecheck passed with one test-only sequential-await warning.
- Focused Edge Workers deploy/enable gate: 4 suites and 68 tests passed.
- Build: 203 files compiled.
- Release smoke: 14/14 checks passed.
- Built help shows `--enable` with `(default: false)` and required no credentials or browser.

## Steps to Test

Do not use a production target. After the platform inactive-create guarantee is deployed, use an approved non-production alias:

1. Build this branch:

   ```sh
   npm run build
   alias vipdev="node $(pwd)/dist/bin/vip.js"
   ```

2. Scaffold and review a project:

   ```sh
   mkdir edge-workers-test
   cd edge-workers-test
   vipdev edge-workers init
   cd edge-workers
   npm install
   vipdev edge-workers new security-headers --location starts_with:/api/
   # Implement and review workers/security-headers/assembly/index.ts.
   vipdev edge-workers build security-headers
   ```

3. Validate and perform the default upload-only create. Confirm `list`/`get` report the worker inactive:

   ```sh
   vipdev @example-app.develop edge-workers validate security-headers
   vipdev @example-app.develop edge-workers deploy security-headers
   vipdev @example-app.develop edge-workers list
   vipdev @example-app.develop edge-workers get security-headers --source
   ```

4. Exercise separate enable, disable, and deploy-with-enable:

   ```sh
   vipdev @example-app.develop edge-workers enable security-headers
   vipdev @example-app.develop edge-workers disable security-headers
   # Update the reviewed source while inactive.
   vipdev @example-app.develop edge-workers deploy security-headers --enable
   ```

5. While the worker is already active, update and deploy again with `--enable`. Verify the plan says it remains active and no redundant enable request is made.

6. Verify location tri-state behavior: omit `location` to preserve the stored rule, then set it to `null` to clear the rule and apply to all requests.

7. Disable the worker, then confirm permanent deletion:

   ```sh
   vipdev @example-app.develop edge-workers disable security-headers
   vipdev @example-app.develop edge-workers delete security-headers
   ```

If an enable request fails or times out at any point, stop: the final active state is unknown. Reconcile it with `edge-workers get <name>` or `edge-workers list` before taking another action. Do not retry, roll back, disable, or delete automatically.


src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,248 @@
...
+ * mutations keyed by `environmentId`. Worker names are unique per environment, so
+ * the CLI reconciles create-vs-update by matching on `name`.
+ *
+ * NOTE: these types are hand-written rather than codegen'd because the edge
+ * worker schema is not part of the public schema bundle the codegen runs against.
+ */
+
+import gql from 'graphql-tag';
+
+import API from '../../lib/api';
```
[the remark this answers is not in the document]
rinatkhaziev: Verified on the current head: no code change is needed here. From src/lib/api/edge-workers.ts, ../../lib/api normalizes to src/lib/api, and TypeScript resolves the sibling src/lib/api.ts file. The same established import is used by the other src/lib/api/* modules; check-types, the 203-file build, and the full Jest suite pass at 7ea4cd82.

src/lib/edge-workers/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** Read the entry source of a worker, for storing alongside the binary. */
+export function readWorkerSource( worker: DiscoveredWorker ): string | undefined {
+	const entry = path.resolve( worker.dir, worker.manifest.entry );
+	try {
+		return fs.readFileSync( entry, 'utf8' );
+	} catch {
+		return undefined;
+	}
+}
```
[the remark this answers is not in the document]
rinatkhaziev: Addressed by the path-containment hardening now on the branch. readWorkerSource() validates the entry with resolvePathWithin()/resolveExistingPathWithin(), reports missing or unreadable source as an explicit UserError, and rejects traversal; the focused project and validation coverage passes on 7ea4cd82.

src/bin/vip-edge-workers-build.js
```diff
@@ -0,0 +1,66 @@
...
+		description: 'Compile every worker in the project to WebAssembly.',
+	},
+	{
+		usage: 'vip edge-workers build my-worker',
+		description: 'Compile a single worker.',
+	},
+];
+
+export async function edgeWorkersBuildCommand( args = [], opt = {} ) {
+	const name = args[ 0 ];
```
rebeccahum, line 27: Minor: Add validation for args amount please, especially when `--all` is present. Example, we don't want someone to pass in `vip edge-workers build alpha --all`

src/bin/vip-edge-workers-deploy.js
```diff
@@ -0,0 +1,203 @@
...
+		usage: 'vip @example-app.develop edge-workers deploy my-worker --skip-build',
+		description: 'Deploy a previously compiled artifact without recompiling.',
+	},
+];
+
+function errorMessage( error ) {
+	return escapeTerminalText( error instanceof Error ? error.message : String( error ) );
+}
+
+function partialFailureMessage( error ) {
```
rebeccahum, line 43: I don't think this will trigger b/c by default we do `exitOnError = true` on mutations.

src/bin/vip-edge-workers-disable.js
```diff
@@ -0,0 +1,51 @@
...
+
+	await trackEventWithEnv( app.id, env.id, 'edge_workers_disable_command_execute', { name } );
+
+	try {
+		const worker = await findEdgeWorkerByName( app.id, env.id, name );
+		if ( ! worker ) {
+			throw new Error( `No edge worker named "${ name }" is deployed to this environment.` );
+		}
+
+		await setEdgeWorkerActive( env.id, worker.id, false );
```
rebeccahum, line 30: This won't get caught properly too.

src/bin/vip-edge-workers-enable.js
```diff
@@ -0,0 +1,69 @@
...
+				appName: app.name,
+				envType: env.type,
+				workerNames: [ worker.name ],
+				skipConfirmation: Boolean( opt.skipConfirmation ),
+				nonInteractive: ! isInteractiveEdgeWorkers( opt ),
+			},
+			confirm
+		);
+
+		await setEdgeWorkerActive( env.id, worker.id, true );
```
rebeccahum, line 47: Same as above ^^.

src/bin/vip-edge-workers-get.js
```diff
@@ -0,0 +1,95 @@
...
+
+	await trackEventWithEnv( app.id, env.id, 'edge_workers_get_command_execute', { name } );
+
+	if ( ! name ) {
+		exit.withError( 'Please supply the name of an edge worker.' );
+	}
+
+	let worker;
+	try {
+		worker = await getEdgeWorker( app.id, env.id, name, { includeSource } );
```
rebeccahum, line 36: Same as above ^^.

src/bin/vip-edge-workers-validate.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+command( {
+	appContext: true,
+	appQuery,
+	envContext: true,
+	usage,
+} )
+	.option( 'path', 'Path to the edge-workers project. Defaults to auto-discovery.' )
+	.option( 'all', 'Validate every worker in the project.', false )
```
rebeccahum, line 106: We need to validate args when `--all` is passed in.

src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,281 @@
...
+}
+
+/** List the edge workers deployed to an environment without source. */
+export async function listEdgeWorkers( appId: number, envId: number ): Promise< EdgeWorker[] > {
+	const api = API();
+	const response = await api.query< EdgeWorkersQueryResult >( {
+		query: gql`
+			query EdgeWorkers($appId: Int!) {
+				app(id: $appId) {
+					environments {
```
rebeccahum: Should we resolve & pass in the envId?

src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,281 @@
...
+	options: { includeSource?: boolean } = {}
+): Promise< EdgeWorker | null > {
+	const api = API();
+	const fields =
+		options.includeSource === true ? `${ EDGE_WORKER_FIELDS }\nsource` : EDGE_WORKER_FIELDS;
+	const response = await api.query< EdgeWorkersQueryResult >( {
+		query: gql`
+			query EdgeWorkerDetail($appId: Int!) {
+				app(id: $appId) {
+					environments {
```
rebeccahum: Same as above.

src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,281 @@
...
+function requireMutationPayload< T >( operation: string, value: T | null | undefined ): T {
+	if ( value === null || value === undefined ) {
+		throw new UserError( `${ operation } returned no result.` );
+	}
+	return value;
+}
+
+/** List the edge workers deployed to an environment without source. */
+export async function listEdgeWorkers( appId: number, envId: number ): Promise< EdgeWorker[] > {
+	const api = API();
```
rebeccahum: For all the API calls, we should set `exitOnError` to `false` rather than resolving to the true value so the errors can be caught and thrown as expected.

src/lib/edge-workers/toolchains/assemblyscript/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+/** Write `contents` to `filePath`, creating any missing parent directories. */
+function writeFileEnsuringDir( filePath: string, contents: string ): void {
+	fs.mkdirSync( path.dirname( filePath ), { recursive: true } );
+	fs.writeFileSync( filePath, contents );
+}
+
+function assertScaffoldTargetAvailable( projectDir: string ): void {
+	if ( ! fs.existsSync( projectDir ) ) return;
+	if ( ! fs.statSync( projectDir ).isDirectory() ) {
```
rebeccahum: Let's reject symlinks too:


```suggestion
	if ( ! fs.lstatSync( projectDir ).isDirectory() ) {
```

src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,281 @@
...
+						}
+					}
+				}
+			}
+		`,
+		variables: { appId },
+		fetchPolicy: 'no-cache',
+	} );
+
+	return pickEnvWorkers( response.data, envId ).find( worker => worker.name === name ) ?? null;
```
rebeccahum: Rather than doing this, we should be passing in the `envId` into the graphQL mutation

src/lib/api/edge-workers.ts
```diff
@@ -0,0 +1,281 @@
...
+						}
+					}
+				}
+			}
+		`,
+		variables: { appId },
+		fetchPolicy: 'no-cache',
+	} );
+
+	return pickEnvWorkers( response.data, envId );
```
rebeccahum: Same as above ^^

src/lib/edge-workers/index.ts
```diff
@@ -0,0 +1,86 @@
...
+	const name = validateWorkerName( worker.manifest.name );
+	const buildRoot = resolvePathWithin( projectDir, BUILD_DIR, 'Worker build directory' );
+	const candidate = resolvePathWithin( buildRoot, `${ name }.wasm`, 'Worker build artifact' );
+	if ( ! fs.existsSync( candidate ) ) {
+		throw new UserError(
+			`No compiled artifact found for "${ worker.manifest.name }" at "${ candidate }". ` +
+				'Run `vip edge-workers build` first, or deploy without `--skip-build`.'
+		);
+	}
+	if ( fs.lstatSync( buildRoot ).isSymbolicLink() ) {
```
rebeccahum, line 43: Can we also check `candidate`?

src/lib/edge-workers/index.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { DiscoveredWorker } from './types';
+
+export * from './types';
+export * from './project';
+export * from './location';
+export { getToolchain } from './toolchains';
+export * from './validation';
+
+/** Conventional output directory for compiled artifacts, relative to the project root. */
+export const BUILD_DIR = 'build';
```
rebeccahum: Can we just import the one in `src/lib/edge-workers/toolchains/assemblyscript/constants.ts` ?

src/lib/edge-workers/location.ts
```diff
@@ -0,0 +1,31 @@
...
+ * request paths a worker runs on; workers without one run on all requests.
+ */
+
+import UserError from '../user-error';
+import { hasTerminalControlCharacters } from './output';
+import { EDGE_WORKER_LOCATION_OPERATORS } from './types';
+
+import type { EdgeWorkerLocation, EdgeWorkerLocationOperator } from './types';
+
+export function parseLocationOption( raw: string ): EdgeWorkerLocation {
```
rebeccahum, line 13: If this flag is passed we should validate there's a value or else `.indexOf()` will throw

src/lib/edge-workers/project.ts
```diff
@@ -0,0 +1,187 @@
...
+ *   1. `--path` if provided (must contain a project descriptor).
+ *   2. Walk up from the current working directory looking for the descriptor.
+ *   3. The conventional `./edge-workers` subfolder, if present.
+ *   4. Otherwise throw a UserError with guidance.
+ */
+export function resolveProjectDir(
+	opts: { path?: string } = {},
+	cwd: string = process.cwd()
+): string {
+	if ( opts.path ) {
```
rebeccahum: Same as this one too

src/lib/edge-workers/project.ts
```diff
@@ -0,0 +1,187 @@
...
+		'No edge-workers project found here. Run `vip edge-workers init` to create one, ' +
+			'run the command from inside a project, or pass `--path` to point at one.'
+	);
+}
+
+export function readProjectDescriptor( projectDir: string ): ProjectDescriptor {
+	const file = path.join( projectDir, PROJECT_DESCRIPTOR_FILE );
+	let raw: string;
+	try {
+		raw = fs.readFileSync( file, 'utf8' );
```
rebeccahum: check for symlinks before reading

rebeccahum (commented):
Left some comments

src/bin/vip-edge-workers-disable.js
rinatkhaziev: Fixed by the `exitOnError:false` change in a0da7b40 — `setEdgeWorkerActive` failures now throw and are caught by the command `try/catch`.

src/bin/vip-edge-workers-enable.js
rinatkhaziev: Same fix as disable — `exitOnError:false` in a0da7b40 lets the enable failure reach the `try/catch`.

src/bin/vip-edge-workers-validate.js
rinatkhaziev: Done in a0da7b40. `validate` now rejects a name together with `--all` with the same message and has test coverage.

src/lib/api/edge-workers.ts
rinatkhaziev: Done in a0da7b40. All seven `API()` calls in this module now pass `{ exitOnError: false }`, so GraphQL errors reject the promise and reach the command `try/catch` instead of `process.exit(1)` in the error link. Added a test asserting each read/mutation constructs the client with `exitOnError:false`.

src/bin/vip-edge-workers-deploy.js
rinatkhaziev: Fixed by the `exitOnError:false` change in a0da7b40 — with the client no longer exiting the process on GraphQL errors, `DeploymentApplyError`/partial-failure handling here now actually runs.

src/bin/vip-edge-workers-build.js
rinatkhaziev: Done in a0da7b40. `build` now throws `Supply either a worker name or --all, not both.` before resolving workers (matching `deploy`), so `edge-workers build alpha --all` is rejected. Covered by a new bin test.

src/bin/vip-edge-workers-get.js
rinatkhaziev: Same fix — `exitOnError:false` in a0da7b40 means `getEdgeWorker` errors now propagate to the command instead of exiting the process.

src/lib/edge-workers/index.ts
rinatkhaziev: Done in a0da7b40. `readPrebuiltWorker` now also rejects a symlinked artifact (`Worker build artifact must not be a symbolic link.`) in addition to the build directory. The two existing escape tests now assert this earlier rejection.

src/lib/edge-workers/project.ts
rinatkhaziev: Done in a0da7b40. `resolveProjectDir` now rejects a non-string/empty `--path` (value-less flag = boolean `true`) with `The --path flag requires a path to the edge-workers project.` before calling `path.resolve()`. Test added.

src/lib/edge-workers/location.ts
rinatkhaziev: Done in a0da7b40. `parseLocationOption` now guards `typeof raw !== "string"` (a value-less `--location` arrives as boolean `true`) and throws a clear UserError instead of a TypeError from `.indexOf()`. Empty string still falls through to the existing `Invalid location` path. Test added.

src/lib/api/edge-workers.ts
rinatkhaziev: Same change as above — `listEdgeWorkers` now uses `environments(id: $envId)` in a0da7b40.

src/lib/api/edge-workers.ts
rinatkhaziev: Done in a0da7b40. `getEdgeWorker` and `listEdgeWorkers` now pass `envId` into the query as `environments(id: $envId)`, so the server returns only the target environment rather than all of them filtered client-side. Follows the existing pattern in `app-logs.ts`/`envvar/*`. `pickEnvWorkers` validates the single returned environment; a test asserts the query text and variables.

src/lib/edge-workers/toolchains/assemblyscript/index.ts
rinatkhaziev: Done in a0da7b40 — switched to `fs.lstatSync( projectDir ).isDirectory()` so a symlink to a directory is rejected instead of followed. Added a scaffold test for a symlinked target.

src/lib/edge-workers/project.ts
rinatkhaziev: Done in a0da7b40. Both `readProjectDescriptor` and `readWorkerManifest` now `lstat` and reject symlinks before reading, via a shared `readProjectFile` helper. Tests added for symlinked descriptor and manifest.

src/lib/edge-workers/index.ts
rinatkhaziev: Deduped in a0da7b40, but I took a slightly different route than importing from `assemblyscript/constants.ts`: I moved `BUILD_DIR` into `project.ts` next to the other project-layout constants (`WORKERS_DIR`, `PROJECT_DESCRIPTOR_FILE`, ...) and have both `index.ts` and the AssemblyScript toolchain import it from there. Importing a toolchain-specific constant up into the general `edge-workers` lib felt like a layering inversion — the build dir is a project-level convention, not AssemblyScript-specific, and both files already import from `project.ts`. Same single source of truth either way; happy to point it at `assemblyscript/constants.ts` instead if you prefer that direction.

src/lib/api/edge-workers.ts
rebeccahum: Do we still need `pickEnvWorkers`? Or can it be simplified more?

rinatkhaziev: Simplified in dbf056fa. Now that the query filters by `environments(id: $envId)`, the client-side `.find()` by id, the `.every()` id-type scan, and the `id === envId` re-check are all redundant — so `pickEnvWorkers` no longer takes `envId` and just takes the single returned environment (matching how `app-logs`/`envvar` consume `environments(id:)`).

I kept the helper rather than inlining it: it's the shared fail-closed boundary for both `listEdgeWorkers` and `getEdgeWorker`, converting any malformed/empty response into a clear `UserError`. The schema types here are hand-written rather than codegen'd, so that guard is deliberate. It's down to two shape checks now. I dropped the two obsolete id-mismatch tests and added an empty-`environments` (target-not-found) fail-closed case.

rebeccahum:
Last one: https://github.com/Automattic/vip-cli/pull/2918#discussion_r3832153881

rebeccahum (approved):

PR #2946, opened 2026-07-10
feat: support VIP_CLI_TOKEN environment variable for keychain-free authentication

spenserhale:
## Description

Adds support for a `VIP_CLI_TOKEN` environment variable that supplies the Personal Access Token directly, bypassing OS keychain storage entirely.

**Use case:** authenticated VIP-CLI usage in environments where the OS keychain is unavailable — SSH sessions on macOS (the keychain ACL authorization prompt requires a GUI and cannot be shown, so token reads fail), CI pipelines, containers, and scripting. This follows the pattern established by `gh`'s `GH_TOKEN` and this repo's own `WPVIP_DEPLOY_TOKEN`.

Behavior:

- When `VIP_CLI_TOKEN` is set (non-empty after trimming), `Token.get()` builds the token from the environment and never touches the keychain. It takes precedence over stored credentials.
- A malformed env token fails with an actionable error naming the variable and the token URL (a new `EnvTokenError`), rather than silently falling back to the keychain or crashing with an opaque message.
- `vip login` prints a note when the env var is set (the newly stored token won't be used until the variable is unset).
- `vip logout` does not invalidate a user-managed `VIP_CLI_TOKEN` server-side (matching `gh` semantics); it purges stored credentials as usual and prints a note that the env var continues to authenticate until unset.

**Known behavior:** if `VIP_CLI_TOKEN` is set but malformed, `vip login` also exits with the actionable error rather than starting the login flow. This is intentional — the env var takes precedence, so a broken value should be fixed or unset first rather than silently ignored.

Related: pairs with #2947, which skips the token read for commands that don't require auth; the two are independent and address different use cases.

### Added

- Added support for the `VIP_CLI_TOKEN` environment variable to authenticate directly without OS keychain storage, for use in SSH sessions, CI, and other non-graphical environments.

## Pull request checklist

- [x] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables.
- [x] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs).

## Steps to Test

1. Check out PR.
2. Run `npm run build`.
3. Get a Personal Access Token from https://dashboard.wpvip.com/me/cli/token
4. Run `VIP_CLI_TOKEN=<token> node ./dist/bin/vip.js whoami` — authenticates using the env token; the keychain read for auth is bypassed. (Note: with `DEBUG=@automattic/vip:keychain` you may still see one keychain line from the analytics UUID lookup — that is unrelated to authentication and its errors are swallowed by the tracker.)
5. Run `VIP_CLI_TOKEN=not-a-jwt node ./dist/bin/vip.js app list` — prints an actionable error naming `VIP_CLI_TOKEN` instead of an unhandled crash.
6. With `VIP_CLI_TOKEN` set, run `node ./dist/bin/vip.js logout` — stored credentials are purged, and a note explains the env var still authenticates until unset.
7. Best exercised over SSH to a macOS machine, where keychain-based auth fails but `VIP_CLI_TOKEN` works.

### DIFF
--- a/src/bin/vip.js
+++ b/src/bin/vip.js
@@ -8,6 +8,7 @@ import { prompt } from 'enquirer';
 
 import command, { containsAppEnvArgument } from '../lib/cli/command';
 import config from '../lib/cli/config';
+import * as exit from '../lib/cli/exit';
 import { loadInternalBin } from '../lib/cli/internal-bin-loader';
 import {
 	rewriteArgvForInternalBin,
@@ -15,7 +16,7 @@ import {
 	isSeaRuntime,
 } from '../lib/cli/sea-dispatch';
 import tokenCache from '../lib/rechallenge/token-cache';
-import Token from '../lib/token';
+import Token, { EnvTokenError, ENV_TOKEN_NAME, TOKEN_URL } from '../lib/token';
 import { aliasUser, trackEvent } from '../lib/tracker';
 
 const debug = debugLib( '@automattic/vip:bin:vip' );
@@ -28,7 +29,7 @@ if ( config && config.environment !== 'production' ) {
 }
 
 // Config
-const tokenURL = 'https://dashboard.wpvip.com/me/cli/token';
+const tokenURL = TOKEN_URL;
 const customDeployToken = process.env.WPVIP_DEPLOY_TOKEN;
 
 async function maybeExecuteSeaTargetCommand() {
@@ -107,6 +108,16 @@ async function runLoginFlow() {
 			tokenURL
 	);
 	console.log();
+
+	if ( Token.isEnvTokenSet() ) {
+		console.log(
+			chalk.yellow(
+				`  Note: ${ ENV_TOKEN_NAME } is set and takes precedence over the stored credentials. ` +
+					`The token you log in with will not be used until you unset ${ ENV_TOKEN_NAME }.`
+			)
+		);
+		console.log();
+	}
 
 	await trackEvent( 'login_command_execute' );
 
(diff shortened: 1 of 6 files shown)


spenserhale:
Hey @sjinks 

Could you review this PR?

src/bin/vip.js
```diff
@@ -193,7 +204,18 @@ const rootCmd = async function () {
...
+		token = await Token.get();
+	} catch ( err ) {
+		// A malformed VIP_CLI_TOKEN already carries an actionable message that
+		// names the env var; surface it directly instead of crashing.
+		if ( err instanceof EnvTokenError ) {
+			exit.withError( err.message );
+		}
+
+		throw err;
+	}
```
sjinks, line 218: An expired or claim-invalid but decodable environment token enters the login flow. The replacement is stored in the keychain, but subsequent requests reread and send the still-authoritative environment token.

We must treat every invalid environment token as an actionable configuration error rather than offering keychain login.

Acceptance criteria: Expired and claim-invalid VIP_CLI_TOKEN values exit without prompting, name the variable, and explain how to replace or unset it.

src/lib/logout.ts
```diff
@@ -1,15 +1,31 @@
...
 import tokenCache from '../lib/rechallenge/token-cache';
-import Token from '../lib/token';
+import Token, { ENV_TOKEN_NAME } from '../lib/token';
 import { trackEvent } from '../lib/tracker';
 
 export default async (): Promise< void > => {
 	try {
-		await http( '/logout', { method: 'post' } );
+		// VIP_CLI_TOKEN is user-managed: logout must not invalidate it server-side.
+		if ( ! Token.isEnvTokenSet() ) {
```
sjinks, line 11: When both stored token A and environment token B exist, logout skips server revocation but deletes A from the keychain (see the `finally` block). Token A remains valid server-side while its local handle is lost.

AC: Mixed-source logout neither revokes B nor silently deletes an unrevoked A.

sjinks (changes requested):

sjinks:
There are two more issues:

1. With environment token A active, `vip login` stores and aliases analytics to the entered token B, even though requests continue to authenticate as A (src/bin/vip.js:176).

2. Elevated tokens are cached only by API host and operation scope. Switching `VIP_CLI_TOKEN` between users does not perform the cache clear that interactive login performs, so user B can receive user A’s cached elevated header (src/lib/rechallenge/token-cache.ts:12).

sjinks:
What do you think if we instead add an env var, say `VIP_DISABLE_SECURE_TOKEN`, that controls which token storage is used?

PR #2947, opened 2026-07-10
fix: do not require a keychain token read for commands that do not need auth

spenserhale:
## Description

Fixes a startup crash affecting commands that don't require authentication, in sessions where the OS keychain is unavailable.

**The bug:** on macOS over SSH, every `vip` command — including `vip -h` — fails with:

```
Error: An unknown error occurred.
Error:  Unexpected error
```

**Root cause:** `rootCmd()` in `src/bin/vip.js` calls `Token.get()` unconditionally, *before* checking whether the invocation needs auth at all. Reading the stored token from the macOS keychain requires an ACL authorization prompt that can only be shown in a graphical session, so over SSH keytar rejects with a bare native error (no stack frames). The crash happens before argument parsing, so `--debug` can't produce any more information.

**The fix:**

- Compute the argv classification flags first and skip the token read entirely for invocations that never consult it: `help`/`-h`/`--help`, `-v`/`--version`, `logout`, `dev-env` without an env argument, and custom deploys keyed by `WPVIP_DEPLOY_TOKEN`. `vip -h` now performs zero keychain access.
- When a token *is* needed and the keychain read fails, exit with an actionable message (keychain locked/unavailable, common over SSH, with remediation hints and a `DEBUG=@automattic/vip:*` pointer) instead of the opaque crash.
- `Token.uuid()` (analytics) degrades to a stable per-process UUID when the keychain is unavailable, so telemetry cannot crash an otherwise-working command.
- `vip logout` tolerates an unreadable keychain: it skips the server-side `/logout` call when there is no readable valid token and always completes local cleanup.
- `doesArgvHaveAtLeastOneParam()` now stops scanning at the `--` terminator, so e.g. `vip wp <env> -- --help` is no longer misclassified as top-level help.
- The classification helpers (`doesArgvHaveAtLeastOneParam()` and the token-read predicate) are extracted into a pure `src/lib/cli/startup-args.ts` module with unit tests covering the `--` terminator handling and the full skip-flag matrix.

Related: pairs with #2946, which adds `VIP_CLI_TOKEN` authentication for these environments; the two are independent and address different use cases.

### Fixed

- Fixed a crash ("Unexpected error") when running commands that do not require authentication (such as `vip -h`, `vip -v`, and local `vip dev-env` commands) in sessions where the OS keychain is unavailable, such as over SSH.

## Pull request checklist

- [x] Update [SETUP.md](https://github.com/Automattic/vip-cli/blob/trunk/docs/SETUP.md#list-of-environmental-variables) with any new environmental variables. (No new variables.)
- [x] Update [the documentation](https://github.com/Automattic/vip-cli/blob/trunk/docs). (No documentation changes needed.)

## Steps to Test

1. Check out PR.
2. Run `npm run build`.
3. Run `DEBUG=@automattic/vip:keychain node ./dist/bin/vip.js -h` — help prints and the debug output shows **no** keychain activity.
4. Run `node ./dist/bin/vip.js -v` and `node ./dist/bin/vip.js dev-env info` — no keychain access required.
5. Full reproduction: SSH into a macOS machine that has VIP-CLI authenticated. Before this change `vip -h` crashes with "Unexpected error"; with this change it prints help, and auth-requiring commands (e.g. `vip app list`) print an actionable keychain message instead of crashing.

### DIFF
--- a/__tests__/lib/cli/startup-args.js
+++ b/__tests__/lib/cli/startup-args.js
@@ -1,0 +1,102 @@
+import { describe, expect, it } from '@jest/globals';
+
+import {
+	doesArgvHaveAtLeastOneParam,
+	isTokenReadRequired,
+} from '../../../src/lib/cli/startup-args';
+
+describe( 'lib/cli/startup-args', () => {
+	describe( 'doesArgvHaveAtLeastOneParam', () => {
+		it( 'matches a param that appears anywhere before `--`', () => {
+			expect(
+				doesArgvHaveAtLeastOneParam( [ 'node', 'vip', '--help' ], [ '-h', '--help', 'help' ] )
+			).toBe( true );
+			expect(
+				doesArgvHaveAtLeastOneParam(
+					[ 'node', 'vip', 'wp', '--help', 'list' ],
+					[ '-h', '--help', 'help' ]
+				)
+			).toBe( true );
+		} );
+
+		it( 'returns false when no param is present', () => {
+			expect(
+				doesArgvHaveAtLeastOneParam( [ 'node', 'vip', 'wp', 'list' ], [ '-h', '--help', 'help' ] )
+			).toBe( false );
+		} );
+
+		it( 'ignores params that appear only after `--`', () => {
+			expect(
+				doesArgvHaveAtLeastOneParam(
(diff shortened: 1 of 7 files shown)


spenserhale:
Hey @sjinks 

Could you review this PR?

PR #3055, opened 2026-09-14
Add native macOS and Windows installers for vip-next

rinatkhaziev:
## Description

The Go CLI currently ships portable archives, so installing it requires extracting both binaries and configuring PATH manually. This adds macOS PKG and Windows MSI installers through separate native Buildkite jobs, while retaining all existing archives for portable installations and self-updates.

### Packaging and installation

- macOS produces separate Apple Silicon and Intel packages containing `vip-next` and `go-search-replace` under `/usr/local/lib/vip-cli/bin`, with a matching `/etc/paths.d` entry. The packages reject conflicting installation paths and downgrades, and include a scoped uninstaller that preserves user configuration and credentials.
- Windows produces an x64, per-machine MSI under `%ProgramFiles%\Automattic\VIP CLI`, with a minimal installation wizard, system PATH management, major upgrades, downgrade refusal, and normal uninstall support. WiX 5.0.2 and its matching UI extension are restored into a temporary build directory; .NET is build tooling only.
- Both packages include ownership metadata. When an update is available, `vip-next update` directs installer-managed installations to the appropriate package instead of replacing their files. Portable installations retain the existing self-update behavior. No `vip` alias is introduced, and the Node/npm runtime is unchanged.

### Signing and release promotion

Binary signing and installer signing run in separate Buildkite jobs. Each installer job depends on its platform's binary job, downloads the signed archives from that exact build/step, verifies their checksums and strict two-file payload, and packages the existing signed bytes. Installer jobs use `soft_fail` so installer tooling or certificate failures cannot fail a binary release. Binary jobs remain required.

macOS binary signing retrieves only the Developer ID Application certificate. The separate installer job retrieves only Developer ID Installer in readonly mode, then signs, notarizes, and staples each PKG. Buildkite #95 confirmed the Installer identity is currently unavailable: the former combined certificate setup failed before binary signing and uploaded zero macOS artifacts. Provisioning that certificate remains necessary to produce PKGs. Windows uses the existing Azure Authenticode setup in both jobs. There is no unsigned fallback.

Installer jobs extract completed packages and compare their contents with the intended inputs before generating checksums. They explicitly upload only after every signature and payload check passes; post-failure automatic installer uploads are disabled. Prerelease promotion requires all five archive/checksum pairs (10 assets) and includes available complete installer/checksum pairs (up to 16 assets). Missing installers do not block binary releases; incomplete pairs, corrupt checksums, unexpected/duplicate/unfinished artifacts, and invalid containers still do. Native signature verification remains in Buildkite.

Both package formats use an ordered numeric version derived from the semantic release version: `major.minor.(patch*1000+stage)`, with development, alpha, beta, RC, and stable stages. Major/minor are limited to 255, patch to 64, and prerelease sequence to 199. Unsupported installer versions fail only installer jobs; the binary release version range is unchanged. Development MSI builds sharing a numeric version may require uninstalling the previous build first.

### Added

- Go CLI: Add native macOS and Windows installers with PATH configuration and installer-managed upgrades.

## Validation

Passed locally:

- `make test` — 56 Go packages reported successful results.
- `make lint`.
- Uncached focused installer, updater, release promotion, and prerelease command tests.
- Real macOS unsigned package creation/expansion, payload verification, mismatched-payload rejection, and preinstall conflict/downgrade checks against temporary volume trees.
- Failure injection proving unavailable Installer signing leaves the real macOS binary script's archives unchanged; actual Fastfile DSL checks keep certificate retrieval separate.
- Optional installer download, verification, and GitHub draft asset replacement with byte comparisons against local HTTP fixtures, plus strict binary payload extraction tests.
- Shell and Ruby syntax, WiX XML parsing, workflow/pipeline YAML parsing, and `git diff --check`.

The existing native Go Actions matrix now exercises actual installer install/upgrade/downgrade/PATH/ownership/uninstall behavior on disposable macOS and Windows runners. These tests use distinct fixture payloads, require an explicit CI-only gate, and refuse existing installations. They do not install the CLI on shared Buildkite agents.

CI on the initial commit passed macOS and Windows installer lifecycle tests, Windows signed MSI production, Linux builds, and the Go/Node checks. Buildkite #95's macOS certificate failure motivated the isolation fix above. The documentation formatting failure is also corrected. On the isolation commit `a444a163`, all three binary jobs passed in Buildkite #96, and all native Actions jobs and formatting passed. Both installer jobs actually soft-failed, despite green GitHub statuses: macOS lacks its Installer identity; Windows could not match the stored backslash artifact paths using a slash-prefixed download query. The follow-up uses a separator-independent query scoped to the Windows binary step, with a PowerShell regression that executes the real installer script against both stored path styles. Signed MSI production on this revised pipeline must be verified from the job exit status and artifacts, not its soft-fail GitHub status. Actual PKG signing still requires the unavailable Installer identity. No release was dispatched.

## Pull request checklist

- [x] Document the installer test environment variable in `docs/SETUP.md`.
- [x] Update signing, release, setup, and updater documentation.
- [x] Add automated installer version, ownership, artifact, payload, and native lifecycle tests.
- [x] Verify local Go tests and real macOS package construction.
- [x] Verify native installer lifecycle and signed Windows MSI in CI.
- [x] Verify all three separate binary jobs on the isolation fix.
- [ ] Verify the corrected Windows installer job exits successfully and uploads its MSI/checksum pair.
- [ ] Produce signed PKGs after the unavailable Installer identity is provisioned.

## Steps to Test

1. Run `make test` and `make lint`.
2. On macOS, run `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s packaging/macos -p 'test_package.py'`. This builds and inspects packages without installing them on the host.
3. Confirm the native macOS/Windows Actions jobs pass, including install, upgrade, downgrade refusal, PATH preservation, ownership instructions, and uninstall.
4. Confirm all three Buildkite binary jobs pass and upload all five archive/checksum pairs, even if an installer job fails. With the current unavailable macOS Installer identity, only the macOS installer job should warn.
5. Confirm successful installer jobs produce signed/notarized PKGs or a signed MSI, validate their extracted payloads, and upload complete package/checksum pairs. See `docs/BUILD-SIGNING.md` for the platform verification commands.


fastlane/Fastfile
```diff
@@ -60,6 +61,44 @@ lane :configure_code_signing do |readonly: true|
...
+  sync_code_signing(
+    type: 'developer_id_installer',
+    platform: 'macos',
+    team_id: APPLE_TEAM_ID,
+    app_identifier: [],
+    skip_provisioning_profiles: true,
+    readonly: true,
+    **CODE_SIGNING_STORAGE_OPTIONS
+  )
+end
```
mokagio:
```suggestion
lane :configure_installer_signing do |readonly: true|
  require_env_vars!(*CODE_SIGNING_ENV_VARS)
  sync_code_signing(
    type: 'developer_id_installer',
    platform: 'macos',
    team_id: APPLE_TEAM_ID,
    app_identifier: [],
    skip_provisioning_profiles: true,
    readonly: readonly,
    **CODE_SIGNING_STORAGE_OPTIONS
  )
end
```

mokagio:
@rinatkhaziev I just realized I didn't sync with you about pushing to the branch straight instead of having a stacked PR. Sorry for the noise.

Anyway, after a few issues, the signing process succeeds in CI.

[Before](https://buildkite.com/automattic/vip-cli/builds/122/canvas?sid=01a0a8c8-33fe-46ab-9045-1c97cbe9d71f&tab=output):

image

[After](https://buildkite.com/automattic/vip-cli/builds/131/canvas?sid=01a0b524-d32b-4b1e-9b20-1cbd1de8aed0&tab=output):

image

_Handing the ball back to you._

